PR #23, opened 2020-08-06
Avatar component

manuelpuyol:
Creates AvatarComponent with the same interface as the React component

image

Adds `avatar` slot to TimelineItem

image

Fixes `BaseComponent` bug where we couldn't set `height` and `width` attributes


app/components/primer/timeline_item_component.html.erb
```diff
@@ -1,4 +1,10 @@
 <%= render Primer::BaseComponent.new(**kwargs) do %>
+  <% if avatar %>
+    <%= render Primer::BaseComponent.new(**avatar.kwargs) do %>
```
joelhawksley, line 3: Do we need this wrapping component, or could we pass the avatar `kwargs` directly to `AvatarComponent`?

app/components/primer/avatar_component.rb
```diff
@@ -0,0 +1,28 @@
...
+      @kwargs[:tag] = :img
+      @kwargs[:src] = src
+      @kwargs[:alt] = alt
+      @kwargs[:size] = size
+      @kwargs[:height] = size
+      @kwargs[:width] = size
+
+      @kwargs[:classes] = class_names(
+        "avatar",
+        kwargs[:classes],
```
joelhawksley, line 18: It's a bit of a nitpick, but do you think we should have a standard practice for where we inject `kwargs[:classes]` in this kind of pattern? I could see it making sense to always have them be the last argument to `class_names`, or maybe the first?

manuelpuyol: adding it to the end will cause an error because the other argumets are `kwargs`

joelhawksley: Ah, so maybe we standardize on having it first?

manuelpuyol: I do prefer having the component main class first 🤷

joelhawksley: I'm not certain, but doesn't the order of the classes matter in terms of how CSS applies them? That's the reason I was thinking it would be nice to have the custom classes first or last, vs. in the middle.

manuelpuyol: it shouldn't matter
what really matters is the order they are defined in the CSS file

manuelpuyol: image

`ml-2` is defined after `ml-1` in the file, so it is more `important`

PR #28, opened 2020-08-07
Layout with flexible cols

manuelpuyol:
## Component change

Allows sidebar `col` to be customizable, calculating `main` col accordingly

col 6
image

col 3
image

## Other

1. creates`STORYBOOK` environment variable and `fetch_or_fallback` will not fail if that env var is `true`
2. separates centered and normal layouts so we can use them depending on how we want to render sotrybook for each component


manuelpuyol:
@jonrohan I don't think the changes will make it harder, it should actually help if we intend on allowing custom sizes for each layout :)

PR #29, opened 2020-08-07
Pagination component

manuelpuyol:
image


emplums:
Hey @manuelpuyol this is looking great! I'm still working through reviewing the code but I pulled storybook down and play around with it a bit and noticed a few things:

- What do you think about aligning the argument names a bit more closely to the Primer React Components argument names used? The idea here would be that switching between the two libraries would be more seamless. There are just a few that don't line up: `margin_pages` -> `margin_page_count`, `page` -> `current_page`, `surrounding_pages` -> `surrounding_page_count`

- I noticed that in some situations with fewer pages the pagination can add ellipses in weird places. For example, if you have 5 pages total and you're using `1` surrounding pages, you end up with ellipses where there shouldn't be ellipses:
image
 Maybe we could add some logic that only shows the ellipses if we are actually skipping rendering a page link? Taking a look at the model used in Primer React Components might be helpful for figuring that out: https://github.com/primer/components/blob/master/src/Pagination/model.js

manuelpuyol:
> * What do you think about aligning the argument names a bit more closely to the Primer React Components argument names used? The idea here would be that switching between the two libraries would be more seamless. There are just a few that don't line up: `margin_pages` -> `margin_page_count`, `page` -> `current_page`, `surrounding_pages` -> `surrounding_page_count`

I didn't add the `_count` because it seemed too verbose, but I can add it no problem :)

> * I noticed that in some situations with fewer pages the pagination can add ellipses in weird places. For example, if you have 5 pages total and you're using `1` surrounding pages, you end up with ellipses where there shouldn't be ellipses:
>   image

Great catch! I'll fix that and add a test

PR #35, opened 2020-08-11
update readme to develop locally with another app

manuelpuyol:


README.md
```diff
@@ -44,7 +44,27 @@ To run storybook:
...
+```
+
+to
+
+```ruby
+gem "primer_view_components", path: "path_to_the_gem" # e.g. path: "~/primer/view_components"
+```
+
+Then, `bundle install` to update references. You'll now be able to see changes from the gem without having to build it.
+Remember that restarting Rails server is necessary to see the changes, since the gem is loaded at boot.
```
joelhawksley, line 64:
```suggestion
Remember that restarting the Rails server is necessary to see changes, as the gem is loaded at boot time.
```

README.md
```diff
@@ -44,7 +44,27 @@ To run storybook:
 
 ```bash
 script/storybook
-``` 
+```
+
+### Developing with another app
+
+In your app, change the Gemfile from
```
joelhawksley, line 51:
```suggestion
In your `Gemfile`, change:
```

README.md
```diff
@@ -44,7 +44,27 @@ To run storybook:
...
+to
+
+```ruby
+gem "primer_view_components", path: "path_to_the_gem" # e.g. path: "~/primer/view_components"
+```
+
+Then, `bundle install` to update references. You'll now be able to see changes from the gem without having to build it.
+Remember that restarting Rails server is necessary to see the changes, since the gem is loaded at boot.
+
+To minimize the number of restarts, we recommend checking the component first in Storybook and when it's in a good state
```
joelhawksley, line 66:
```suggestion
To minimize the number of restarts, we recommend checking the component in Storybook first, and then when it's in a good state,

PR #37, opened 2020-08-11
Blankslate component

BenEmdon:
This PR adds the blank slate component from our primer dotcom components.

## Storybook
Screen Shot 2020-08-11 at 4 23 03 PM

### DIFF
--- a/app/components/primer/blankslate_component.html.erb
+++ b/app/components/primer/blankslate_component.html.erb
@@ -1,10 +1,6 @@
 <%= render Primer::BaseComponent.new(**@kwargs) do %>
   <% if @icon.present? %>
-    <%= render(GitHub::OcticonComponent.new(
-      icon: @icon,
-      size: :medium,
-      class: "blankslate-icon"
-    )) %>
+    <%= octicon @icon, height: 32, class: "blankslate-icon" %>
   <% end %>
 
   <% if @image_src.present? && @image_alt.present? %>
--- a/app/components/primer/blankslate_component.html.erb
+++ b/app/components/primer/blankslate_component.html.erb
@@ -7,7 +7,7 @@
     <%= image_tag "#{@image_src}", class: "mb-3", size: "56x56", alt: "#{@image_alt}" %>
   <% end %>
 
-  <h3 class="mb-1"><%= @title %></h3>
+  <<%= @title_tag %> class="mb-1"><%= @title %></<%= @title_tag %>>
 
   <% if @description.present? %>
     <p><%= @description %></p>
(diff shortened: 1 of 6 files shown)


app/components/primer/blankslate_component.html.erb
```diff
@@ -0,0 +1,31 @@
+<%= render Primer::BaseComponent.new(**@kwargs) do %>
+  <% if @icon.present? %>
+    <%= render(GitHub::OcticonComponent.new(
+      icon: @icon,
+      size: :medium,
+      class: "blankslate-icon"
+    )) %>
```
manuelpuyol, line 7: this is a big question, dotcom's `OcticonComponent` has some caching involved + some mappings thata re dotcom specific
maybe we'll need to change how the `icon` works to be a slot of some kind, which either accepts an icon name or a custom component

BenEmdon: Personally, I like that these refer to octicons, and maybe the gem should only rely on octicons. In dotcom, we could patch in the dotcom behaviour.

manuelpuyol: yeah! that's my idea

we could make `Blankslate` have an `icon` slot
the slot should receive a `icon_name` (non required) parameter
then, we could check if `icon_name` is present, if so, we render an octicon with that name. If not, we render the content provided by the user

BenEmdon: I thought b757219 seemed appropriate because we apply the blank slate class, and I think we might always want to do so? Do you think there might be cases were we don't want octicons?

manuelpuyol: I think we should be open to receive custom components there (dotcom will need to do so because of caching and all)
cc @joelhawksley

joelhawksley: @BenEmdon @manuelpuyol I'm actively working on moving the octicon caching we do in dotcom upstream so we can use it here. I'm hoping we can move to having an OcticonComponent in this library in the next couple of weeks.

BenEmdon: > we could just build one now and not use it in dotcom?

I'm a bit confused, what do you mean?

I'm in favour of whatever will unblock this PR so we can use this component in heaven.

BenEmdon:
I've been unsuccessful figuring out why the tests are failing. Am I using the slot api wrong?
CC: @manuelpuyol @joelhawksley

app/components/primer/blankslate_component.html.erb
BenEmdon: I am going to revert to use an octicon because we already use an octicon in `TimelineItemComponent`. When we decide on caching octicons, we can make loop back and make that change in Primer.

PR #39, opened 2020-08-11
WIP composable primer styles

BlakeWilliams:
This introduces a new module, `Primer::Styleable` which adds methods to a
component that generate primer based class names.

### DIFF
--- a/lib/primer/styleable.rb
+++ b/lib/primer/styleable.rb
@@ -1,0 +1,112 @@
+# frozen_string_literal: true
+
+module Primer
+  module Styleable
+    MARGIN_DIRECTION_KEYS = [:mt, :ml, :mb, :mr]
+    SPACING_KEYS = ([:m, :my, :mx, :p, :py, :px, :pt, :pl, :pb, :pr]).freeze
+
+    VALID_POSITIONS = {
+      relative: "position-relative",
+      absolute: "position-absolute",
+      fixed: "position-fixed"
+    }
+
+    def primer_class_list
+      @_primer_class_list ||= []
+    end
+
+    def primer_classes
+      primer_class_list.join(" ")
+    end
+
+    # Generate margin methods
+    # These are separate from the SPACING_KEYS since margins with directions accept negative values
+    MARGIN_DIRECTION_KEYS.each do |margin_direction|
+      define_method margin_direction do |val|
+        raise ArgumentError, "value must be between -6 and 6" if (val < -6 || val > 6)
+
+        if val < 0
+          primer_class_list << "#{margin_direction}-n#{val.abs}"
+        else
+          primer_class_list << "#{margin_direction}-#{val}"
+        end
+
+        self
+      end
+    end
+
+    # Generate spacing methods
+    SPACING_KEYS.each do |spacing|
+      define_method spacing do |val|
+        raise ArgumentError, "value must be between 0 and 6" if (val < 0 || val > 6)
+
+        primer_class_list << "#{spacing}-#{val}"
+
+        self
+      end
+    end
+
+    def f(size)
+      primer_class_list << "f#{size}"
+      self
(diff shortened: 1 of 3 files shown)


test/primer/styleable_test.rb
```diff
@@ -0,0 +1,279 @@
...
+    assert_equal "right-0", EmptyComponent.new.right0.primer_classes
+  end
+
+  def test_display
+    assert_equal "d-block", EmptyComponent.new.block.primer_classes
+    assert_equal "d-none", EmptyComponent.new.none.primer_classes
+    assert_equal "d-inline", EmptyComponent.new.inline.primer_classes
+    assert_equal "d-inline-block", EmptyComponent.new.inline_block.primer_classes
+    assert_equal "d-table", EmptyComponent.new.table.primer_classes
+    assert_equal "d-table-cell", EmptyComponent.new.table_cell.primer_classes
```
BlakeWilliams, line 92: Should probably prefix these methods with `d_`?

lib/primer/styleable.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    def position(value)
+      raise ArgumentError, "must be one of #{VALID_POSITIONS.keys.join(", ")}" unless value.in?(VALID_POSITIONS.keys)
+
+      primer_class_list << VALID_POSITIONS[value.to_sym]
+      self
+    end
+
+    def float(direction)
+      primer_class_list << "float-#{direction}"
```
manuelpuyol, line 62: should we have `VALID_DIRECTIONS`?

nuthinking:
Personally I think that View Components should treat primer classes as second class citizens (in the future these perhaps shouldn't be allowed?). Hence I'm actually more in favour of using them (for now) explicitly (i.e.: `classes: "d-inline-block etc..."`) or maybe even `primer_classes: "..."`.

test/primer/styleable_test.rb
joelhawksley: Or, could we do `d(:table_cell)`?

BlakeWilliams: > Or, could we do `d(:table_cell)`?

We totally could. I went with methods for this first attempt at an implementation instead of arguments since it will immediately raise an error instead of requiring us to validate the arguments. Thoughts?

BlakeWilliams:
> Personally I think that View Components should treat Primer classes as second class citizens (in the future these perhaps shouldn't be allowed?). Hence I'm actually more in favour of using them (for now) explicitly (i.e.: `classes: "d-inline-block etc..."`) or maybe even `primer_classes: "..."`. Finding new ways to express them could make it hard to get rid off them.

I think that's an interesting thought, and brings up a really good point, how much do we want to couple these components to Primer through methods? Additionally, how coupled do we want to allow people to opt-in to components being coupled to primer?

The original goal of this approach is to help remove the need to use `**kwargs` in so many components. The thought behind this approach was to build a module that you can include that gives you explicit methods that generate primer classes. I think one advantage here is that the DSL could be replaced/updated to support any CSS framework's classes. So if we update primer classes upstream we could easily update these methods to support the new version's CSS classes.

I'd also be 👍 on removing `**kwargs` in favor of something like:

```ruby
def initialize(tag_name:, primer_classes:)
```

Assuming that would be a 1-to-1 replacement.

nuthinking:
> So if we update primer classes upstream we could easily update these methods to support the new version's CSS classes.

Is this really a realistic scenario?

BlakeWilliams:
> Is this really a realistic scenario?

I believe so, it looks like one of the last times we had a breaking CSS name change was in December 2019 here, https://github.com/primer/css/pull/966 so there is some precedent.

BlakeWilliams:
Closing for now, since we want to stick with a more React-like API

PR #40, opened 2020-08-12
Deprecation DSL

manuelpuyol:
Implements an easy way of deprecating component's options

## Wrapper

Created `Primer::Deprecation` as a wrapper to `ActiveSupport::Deprecation` so we don't spread AS calls throughout the gem

## DSL

Added `deprecated_options_for` which receives a option and a list of values that are deprecated
e.g

```rb
deprecated_options_for :scheme, :outline, :green_outline

def initialize(title:, scheme: nil, variant: nil, **kwargs)
   @scheme = scheme
```

If `@scheme` is either `:outline` or `:green_outline` a deprecation warning will be raised


app/components/primer/component.rb
```diff
@@ -5,5 +5,31 @@ class Component < ViewComponent::Base
...
+      end
+
+      def deprecated_options_for(opt, *args)
+        deprecations[opt] = args
+      end
+    end
+
+    def before_render
+      deprecations.each do |(opt, values)|
+        val = instance_variable_get("@#{opt}")
```
joelhawksley, line 30: What happens if we never assign the value to an instance variable, or if we do, if such assignment is removed?

manuelpuyol: then `val` will be `nil` and not warn :/

I couldn't find any good (non-intrusive) way of checking the values

joelhawksley: Hmm. What if we were to pass the deprecations to `Classify` as part of `kwargs`?

manuelpuyol: hm, but not all deprecated options are from classify 🤔

joelhawksley: Seeing how this is only for our components, I think we can be pretty strict and raise an error. I assume we'd address the errors in the same PR.

manuelpuyol: done 😉

manuelpuyol:
@joelhawksley done

PR #48, opened 2020-08-18
Add PopoverComponent

cheshire137:
This implements a component for https://primer.style/css/components/popover. This also adds support for `.box-shadow` classes as seen on https://primer.style/css/utilities/box-shadow.

image


app/components/primer/popover_component.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= render Primer::BaseComponent.new(**@kwargs) do %>
+  <%= render Primer::BaseComponent.new(**@message_kwargs) do %>
+    <%= render Primer::BaseComponent.new(**@heading_kwargs) do %>
+      <%= heading %>
+    <% end %>
+    <%= body %>
+    <%= render Primer::BaseComponent.new(**@button_kwargs) do %>
```
cheshire137, line 7: Maybe I don't even want this button, but should leave it to the caller to include in `body` if they want. 🤔

app/components/primer/popover_component.rb
```diff
@@ -0,0 +1,42 @@
...
+      :left_bottom => "Popover-message--left-bottom",
+      :left_top => "Popover-message--left-top",
+      :right => "Popover-message--right",
+      :right_bottom => "Popover-message--right-bottom",
+      :right_top => "Popover-message--right-top",
+      :top_left => "Popover-message--top-left",
+      :top_right => "Popover-message--top-right"
+    }.freeze
+
+    with_content_areas :heading, :body, :button
```
manuelpuyol, line 21: can we change this to `slots`? We are trying to enforce it on new components :)

cheshire137: @joelhawksley just suggested using a slot for the message element, so that's switched over now. I'll slotify these other areas, too!

app/components/primer/popover_component.rb
```diff
@@ -0,0 +1,79 @@
...
+        )
+      end
+    end
+
+    class Heading < ViewComponent::Slot
+      attr_reader :kwargs
+
+      def initialize(**kwargs)
+        @kwargs = kwargs
+        @kwargs[:mb] = 2 unless @kwargs.key?(:mb)
```
cheshire137, line 43: This is me trying to set the default styling I see in examples on https://primer.style/css/components/popover while still allowing the user to override the bottom margin.

app/components/primer/popover_component.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= render Primer::BaseComponent.new(**@kwargs) do %>
+  <% if message %>
+    <%= render message.component do %>
```
cheshire137, line 3: I thought this might be a nice way of keeping the Message slot's `kwargs` private to itself, instead of exposing them via `attr_reader :kwargs`, which felt a bit dirty. However, my test fails with:

```
PrimerPopoverComponentTest#test_renders_default_styling:
NoMethodError: undefined method `component' for #<ViewComponent::Slot:0x00007f85f522d738 @content="">
Did you mean?  content
    test/components/popover_component_test.rb:9:in `test_renders_default_styling'
```

app/components/primer/popover_component.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      def component
+        Primer::BoxComponent.new(**@kwargs)
+      end
+    end
+
+    class Heading < ViewComponent::Slot
+      def initialize(**kwargs)
+        @kwargs = kwargs
+        @kwargs[:mb] = 2 unless kwargs.key?(:mb)
```
manuelpuyol, line 48:
```suggestion
        @kwargs[:mb] ||= 2
```

we normally use `||=` in this case

app/components/primer/popover_component.rb
```diff
@@ -0,0 +1,91 @@
...
+        @kwargs[:classes] = class_names(
+          kwargs[:classes],
+          "Popover-message box-shadow-large",
+          CARET_MAPPINGS[fetch_or_fallback(CARET_MAPPINGS.keys, caret, CARET_DEFAULT)],
+          "Popover-message--large" => large
+        )
+        @kwargs[:p] = 4 unless kwargs.key?(:p)
+        @kwargs[:mt] = 2 unless kwargs.key?(:mt)
+        @kwargs[:mx] = :auto unless kwargs.key?(:mx)
+        @kwargs[:text_align] = :left unless kwargs.key?(:text_align)
```
manuelpuyol, line 37: same `||=` comment here

app/components/primer/popover_component.rb
```diff
@@ -0,0 +1,91 @@
...
+        :right_top => "Popover-message--right-top",
+        :top_left => "Popover-message--top-left",
+        :top_right => "Popover-message--top-right"
+      }.freeze
+
+      def initialize(caret: CARET_DEFAULT, large: false, **kwargs)
+        @kwargs = kwargs
+        @kwargs[:classes] = class_names(
+          kwargs[:classes],
+          "Popover-message box-shadow-large",
```
manuelpuyol, line 30: do we **always** want `box-shadow-large`? 🤔

app/components/primer/popover_component.html.erb
manuelpuyol: I thinkyou need to declare the class you are using on the `with_slot` declaration

```rb
with_slot :message, class_name: 'Message'
with_slot :heading, class_name: 'Heading'
...
```

app/components/primer/popover_component.rb
```diff
@@ -0,0 +1,91 @@
...
+      def component
+        if @kwargs[:tag] == :button
+          Primer::ButtonComponent.new(**@kwargs)
+        else
+          Primer::BaseComponent.new(**@kwargs)
+        end
+      end
+    end
+
+    def initialize(**kwargs)
```
manuelpuyol, line 79: I'd put this `initialize` first, it was really weird seeing all the classes and only then this

app/components/primer/popover_component.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+module Primer
+  class PopoverComponent < Primer::Component
```
manuelpuyol, line 4: maybe we should add a `render?` definition, checking the necessary slots

app/components/primer/popover_component.rb
cheshire137: Okay, cool. I’d been checking #key? to allow for someone passing nil to remove any kind of bottom margin, but I suppose they could just pass 0 instead.

app/components/primer/popover_component.rb
cheshire137: I wondered about this, I didn’t see some kind of box shadow mapping in classify. Guessing it hasn’t come up before. I can add a parameter to control this.

app/components/primer/popover_component.html.erb
cheshire137: Ohh gotcha. The readme example made me think the class declaration was only necessary if you weren’t using the expected name.

app/components/primer/popover_component.rb
cheshire137: Haha I waffled on the order, too! I can put this first.

app/components/primer/popover_component.rb
cheshire137: Yeah! I feel like the message slot should be required at least.

manuelpuyol:
ohh, and before I forget. Could you add a story for the new component? 😄 

https://github.com/primer/view_components/blob/main/test/components/stories/primer/timeline_item_component_stories.rb -> is a good example on how to create stories for components with slots

app/components/primer/popover_component.rb
cheshire137: I'm trying to add `box_shadow:` to classify.rb but I'm hitting trouble allowing for the default class when you say `box_shadow: true` and a different class when you say `box_shadow: :large`. I have these tests:

```ruby
    assert_generated_class("box-shadow",             { box_shadow: true })
    assert_generated_class("box-shadow-medium",      { box_shadow: :medium })
    assert_generated_class("box-shadow-large",       { box_shadow: :large })
    assert_generated_class("box-shadow-extra-large", { box_shadow: :extra_large })
    assert_generated_class("box-shadow-none",        { box_shadow: :none })
```

I can get `assert_generated_class("box-shadow",             { box_shadow: true })` to pass but then `assert_generated_class("box-shadow-medium",      { box_shadow: :medium })` fails, or vice versa. Any thoughts?

app/components/primer/popover_component.rb
```diff
@@ -0,0 +1,91 @@
...
+    include ViewComponent::Slotable
+
+    with_slot :message, :heading, :body, :button
```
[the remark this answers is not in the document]
cheshire137: Check the component now, I dropped `body` and just use `message.content` in its place. The structure I have now looks like it will support what I was trying to do with manual Primer `.Popover` HTML before.

app/components/primer/popover_component.rb
cheshire137: Never mind, got it working in https://github.com/primer/view_components/pull/48/commits/4c2249f5ae1bcf1dab70269b0d9bbe55953e12fa!

manuelpuyol: I noticed you added `box_shadow` to `BOOLEAN_MAPPINGS`
I'm pretty sure we don't allow a prop to be boolean and concat at the same time 🤔

cheshire137: Yep, I removed that and handle it in #extract_hash now.

app/components/primer/popover_component.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= render Primer::BaseComponent.new(**@kwargs) do %>
+  <%= render message.component do %>
+    <% if heading %>
+      <%= render heading.component do %>
+        <%= heading.content %>
+      <% end %>
+    <% end %>
+    <%= message.content %>
+    <% if button %>
```
cheshire137, line 9: I find myself not setting this button slot but rather including a `<form>` with a button inside `message`. I may remove this slot.

app/components/primer/popover_component.rb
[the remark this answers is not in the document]
emplums: Hmm this is a tough one - I think we previously had different sub components for the message/heading/button/etc in Primer React Components but didn't find it to be flexible enough for all the use cases. That being said, I do like that this makes the Popover a bit more opinionated 🤔 It won't allow folks to use any button or not have a button, or put something weird like a select menu inside of the Popover. @cheshire137 do you have a code snippet of how you used this in .com that you could share with me (maybe on Slack since this repo is public)?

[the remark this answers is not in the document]
emplums: Took a look at @cheshire137 usage in .com and I think this API is reasonable. It will be a good way to dip our toes into being a little more opinionated, let's give it a shot! Nice work on this @cheshire137! 👏👏👏👏

lib/primer/view_components/version.rb
```diff
@@ -5,7 +5,7 @@ module ViewComponents
     module VERSION
       MAJOR = 0
       MINOR = 0
-      PATCH = 4
+      PATCH = 5
```
manuelpuyol, line 8: I'd like this change to be in a different PR if you don't mind
Also, could you add a `PopoverComponent` entry to the `CHANGELOG`?

cheshire137: Oh! Yep, I'll remove this version bump and modify the changelog in this branch instead.

cheshire137: Version change removed in https://github.com/primer/view_components/pull/48/commits/8c6b953dfeb705d42677d68f17132d6211d056f1, changelog updated in https://github.com/primer/view_components/pull/48/commits/8c57c8c972785ec87a10685bd13355d48a0f1463. I also added a line for `box_shadow` to the readme in https://github.com/primer/view_components/pull/48/commits/dc4d5f86f4d9aff3edd7ce3d7810d9ebe2696a85.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # main
 
+# 0.0.5
+
```
joelhawksley, line 4:
```suggestion
```

PR #50, opened 2020-08-18
Update DetailsComponent summary to be a slot

manuelpuyol:
We wanted to have `btn-sm` as our trigger to DetailsComponent, but the button was too strict and didn't accept any changes.
This PR updates the `summary` to be a slot that could either be a `ButtonComponent` or a `BaseComponent`


app/components/primer/details_component.rb
```diff
@@ -20,19 +20,23 @@ class DetailsComponent < Primer::Component
 
     with_content_areas :summary, :body
 
-    def initialize(overlay: OVERLAY_DEFAULT, button: BUTTON_DEFAULT, **kwargs)
-      @button = fetch_or_fallback(BUTTON_OPTIONS, button, BUTTON_DEFAULT)
+    def initialize(overlay: OVERLAY_DEFAULT, button: BUTTON_DEFAULT, button_props: {}, **kwargs)
```
joelhawksley, line 23: Hmm. What do you think about using a Slot for the button? That would keep us more in line with the React API: https://primer.style/components/Details

manuelpuyol: I wanted to! but that'd be a breaking change (which I'm trying to avoid for now)

manuelpuyol: If you think it is still worth, I can change it to a slot 😉

joelhawksley: For the sake of this change, I think it's worth it. Would you be willing to make the necessary changes in dotcom?

cc @muan

demo/Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
-    primer_view_components (0.0.4)
+    primer_view_components (0.0.5)
```
cheshire137, line 4: I think this will have to be something else now; 0.0.5 is already published.

cheshire137:
I liked this idea so much I wrote my own implementation of it before finding this pull request. 😅 See https://github.com/primer/view_components/pull/57.

I opted to make both summary and body into slots instead of content areas; I was thinking slots were going to replace content areas eventually. May I ask why you kept body as a content area?

app/components/primer/details_component.rb
cheshire137: > that'd be a breaking change (which I'm trying to avoid for now)

Won't going from a content area to a slot already be a breaking change?

app/components/primer/button_component.rb
```diff
@@ -20,7 +20,7 @@ class ButtonComponent < Primer::Component
     VARIANT_OPTIONS = VARIANT_MAPPINGS.keys
 
     DEFAULT_TAG = :button
-    TAG_OPTIONS = [DEFAULT_TAG, :a].freeze
+    TAG_OPTIONS = [DEFAULT_TAG, :a, :summary].freeze
```
cheshire137, line 23: Ha, just realizing we don't support `<input type="button">`. Haven't used that for years anyway!

app/components/primer/details_component.html.erb
```diff
@@ -1,6 +1,8 @@
 <%= render Primer::BaseComponent.new(**@kwargs) do %>
-  <%= render Primer::BaseComponent.new(**@summary_kwargs) do %>
-    <%= summary %>
+  <%= render summary.component do %>
+    <%= summary.content %>
+  <% end %>
+  <%= render Primer::BaseComponent.new(**body.kwargs) do %>
```
cheshire137, line 5: Could add a #component method to Body to avoid having to expose its `kwargs`.

app/components/primer/details_component.rb
```diff
@@ -7,32 +7,55 @@ module Primer
...
+        "details-reset" => reset
       )
+    end
+
+    def render?
+      summary.present? && body.present?
+    end
+
+    class Summary < Primer::Slot
+      attr_reader :kwargs, :button
```
cheshire137, line 37: I think `kwargs` doesn't need to be exposed.

stories/primer/details_component_stories.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+
+  story(:without_button) do
+    controls do
+      select(:overlay, Primer::StoriesHelper.array_to_options(Primer::DetailsComponent::OVERLAY_MAPPINGS.keys), :none)
+    end
+
+    content do |component|
+      component.slot(:summary, button: false) { "Click me" }
```
cheshire137, line 35: What does it generate with `button: false`? It looks like because `kwargs[:tag]` is still `:summary`, `return Primer::BaseComponent.new(**kwargs) unless button` will still return a `<summary>` tag.

manuelpuyol: it will still be a `summary` tag, the difference is the **component** being used.

In case we have `button: true` we are using a `Primer::ButtonComponent`, which has its own `variant`, `button_type` and other customizations

when using `button: false` it will use a `Primer::BaseComponent`, which allows a more basic implementation, without all the `Button` classes involved.

The separation was made to suppor the old `button: :reset` that was used

app/components/primer/button_component.rb
manuelpuyol: wow, that's true!
when I created this component I tried to make it as restrict as possible, only allowing tags I needed at the moment
I didn't need the `input` tag so it got ignored 🤷 

We should add support for it on a separate PR

cheshire137: > We should add support for it

We should?? I was thinking the use of `<button>` was preferred over `<input type="button">`.

stories/primer/details_component_stories.rb
cheshire137: Cool! So if I wanted just `<summary class="btn-link">`, I could do `button: false, classes: "btn-link"`?

manuelpuyol: you could!

app/components/primer/button_component.rb
manuelpuyol: now I'm confused hahaha
maybe you are right, `button` + `role` should work fine 🤔

manuelpuyol:
@cheshire137 @joelhawksley I think we reached a stable version here, do you mind taking a final look at it? :)

PR #53, opened 2020-08-19
Bump to version 0.0.5

cheshire137:
Adds some documentation about how to publish a release of the gem, as well as adds a code of conduct and contributing doc as seen in github/view_component.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,8 +38,9 @@ If you are the current maintainer of this gem:
 1. Bump gem version in `lib/primer/view_components/version.rb`. Try to adhere to [SemVer](https://semver.org).
 1. Add version heading/entries to `CHANGELOG.md`.
 1. Make a PR to primer/view_components.
-1. Build a local gem: `gem build primer_view_components.gemspec`
 1. Merge primer/view_components PR
+1. Get latest changes from default branch: `git pull origin main`
+1. Build a local gem: `gem build primer_view_components.gemspec`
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,46 @@
...
+1. Create a branch for the release: `git checkout -b release-vxx.xx.xx`
+1. Bump gem version in `lib/primer/view_components/version.rb`. Try to adhere to [SemVer](https://semver.org).
+1. Add version heading/entries to `CHANGELOG.md`.
+1. Make a PR to primer/view_components.
+1. Merge primer/view_components PR
+1. Get latest changes from default branch: `git pull origin main`
+1. Build a local gem: `gem build primer_view_components.gemspec`
+1. Tag and push: `git tag vx.xx.xx; git push --tags`
+1. Create a GitHub [release](https://github.com/primer/view_components/releases/new) with the pushed tag and populate it with a list of the changes from `CHANGELOG.md`.
+1. Push to rubygems.org -- `gem push primer_view_components-VERSION.gem`
```
manuelpuyol, line 46: Doesn't `rake release` do all of this for you?
https://schneems.com/blogs/2016-03-18-bundler-release-tasks

cheshire137: I'd like to be able to test to verify it all works as expected, so I don't necessarily want to update these docs now since they are working, but for the next version bump, that'd be great to check.

We could also drop the manual `git tag` command and pushing the tag, since you can create a tag from the GitHub /releases/new page.

PR #56, opened 2020-08-20
Deploying storybook to heroku

manuelpuyol:
See [here](https://primer-view-components.herokuapp.com/)

## Build

Removed all `public/` files from Rails, since Rails will only work as an API to render stories.
`public/` will be populated by `yarn build`, which builds storybook as static files and places them in the dir

## Heroku

Had to use an `inline-buildpack` because Heroku only accepts apps in the root folder (ours is on /demo).
The buildpack (check `bin/compile` script):
1. moves folders around so we can have `demo` as root and `demo/view_components` to be the gem
2. Changes `path: "../"` to `path: "./view_components` on Gemfile and lock
  - Heroku does not allow for conditionals on Gemfile, so I had to make this change manually


demo/config/environments/production.rb
```diff
@@ -10,7 +10,7 @@
   # your application in memory, allowing both threaded web servers
   # and those relying on copy on write to perform better.
   # Rake tasks automatically ignore this option for performance.
-  config.eager_load = true
+  config.eager_load = false
```
manuelpuyol, line 13: for some reason Rails wasn't able to load the gem's files if this was true

PR #57, opened 2020-08-20
Allow customizing summary classes in DetailsComponent

cheshire137:
This converts the content areas in DetailsComponent to be slots so that you can better customize them. I found myself wanting to use `btn-link` on the summary instead of `btn` but this didn't appear possible before.


cheshire137:
@manuelpuyol is on top of it! Closing this in favor of #50.

PR #58, opened 2020-08-24
Suggestion: a modest helper to reduce boilerplate

inopinatus:
I've been following the View Components journey and I was inevitably curious about the Primer components repo.  The instantiation boilerplate immediately struck me as verbose, certainly for own my personal taste, and my instinct was to immediately tuck that into a helper to improve legibility.

So I'm just sending this PR by way of suggestion, you may of course have other ideas or style preferences!

### DIFF
--- a/lib/primer/view_helper.rb
+++ b/lib/primer/view_helper.rb
@@ -1,12 +1,51 @@
 # frozen_string_literal: true
 
-# Provides a convenient shorthand for components.
-module Primer::ViewHelper
-  Components = Hash.new do |hash, component|
-    hash[component] = "Primer::#{component.to_s.camelize}Component".constantize
-  end
+module Primer
+  MissingComponentError = Class.new(NameError)
 
-  def primer(component, **kwargs, &block)
-    render Components[component].new(**kwargs), &block
+  # Provides a convenient shorthand for rendering Primer components.
+  module ViewHelper
+    # Writing to the Components hash in an initializer is supported for class names
+    # that can't be inferred by camelizing, e.g.
+    #
+    #   Primer::ViewHelper::Components[:special] = Primer::VerySpecialComponent
+    #
+    # or to prevent stale classes when reloading in development:
+    #
+    #   Rails.configuration.to_prepare do
+    #     Primer::ViewHelper::Components.clear if Rails.env.development?
+    #   end
+    Components = Hash.new do |hash, component|
+      hash[component] = primer_component_class_for(component)
+    end
+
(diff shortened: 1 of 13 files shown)


lib/primer/view_helper.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Provides a convenient shorthand for components.
+module Primer::ViewHelper
```
manuelpuyol, line 4: could you break it into

```rb
module Primer
  module ViewHelper
```
?

lib/primer/view_helper.rb
```diff
@@ -0,0 +1,12 @@
...
+  Components = Hash.new do |hash, component|
+    hash[component] = "Primer::#{component.to_s.camelize}Component".constantize
+  end
+
+  def primer(component, **kwargs, &block)
+    render Components[component].new(**kwargs), &block
+  end
```
manuelpuyol, line 11: I'd like to see some error handling here 😄

lib/primer/view_helper.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Provides a convenient shorthand for components.
+module Primer::ViewHelper
+  Components = Hash.new do |hash, component|
+    hash[component] = "Primer::#{component.to_s.camelize}Component".constantize
```
manuelpuyol, line 6: for now I guess this should be fine, but I imagine we'll have some more complex namespaces in the future

joelhawksley:
@inopinatus thanks for taking the time to make this suggestion via a PR! It's super helpful to have proposed code when discussing ideas likes this. I too share a distaste for how verbose `render` calls can look with ViewComponents.

However, we've generally avoided introducing a shorthand for rendering components, as we have several static analysis tools internally that rely on component class names being referenced directly. I'm hesitant to make this change now given that dependency.

lib/primer/view_helper.rb
inopinatus: 👍

lib/primer/view_helper.rb
inopinatus: Good thought. The most likely due to the helper is a NameError. Although my default handling would be to re-throw it.

inopinatus:
@joelhawksley @manuelpuyol thanks for taking a look! I'll close this PR for now.

I am making a ViewComponent-based library for my own app's UI, so I fully intend to shamelessly borrow from whatever structures you land on.

PR #59, opened 2020-08-25
Change error message

emplums:
This PR changes the Primer CSS class name error message to be a bit more user friendly


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # main
 
+
+# 0.0.6
+
+* Updated the invalid class name error message
+
+# 0.0.5
```
emplums, line 8: it looks like we already released v 0.0.5 but there's nothing in it in the changelog so I just left it blank

PR #61, opened 2020-08-25
Pass down inline styles in ProgressBar

emplums:
This PR allows users to pass down inline styles in the ProgressBar component for ProgressBar item slots - this allows people to animate the progress bar when needed. This is used in the files viewed progress bar in the PR review process: 
image

I think this opens up a broader question on whether or not we should allow inline styles on other components - my hunch is that we should only allow it on a as needed basis for now, to keep the component API a little more strict/opinionated.

### DIFF
--- a/app/components/primer/progress_bar_component.rb
+++ b/app/components/primer/progress_bar_component.rb
@@ -52,13 +52,14 @@ module Primer
       include ClassNameHelper
       attr_reader :kwargs
 
-      def initialize(percentage: 0, bg: :green, **kwargs)
+      def initialize(percentage: 0, bg: :green, style: "", **kwargs)
         @percentage = percentage
         @kwargs = kwargs
+        @style = style
 
         @kwargs[:tag] = :span
         @kwargs[:bg] = bg
-        @kwargs[:style] = "width: #{@percentage}%;"
+        @kwargs[:style] = @style + "width: #{@percentage}%;"
         @kwargs[:classes] = class_names("Progress-item", @kwargs[:classes])
       end
     end
--- a/app/components/primer/progress_bar_component.rb
+++ b/app/components/primer/progress_bar_component.rb
@@ -52,7 +52,7 @@ module Primer
       include ClassNameHelper
       attr_reader :kwargs
 
-      def initialize(percentage: 0, bg: :green, style: "", **kwargs)
+      def initialize(percentage: 0, bg: :green, **kwargs)
         @percentage = percentage
         @kwargs = kwargs
         @style = style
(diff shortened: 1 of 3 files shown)


app/components/primer/progress_bar_component.rb
```diff
@@ -52,13 +52,14 @@ class Item < ViewComponent::Slot
...
-      def initialize(percentage: 0, bg: :green, **kwargs)
+      def initialize(percentage: 0, bg: :green, style: "", **kwargs)
         @percentage = percentage
         @kwargs = kwargs
+        @style = style
 
         @kwargs[:tag] = :span
         @kwargs[:bg] = bg
-        @kwargs[:style] = "width: #{@percentage}%;"
+        @kwargs[:style] = @style + "width: #{@percentage}%;"
```
joelhawksley, line 62: If someone forgets to include a terminating semicolon when passing in `style`, this will break subtly. What if we instead added the provided `style` to the end?

```suggestion
        @kwargs[:style] = "width: #{@percentage}%; #{@kwargs[:style]}"
```

app/components/primer/progress_bar_component.rb
```diff
@@ -52,13 +52,14 @@ class Item < ViewComponent::Slot
       include ClassNameHelper
       attr_reader :kwargs
 
-      def initialize(percentage: 0, bg: :green, **kwargs)
+      def initialize(percentage: 0, bg: :green, style: "", **kwargs)
```
joelhawksley, line 55: I believe we can lean on `kwargs` here. See suggestion below 😄 

```suggestion
      def initialize(percentage: 0, bg: :green, **kwargs)
```

app/components/primer/progress_bar_component.rb
emplums: good catch!

jonrohan:
One concern here from me, is passing inline styles takes that bit of css out of the build pipeline. If someone were to pass a property that requires a prefix `-webkit-` it could lead to a situation where it won't work in some browsers. 

Currently we rely on autoprefixer and postcss to do this for us.

emplums:
@jonrohan that's a good point, I would like it if folks only used this for rare edge cases. In this case this was added because there was already - https://github.com/github/github/pull/152516/files#diff-d7b8822de0a9098171f81a7ed1182110L25

maybe it would be better to add a custom CSS class here instead?

manuelpuyol:
I agree with @jonrohan. Using styles directly may have weird behaviors..
I wonder if we should provide new helper classes here or if we should only rely on Primer

manuelpuyol:
Closing since this was resolved in another PR

PR #69, opened 2020-09-08
use `content_tag` instead of tag.send

joelhawksley:
It's more or less just a matter of
semantics, but I think `content_tag`
is a cleaner implementation here than
using `tag.send`.


BlakeWilliams (approved):
Nice! 👍

PR #99, opened 2020-10-02
DropdownMenuComponent when used with DetailsComponent must be a direct descendant 

jonrohan:
The `<details-menu>` custom element is required to be a direct descendant of `<details>` https://github.com/github/details-menu-element/blob/73048a145256e7fabf9e0a1799125de427942ab8/src/index.ts#L125

There is functionality that's missing in the combination of using `DetailsComponent` and `DropdownMenuComponent`. Take this example:

```ruby
    render(Primer::DetailsComponent.new) do |component|
      component.slot(:summary, variant: :small, button_type: :primary) do
        "Summary"
      end
      component.slot(:body) do
        render(Primer::DropdownMenuComponent.new(direction: :w)) { "Body" }
      end
    end
```

When rendered it produces a wrapping `<div>` tag for the body. This breaks close functionality built into the custom element. The requirement of a direct child `details > details-menu` is because there is instances where a `details-menu` contains a `details` element. **So the menu cannot close on click**

**What should be happening.gif**

Untitled

```html
<details open="">
    <summary aria-haspopup="menu" role="button">Best robot: <span data-menu-button="">Hubot</span></summary>
    <details-menu role="menu">
      <button type="button" role="menuitem" data-menu-button-text="">Hubot</button>
      <button type="button" role="menuitem" data-menu-button-text="">Bender</button>
      <button type="button" role="menuitem" data-menu-button-text="">BB-8</button>
    </details-menu>
  </details>
```

**What is happening.gif**

Untitled 2 

```html
<details open="">
    <summary aria-haspopup="menu" role="button">Best robot: <span data-menu-button="">Hubot</span></summary>
    <div><details-menu role="menu">
      <button type="button" role="menuitem" data-menu-button-text="">Hubot</button>
      <button type="button" role="menuitem" data-menu-button-text="">Bender</button>
      <button type="button" role="menuitem" data-menu-button-text="">BB-8</button>
    </details-menu></div>
  </details>
```

### Ideas for fixing

* We could remove the wrapping primer element.
* Create a combo component DetailsMenuComponent that does the correct thing.

### DIFF
--- a/test/components/details_component_test.rb
+++ b/test/components/details_component_test.rb
@@ -94,4 +94,18 @@ class PrimerDetailsComponentTest < Minitest::Test
     refute_selector("details")
     refute_selector("summary")
   end
+
+  def test_details_menu_is_direct_child
+    menu = render_inline(Primer::DropdownMenuComponent.new(direction: :w)) { "Body" }
+    render_inline(Primer::DetailsComponent.new) do |component|
+      component.slot(:summary, variant: :small, button_type: :primary) do
+        "Summary"
+      end
+      component.slot(:body) do
+        menu.to_s.html_safe
+      end
+    end
+
+    assert_selector("details > details-menu", visible: false)
+  end
 end


natashaU:
I just had a brief discussion with @joelhawksley about something similar yesterday --- it seems the DropdownMenuComponent is wrapped within a DetailsMenu so it would make sense to have a combo component DetailsMenuComponent or include the dropdown within a DetailsMenuComponent?

jonrohan:
> it seems the DropdownMenuComponent is wrapped within a DetailsMenu so it would make sense to have a combo component DetailsMenuComponent or include the dropdown within a DetailsMenuComponent?

A combo component makes the most sense I think, but I do have a few questions. How much cleanup work would it be to replace any existing uses of DropdownMenuComponent and DetailsMenu in dotcom? How will we let people in the future know that they shouldn't use these two components together?

natashaU:
> DropdownMenuComponent

Doing a quick search, there are only 2 uses of the DropdownMenuComponent and 7 uses of the DetailsComponent in dotcom --- so I imagine it wouldn't be that much work to clean up. 

I have a question about your question (we getting meta here, haha :neckbeard:  ): Would it make sense to get rid of the DropdownMenu Component all together and just include it within the DetailsMenu? Or are there are any uses cases where we'd use a DropdownMenu without a details component?

jonrohan:
> I have a question about your question (we getting meta here, haha :neckbeard: ): Would it make sense to get rid of the DropdownMenu Component all together and just include it within the DetailsMenu? Or are there are any uses cases where we'd use a DropdownMenu without a details component?

That makes sense to me. We'd only ever put DropdownMenuComponent inside a DetailsComponent, so we could combine them into one component. Similar to the react version https://primer.style/components/Dropdown

But there are cases where we'll still want to have a separate DetailsComponent that's just the surrounding `<details>` html for show hiding things https://primer.style/components/Details

manuelpuyol:
Closing in favor of https://github.com/primer/view_components/pull/192

PR #100, opened 2020-10-06
Primer::OcticonComponent classify options are leaking as invalid html attributes

heynan0:
Before: `Primer::OcticonComponent` is leaking Classify attributes into the HTML element attributes. These attributes are invalid. See screenshot below (`classes`, `display`).

Screen Shot 2020-10-05 at 3 39 57 PM

After:

* ~~Passed in `class` and `classes` arguments are concatenated into one and show under html `class` attribute.~~
  * Update: now, only `classes` option is supported. `class` will no longer work.
* Classify attributes are filtered out before passing options to octicon helper, preventing any leakage into invalid html attributes. 

Screen Shot 2020-10-06 at 12 02 25 PM
legend for screenshot immediately above

* red: from classify options: `d-none d-sm-inline`
* ~~green: from `class`: `abc`~~ (update not shown in screenshot: no longer supported)
* cyan: from `classes`: `UnderlineN2322av-octicon`

### DIFF
--- a/app/components/primer/octicon_component.rb
+++ b/app/components/primer/octicon_component.rb
@@ -15,9 +15,10 @@ module Primer
 
     def initialize(icon:, size: SIZE_DEFAULT, **kwargs)
       @icon, @kwargs = icon, kwargs
-
+      classes = class_names(@kwargs.delete(:class), @kwargs.delete(:classes))
+      @kwargs[:class] = class_names(Primer::Classify.call(**@kwargs)[:class], classes)
+      @kwargs = @kwargs.except(*Primer::Classify::VALID_KEYS)
       @kwargs[:height] = SIZE_MAPPINGS[size]
-      @kwargs[:class] = class_names(@kwargs[:class], Primer::Classify.call(**@kwargs)[:class])
     end
 
     def call
--- a/app/components/primer/octicon_component.rb
+++ b/app/components/primer/octicon_component.rb
@@ -16,13 +16,13 @@ module Primer
     def initialize(icon:, size: SIZE_DEFAULT, **kwargs)
       @icon, @kwargs = icon, kwargs
 
-      @kwargs[:class] = class_names(Primer::Classify.call(**@kwargs)[:class], @kwargs.delete(:classes))
+      @kwargs[:class] = Primer::Classify.call(**@kwargs)[:class]
       @kwargs[:height] ||= SIZE_MAPPINGS[size]
 
       # Filter out classify options to prevent them from becoming invalid html attributes.
       # Note height and width are both classify options and valid html attributes.
       octicon_helper_options = @kwargs.slice(:height, :width)
-      @kwargs = @kwargs.except(*Primer::Classify::VALID_KEYS).merge(octicon_helper_options)
+      @kwargs = @kwargs.except(*Primer::Classify::VALID_KEYS, :classes).merge(octicon_helper_options)
     end
 
     def call
--- a/app/components/primer/octicon_component.rb
+++ b/app/components/primer/octicon_component.rb
@@ -15,10 +15,15 @@ module Primer
 
     def initialize(icon:, size: SIZE_DEFAULT, **kwargs)
       @icon, @kwargs = icon, kwargs
+
       classes = class_names(@kwargs.delete(:class), @kwargs.delete(:classes))
       @kwargs[:class] = class_names(Primer::Classify.call(**@kwargs)[:class], classes)
-      @kwargs = @kwargs.except(*Primer::Classify::VALID_KEYS)
-      @kwargs[:height] = SIZE_MAPPINGS[size]
+      @kwargs[:height] ||= SIZE_MAPPINGS[size]
+
+      # Filter out classify options to prevent them from becoming invalid html attributes.
+      # Note height and width are both classify options and valid html attributes.
+      octicon_helper_options = @kwargs.slice(:height, :width)
+      @kwargs = @kwargs.except(*Primer::Classify::VALID_KEYS).merge(octicon_helper_options)
     end
 
     def call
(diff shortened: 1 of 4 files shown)


test/components/octicon_component_test.rb
```diff
@@ -59,9 +59,27 @@ def test_renders_class_passed_in
...
+  def test_renders_concatenated_class_and_classes_passed_in
+    render_inline(Primer::OcticonComponent.new(icon: "star", class: "foo", classes: "bar"))
+
+    assert_selector(".foo.bar")
+  end
+
+  def test_does_not_render_classify_attributes_as_html_attributes
+    render_inline(Primer::OcticonComponent.new(icon: "star", classes: "foo", my: 4, display: [:none]))
+
+    refute_selector('[classes="foo"][display="none"]')
```
joelhawksley, line 83: Might we simplify this test as follows:

```suggestion
    render_inline(Primer::OcticonComponent.new(icon: "star", classes: "foo"))

    refute_selector('[classes="foo"]')
```

app/components/primer/octicon_component.rb
```diff
@@ -15,9 +15,10 @@ class OcticonComponent < Primer::Component
 
     def initialize(icon:, size: SIZE_DEFAULT, **kwargs)
       @icon, @kwargs = icon, kwargs
-
+      classes = class_names(@kwargs.delete(:class), @kwargs.delete(:classes))
+      @kwargs[:class] = class_names(Primer::Classify.call(**@kwargs)[:class], classes)
+      @kwargs = @kwargs.except(*Primer::Classify::VALID_KEYS)
```
joelhawksley, line 20: Is there a specific case this line this is meant to filter out?

test/components/octicon_component_test.rb
heynan0: @joelhawksley the point of the test is so it doesn't put `display` attribute in the html test. I think having the two, classes, and display, reinforces the test.

app/components/primer/octicon_component.rb
heynan0: @joelhawksley if you don't remove keys like `display` it will pass to the html element:

```
<svg display="..." />
```

joelhawksley: @heynan0 makes sense. We're passing these options to `octicon`, which ends up calling `Octicon.new`. Can we confirm that at least one of these options works:

https://github.com/primer/octicons/blob/master/lib/octicons_gem/lib/octicons/octicon.rb#L11

test/components/octicon_component_test.rb
joelhawksley: @heynan0 my concern with asserting both at the same time is that it makes `refute_selector` risky to use here. 

If we introduced a bug that say, didn't filter out `classes`, but did filter out `display`, this test would still pass, right?

app/components/primer/octicon_component.rb
heynan0: Forgot to mention I had done a search for instances into how the component is used and sometimes people pass classes through `class`, sometimes through `classes`, that's why I kept the support for both.

joelhawksley: @heynan0 what do you think about refactoring them to always pass `class`? That would simplify things...

test/components/octicon_component_test.rb
heynan0: @joelhawksley makes sense. I split it into two statements.

app/components/primer/octicon_component.rb
heynan0: @joelhawksley we talked in Slack about how this would break to the consumers of the library. You said it was OK since it's in alpha.

I also thought this:

Also I thought you had set on "classes" as the standard name? not class? 
(you replied: you meant to say `classes`)

Thinking back at `classes` vs `class` this is what I consider:
```
classes adds to the underlying component classes. e.g. Label already has "Label" class so classes adds more.

class would replace all classes in the component , overwriting the underlying ones. e.g. Label would lose "Label" if you passed class: "something-else"
```

heynan0: I changed the code so it only takes `classes`. The option `class` is no longer supported.

app/components/primer/octicon_component.rb
```diff
@@ -16,8 +16,7 @@ class OcticonComponent < Primer::Component
     def initialize(icon:, size: SIZE_DEFAULT, **kwargs)
       @icon, @kwargs = icon, kwargs
 
-      classes = class_names(@kwargs.delete(:class), @kwargs.delete(:classes))
-      @kwargs[:class] = class_names(Primer::Classify.call(**@kwargs)[:class], classes)
+      @kwargs[:class] = class_names(Primer::Classify.call(**@kwargs)[:class], @kwargs.delete(:classes))
```
joelhawksley, line 19: `Classify` should handle `classes` for us, right? I think we can just do:

```suggestion
      @kwargs[:class] = Primer::Classify.call(**@kwargs)[:class]
```

heynan0: @joelhawksley , if we do as suggested, the `classes` property will be passed along to the call to octicon and regress to the problem we are trying to solve.

I simplified the code doing the following:

https://github.com/heynan0/view_components/commit/7bbafe0ad602796d87adf3e16028ef8ca7f13c76

joelhawksley (approved):
Thanks @heynan0 ❤️ 

Mind adding a CHANGELOG entry? We should probably call out the breaking change here.

heynan0:
@joelhawksley I added a new version, 0.0.9:

https://github.com/primer/view_components/pull/100/commits/f9ff81bfdd44dea6576627c0003dc810dc5aec1f

joelhawksley (approved):

PR #102, opened 2020-10-12
Render documentation examples

joelhawksley:
This PR introduces basic rendering support for YARD examples, allowing us to render them in the generated markdown!

cc @manuelpuyol @colebemis


Rakefile
```diff
@@ -64,28 +66,45 @@ namespace :docs do
...
+      f.puts
+
+      initialize_method = documentation.meths.find(&:constructor?)
+
+      initialize_method.tags(:example).each do |tag|
+        f.puts("### #{tag.name}")
+        f.puts
+
+        # rubocop:disable Security/Eval
+        html = render_inline(eval(tag.text.gsub("<%= render(", "").gsub(") %>", ""))).to_html
```
joelhawksley, line 80: This rake task is only run locally, so this `eval` should be OK here.

BlakeWilliams: Do we think we'll ever run this via actions? Is there any kind of safety we could add here to ensure we don't run it in that environment?

BlakeWilliams (approved):
one minor question, but LGTM

PR #107, opened 2020-10-16
Add SpinnerComponent

colebemis:
## Problem

Spinners are a high priority image asset that we need to update for dark mode because they're used in _a lot_ of views and look terrible on a dark background.

## Solution

This PR adds a `Primer::SpinnerComponent` to replace our old spinner GIF:

| Before | After |
| --- | --- |
|   |   |
|   |   |

## Checklist

- [x] Update documentation
- [x] Update tests
- [x] Update changelog


app/components/primer/spinner_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render Primer::BaseComponent.new(**@kwargs) do %>
+  <circle cx="8" cy="8" r="7" stroke="currentColor" stroke-opacity="0.25" stroke-width="2" vector-effect="non-scaling-stroke" />
+  <path d="M15 8a7.002 7.002 0 00-7-7" stroke="currentColor" stroke-width="2" stroke-linecap="round" vector-effect="non-scaling-stroke">
+    <animateTransform attributeName="transform" type="rotate" from="0 8 8" to="360 8 8" dur="1s" repeatCount="indefinite" />
+  </path>
+<% end %>
+
+
```
joelhawksley, line 8:
```suggestion
```

app/components/primer/spinner_component.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    #
+    # @example 48|Default
+    #   <%= render(Primer::SpinnerComponent.new) %>
+    #
+    # @example 32|Small
+    #   <%= render(Primer::SpinnerComponent.new(size: 16)) %>
+    #
+    # @param size [Integer] Sets the size of the spinner
+    def initialize(size: 32, **kwargs)
```
joelhawksley, line 15: Do we want to standardize the `size` value? I could see us matching `OcticonComponent`, which has `small`, `medium`, and `large`. (16, 32, and 64)

app/components/primer/spinner_component.rb
```diff
@@ -0,0 +1,27 @@
...
+      @kwargs = kwargs
+      @kwargs[:tag] ||= :svg
```
joelhawksley, line 17: I believe this should _always_ be `svg`:

```suggestion
      @kwargs[:tag] = :svg
```

app/components/primer/spinner_component.rb
colebemis: Good idea! Should we restrict the size to small/medium/large or should we allow a number to be passed too?

colebemis: I think OcticonComponent allows you to pass in a number too, so maybe we should be consistent with that?

joelhawksley: How about we start with a fixed list? We can always relax the API to accept numbers.

colebemis: Works for me 👍

colebemis: Updated ✅ 

Is there a way to write the comments so this arguments table is more helpful? 
image

joelhawksley: @colebemis What might a more helpful table look like? 

If you're looking to list the possible values, this might be a good reference:

https://github.com/primer/view_components/blob/main/app/components/primer/blankslate_component.rb#L59

Perhaps we could write a helper for this?

colebemis: Exactly what I was looking for. Thanks! A helper could be nice but I'm not sure what we'd name it.

joelhawksley (approved):
_Fantastic!_

PR #119, opened 2020-10-22
Update details_component.rb

kenyonj:
This PR:
  - Renames the constant associated with the `:none` for overlay mapping
  - Adds a CHANGELOG entry

Why?
  - This removes confusion that could be caused from having an actual `:default` key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## main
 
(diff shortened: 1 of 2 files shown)


joelhawksley (approved):
Looks good. Mind adding a changelog?

natashaU (approved):
Look great ✅

PR #121, opened 2020-10-23
Add Menu component

joelhawksley:
This PR implements the Primer CSS Menu component in ViewComponents.

cc @BlakeWilliams as another example of Slots.


app/components/primer/menu_component.rb
```diff
@@ -0,0 +1,68 @@
...
+      @kwargs = kwargs
+      @kwargs[:tag] = :nav
+      @kwargs[:classes] = class_names(kwargs[:classes], "menu")
+    end
+
+    def before_render
+      @kwargs['aria-labelledby'] = heading.kwargs[:id] if heading.present?
+    end
+
+    class Heading < Primer::Slot
```
BlakeWilliams, line 37: I was curious what it would look like with v2 + the lambda proposal, and I think we could get away with the following:

```ruby
renders_one :heading, -> (id: "menu-heading", tag: :span, classes: "") do
  Primer::BaseComponent.new(id: id, tag: tag, classes: class_names(classes, "menu)
end
```

Then the view could be changed from:

```erb
  <% if heading %>
    <%= render Primer::BaseComponent.new(**heading.kwargs) do %>
      <%= heading.content %>
    <% end %>
  <% end %>
```

to

```erb
<%= heading if heading.present? %>
```

Since it looks like the slot is essentially delegating to a component, we could remove some boilerplate with the above approach.

BlakeWilliams (approved):

Rakefile
```diff
@@ -96,7 +97,7 @@ namespace :docs do
       # Primer::AvatarComponent => Avatar
       short_name = component.name.demodulize.gsub("Component", "")
 
-      File.open("docs/content/components/#{short_name}.md", "w") do |f|
+      File.open("docs/content/components/#{short_name.downcase}.md", "w") do |f|
```
manuelpuyol, line 100:
```suggestion
      File.open("docs/content/components/#{short_name.underscore}.md", "w") do |f|
```

This will make `ProgressBar` -> `progress_bar` instead of `progressbar`

manuelpuyol (approved):

PR #124, opened 2020-10-30
Allow no box-sizing style for SpinnerComponent

MrChrisW:
### Summary

This adds the option to disable adding the `box-sizing` style to the `SpinnerCompnent` added in https://github.com/primer/view_components/pull/107

Having a style causes CSP errors as we don't allow inline stlyes over at https://github.com/github/helphub/issues/925. We don't need to add padding to the spinner so the ability to disable this style is needed. 

@colebemis Let me know if this sounds alright or you recommend another approach?


colebemis (approved):
This is fine with me 👍

PR #126, opened 2020-11-02
Add prerelease disclaimer

joelhawksley:
Adds a disclaimer about the pre-release status of the library.


colebemis (approved):
👍 You could also use the `Note` component if you want to add more emphasis.

PR #132, opened 2020-11-09
rename kwargs to system_arguments

joelhawksley:
This PR renames the `kwargs` argument in the codebase to `system_arguments`, making the purpose of the argument clear.

### DIFF
--- a/app/components/primer/avatar_component.rb
+++ b/app/components/primer/avatar_component.rb
@@ -15,25 +15,25 @@ module Primer
     # @param alt [String] Passed through to alt on img tag
     # @param size [Integer] Adds the avatar-small class if less than 24
     # @param square [Boolean] Used to create a square avatar.
-    def initialize(src:, alt:, size: 20, square: false, **kwargs)
-      @kwargs = kwargs
-      @kwargs[:tag] = :img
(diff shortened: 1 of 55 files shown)


BlakeWilliams (approved):
I like this a lot!

What are your thoughts on removing, or renaming the `_arguments` part of the name since it's now the "system_arguments argument". Maybe `system_options` or possibly `primer_options`? Not a blocker though 🙂

joelhawksley:
> What are your thoughts on removing, or renaming the _arguments part of the name since it's now the "system_arguments argument".

I think there's room for improvement here for sure! For now, I wanted to match the naming exactly, so it's clear to readers of the docs that these arguments correspond to the System Arguments part of the documentation.

natashaU (commented):
✅

natashaU:
> > What are your thoughts on removing, or renaming the _arguments part of the name since it's now the "system_arguments argument".
> 
> I think there's room for improvement here for sure! For now, I wanted to match the naming exactly, so it's clear to readers of the docs that these arguments correspond to the System Arguments part of the documentation.

Agree here. I prefer system_arguments myself since this is the most clear to me.

PR #135, opened 2020-11-16
add support for functional colors to Label

joelhawksley:
This PR adds support for functional colors to Label to unblock @colinkeany.

We'll need to follow up with proper docs / storybook support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## main
 
+## 0.0.14
+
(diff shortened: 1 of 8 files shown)


natashaU (approved):
The PR Looks great ✅ 

Quick Question: Why the change from functional names rather than explicit names? I find something like "green" more clear than "success" for example and I find the primary names a bit confusing. For example, I found myself thinking: "What corresponds to yellow? Info? "

Issue #120, opened 2020-10-22
Improve style arguments docs


joelhawksley:
Currently, the [style arguments](https://view-components.primer.vercel.app/view-components/style-arguments) documentation is written from the perspective of migrating from Primer CSS. 

While this is helpful for some users of the library, we aren't providing guidance for using style arguments from scratch. I think we should rewrite this documentation to better serve folks using the library for the first time.

joelhawksley:
Note: These docs should include section on passing in a `classes` argument.

joelhawksley:
Note: We should add a section on data attributes: https://github.slack.com/archives/C0ZCGGGJ2/p1603810527286100

joelhawksley:
Closing as https://primer.style/view-components/system-arguments is meaningfully complete.

PR #138, opened 2020-11-19
Another pass at system arguments docs

joelhawksley:
This PR fills out the remaining gaps of our documentation for system arguments.

While what we have here isn't the most easily digested, it is more or less complete. I hope we can introduce a more friendly structure soon.

cc @yaili as I've addressed your capitalization concerns here 😄 
cc @kevinsawicki as I've added some examples based on your feedback. (Such as `hidden`, `data` attributes, etc)

Related: https://github.com/primer/view_components/issues/120

### DIFF
--- a/app/components/primer/base_component.rb
+++ b/app/components/primer/base_component.rb
@@ -25,13 +25,14 @@ module Primer
   #
   # ## HTML attributes
   #
-  # System arguments sinclude most HTML attributes. For example:
+  # System arguments include most HTML attributes. For example:
   #
   # | Name | Type | Description |
   # | :- | :- | :- |
   # | `width` | `Integer` | Width. |
   # | `height` | `Integer` | Height. |
   # | `data` | `Hash` | Data attributes. For example: `data: { foo: :bar }` will render `data-foo='bar'`. |
+  # | `aria` | `Hash` | Aria attributes. For example: `aria: { label: "foo" }` will render `aria-label='foo'`. |
   # | `title` | `String` | The `title` attribute. |
   # | `hidden` | `Boolean` | Whether to assign the `hidden` attribute. |
   class BaseComponent < Primer::Component
--- a/app/components/primer/base_component.rb
+++ b/app/components/primer/base_component.rb
@@ -33,13 +33,13 @@ module Primer
     # @param ml [Integer] Margin left. <%= one_of((-6..6).to_a) %>
     # @param mx [Integer] Horizontal margins. <%= one_of((-6..6).to_a + [:auto]) %>
     # @param my [Integer] Vertical margins. <%= one_of((-6..6).to_a) %>
-    # @param m [Integer] Padding. <%= one_of((0..6).to_a) %>
-    # @param mt [Integer] Padding left. <%= one_of((0..6).to_a) %>
-    # @param mr [Integer] Padding right. <%= one_of((0..6).to_a) %>
-    # @param mb [Integer] Padding bottom. <%= one_of((0..6).to_a) %>
-    # @param ml [Integer] Padding left. <%= one_of((0..6).to_a) %>
-    # @param mx [Integer] Horizontal padding. <%= one_of((0..6).to_a) %>
-    # @param my [Integer] Vertical padding. <%= one_of((0..6).to_a) %>
+    # @param p [Integer] Padding. <%= one_of((0..6).to_a) %>
+    # @param pt [Integer] Padding left. <%= one_of((0..6).to_a) %>
+    # @param pr [Integer] Padding right. <%= one_of((0..6).to_a) %>
+    # @param pb [Integer] Padding bottom. <%= one_of((0..6).to_a) %>
+    # @param pl [Integer] Padding left. <%= one_of((0..6).to_a) %>
+    # @param px [Integer] Horizontal padding. <%= one_of((0..6).to_a) %>
+    # @param py [Integer] Vertical padding. <%= one_of((0..6).to_a) %>
     #
     # @param position [Symbol] <%= one_of([:relative, :absolute, :fixed]) %>
     #
--- a/app/components/primer/base_component.rb
+++ b/app/components/primer/base_component.rb
@@ -78,7 +78,7 @@ module Primer
     # @param color [Symbol] Text color. <%= one_of([:blue, :red, :gray_light, :gray, :gray_dark, :green, :orange, :orange_light, :purple, :pink, :white, :inherit]) %> Note: this API is subject to change as we move to functional colors.
     # @param bg [String, Symbol] Background color. Accepts either a hex value as a String or a color name as a Symbol.
     #
-    # @param box_shadow [Boolean, Symbol] Box shadow. <%= one_of([true, :medium, :large, :extra_large, :none])
+    # @param box_shadow [Boolean, Symbol] Box shadow. <%= one_of([true, :medium, :large, :extra_large, :none]) %>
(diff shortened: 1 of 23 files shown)


app/components/primer/base_component.rb
```diff
@@ -23,23 +23,35 @@ module Primer
   # <h1 class="mt-0 mt-lg-4 mt-xl-2">Hello world</h1>
   # ```
   #
+  # ## HTML attributes
+  #
+  # System arguments sinclude most HTML attributes. For example:
```
colebemis, line 28:
```suggestion
  # System arguments include most HTML attributes. For example:
```

app/components/primer/base_component.rb
```diff
@@ -23,23 +23,35 @@ module Primer
...
+  #
+  # | Name | Type | Description |
+  # | :- | :- | :- |
+  # | `width` | `Integer` | Width. |
+  # | `height` | `Integer` | Height. |
+  # | `data` | `Hash` | Data attributes. For example: `data: { foo: :bar }` will render `data-foo='bar'`. |
+  # | `title` | `String` | The `title` attribute. |
+  # | `hidden` | `Boolean` | Whether to assign the `hidden` attribute. |
```
colebemis, line 36: I think an `aria` example here could also be helpful (e.g. `aria-label`).

docs/content/system-arguments.md
```diff
@@ -24,33 +24,69 @@ Renders:
...
+| `pl` | `Integer` | Padding left. One of `0`, `1`, `2`, `3`, `4`, `5`, or `6`. |
+| `px` | `Integer` | Horizontal padding. One of `0`, `1`, `2`, `3`, `4`, `5`, or `6`. |
+| `py` | `Integer` | Vertical padding. One of `0`, `1`, `2`, `3`, `4`, `5`, or `6`. |
 | `position` | `Symbol` | One of `:relative`, `:absolute`, or `:fixed`. |
 | `top` | `Boolean` | If `false`, sets `top: 0`. |
 | `right` | `Boolean` | If `false`, sets `right: 0`. |
 | `bottom` | `Boolean` | If `false`, sets `bottom: 0`. |
 | `left` | `Boolean` | If `false`, sets `left: 0`. |
 | `display` | `Symbol` | One of `:block`, `:none`, `:inline`, `:inline_block`, `:table`, or `:table_cell`. |
+| `v` | `Symbol` | Visibility. One of `:hidden` and `:visible`. |
```
kevinsawicki, line 64: This is interesting, I would have expected this to be `visibliity` given their is `display` instead of `d`.

docs/content/system-arguments.md
```diff
@@ -24,33 +24,69 @@ Renders:
...
 | `hide` | `Symbol` | Hide the element at a specific breakpoint. One of `:sm`, `:md`, `:lg`, or `:xl`. |
 | `vertical_align` | `Symbol` | One of `:baseline`, `:top`, `:middle`, `:bottom`, `:text_top`, or `:text_bottom`. |
 | `float` | `Symbol` | One of `:left` and `:right`. |
+| `col` | `Integer` | Number of columns. |
+| `underline` | `Boolean` | Whether text should be underlined. |
+| `color` | `Symbol` | Text color. One of `:blue`, `:red`, `:gray_light`, `:gray`, `:gray_dark`, `:green`, `:orange`, `:orange_light`, `:purple`, `:pink`, `:white`, or `:inherit`. Note: this API is subject to change as we move to functional colors. |
```
kevinsawicki, line 71: So `color` doesn't accept a hex value string like `bg` does?

joelhawksley: You're correct that it does not! But I think it should. What do you think @emplums?

docs/content/system-arguments.md
```diff
@@ -24,33 +24,69 @@ Renders:
...
+| `bg` | `String, Symbol` | Background color. Accepts either a hex value as a String or a color name as a Symbol. |
+| `box_shadow` | `Boolean, Symbol` | Box shadow. <%= one_of([true, :medium, :large, :extra_large, :none]) |
+| `border` | `Symbol` | One of `:left`, `:top`, `:bottom`, `:right`, `:y`, or `:x`. |
+| `border_color` | `Symbol` | One of `:blue`, `:blue_light`, `:gray`, `:gray_dark`, `:green`, `:purple`, `:red`, `:red_light`, `:white`, `:yellow`, or `:black_fade`. Note: this API is subject to change as we move to functional colors. |
+| `border_top` | `Integer` | Set to `0` to remove the top border. |
+| `border_bottom` | `Integer` | Set to `0` to remove the bottom border. |
+| `border_left` | `Integer` | Set to `0` to remove the left border. |
+| `border_right` | `Integer` | Set to `0` to remove the right border. |
 | `font_size` | `String` | One of `00`, `0`, `1`, `2`, `3`, `4`, `5`, or `6`. |
```
kevinsawicki, line 80: Perhaps these values should be quoted since they are strings and not integers.

docs/content/system-arguments.md
kevinsawicki: Same question for `border_color` as well.

docs/content/system-arguments.md
```diff
@@ -24,33 +24,70 @@ Renders:
...
 | `display` | `Symbol` | One of `:block`, `:none`, `:inline`, `:inline_block`, `:table`, or `:table_cell`. |
+| `v` | `Symbol` | Visibility. One of `:hidden` and `:visible`. |
 | `hide` | `Symbol` | Hide the element at a specific breakpoint. One of `:sm`, `:md`, `:lg`, or `:xl`. |
 | `vertical_align` | `Symbol` | One of `:baseline`, `:top`, `:middle`, `:bottom`, `:text_top`, or `:text_bottom`. |
 | `float` | `Symbol` | One of `:left` and `:right`. |
+| `col` | `Integer` | Number of columns. |
+| `underline` | `Boolean` | Whether text should be underlined. |
+| `color` | `Symbol` | Text color. One of `:blue`, `:red`, `:gray_light`, `:gray`, `:gray_dark`, `:green`, `:orange`, `:orange_light`, `:purple`, `:pink`, `:white`, or `:inherit`. Note: this API is subject to change as we move to functional colors. |
+| `bg` | `String, Symbol` | Background color. Accepts either a hex value as a String or a color name as a Symbol. |
+| `box_shadow` | `Boolean, Symbol` | Box shadow. <%= one_of([true, :medium, :large, :extra_large, :none]) |
```
kevinsawicki, line 73: Should this have a `one_of` here? The other ones seems to be inlined.

docs/content/system-arguments.md
joelhawksley: They can be both Strings and Integers. I've updated this doc to reflect that.

docs/content/system-arguments.md
joelhawksley: Oh weird. You're right! ,this should have been executed. Looks like a syntax error 😬

docs/content/system-arguments.md
joelhawksley: I agree. Let's switch it to `visibility` 😄

joelhawksley: Changed!

docs/content/system-arguments.md
joelhawksley: As this is a docs PR, I think we should table this for an issue discussion: https://github.com/primer/view_components/issues/139

colebemis (approved):

PR #140, opened 2020-11-20
Added no limit option to counter

nuthinking:
Some instances can have very high numbers. It feels very wrong to simply put a very arbitrary number when we basically don't want limits.

### DIFF
--- a/app/components/primer/counter_component.rb
+++ b/app/components/primer/counter_component.rb
@@ -32,7 +32,7 @@ module Primer
     )
       @count, @limit, @hide_if_zero, @text, @round, @system_arguments = count, limit, hide_if_zero, text, round, system_arguments
 
-      @has_limit = @limit != nil
+      @has_limit = !@limit.nil?
       @system_arguments[:title] = title
       @system_arguments[:tag] = :span
       @system_arguments[:classes] = class_names(
--- a/app/components/primer/counter_component.rb
+++ b/app/components/primer/counter_component.rb
@@ -16,7 +16,7 @@ module Primer
     #
     # @param count [Integer, Float::INFINITY, nil] The number to be displayed (e.x. # of issues, pull requests)
     # @param scheme [Symbol] Color scheme. One of `SCHEME_MAPPINGS.keys`.
-    # @param limit [Integer, nil] Maximum value to display, pass nil if you don't want any limit. (e.x. if count == 6,000 and limit == 5000, counter will display "5,000+")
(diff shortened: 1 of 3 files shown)


nuthinking:
The documentation would need to be updated...

nuthinking:
Updated also the doc.

app/components/primer/counter_component.rb
```diff
@@ -32,6 +32,7 @@ def initialize(
     )
       @count, @limit, @hide_if_zero, @text, @round, @system_arguments = count, limit, hide_if_zero, text, round, system_arguments
 
+      @has_limit = @limit != nil
```
joelhawksley, line 35:
```suggestion
      @has_limit = !@limit.nil?
```

app/components/primer/counter_component.rb
```diff
@@ -16,7 +16,7 @@ class CounterComponent < Primer::Component
     #
     # @param count [Integer, Float::INFINITY, nil] The number to be displayed (e.x. # of issues, pull requests)
     # @param scheme [Symbol] Color scheme. One of `SCHEME_MAPPINGS.keys`.
-    # @param limit [Integer] Maximum value to display. (e.x. if count == 6,000 and limit == 5000, counter will display "5,000+")
+    # @param limit [Integer, nil] Maximum value to display, pass nil if you don't want any limit. (e.x. if count == 6,000 and limit == 5000, counter will display "5,000+")
```
joelhawksley, line 19:
```suggestion
    # @param limit [Integer, nil] Maximum value to display. Pass `nil` for no limit. (e.x. if `count` == 6,000 and `limit` == 5000, counter will display "5,000+")
```

app/components/primer/counter_component.rb
```diff
@@ -73,16 +74,16 @@ def value
...
         if @round
-          count = [@count.to_i, @limit].min
+          count = @has_limit ? [@count.to_i, @limit].min : @count.to_i
           precision = count.between?(100_000, 999_999) ? 0 : 1
           units = {thousand: "k", million: "m", billion: "b"}
           str = number_to_human(count, precision: precision, significant: false, units: units, format: "%n%u")
         else
           @count = @count.to_i
-          str = number_with_delimiter([@count, @limit].min)
+          str = number_with_delimiter(@has_limit ? [@count, @limit].min : @count)
```
joelhawksley, line 83: Since we're changing the logic for the `round: false` code path, might we add a test for it? I see that our only test is for when `round` is `true`.

joelhawksley (approved):

PR #146, opened 2020-11-24
Adding a Spinner slot to the blankslate component

jonrohan:
This addresses a need from @colebemis and I, we're replacing `gif` spinners on github.com with the new SpinnerComponent. We came across a couple of uses where the `image_src` property was being used to display a loading image in the blankslate component. 

This PR adds a `spinner` slot that passes in the SpinnerComponent directly.

image


app/components/primer/blankslate_component.rb
```diff
@@ -17,6 +21,14 @@ class BlankslateComponent < Primer::Component
     #     description: "Description",
     #   ) %>
     #
+    # @example 150|Loading|Add a [SpinnerComponent](https://primer.style/view-components/components/spinner) to the blankslate in place of an icon.
+    #   <%= render Primer::BlankslateComponent.new(
+    #     title: "Title",
+    #     description: "Description",
+    #   ) do |component| %>
+    #     <% component.slot(:spinner) %>
```
joelhawksley, line 29: Might we provide an example of an argument passed to the spinner, such as the size?

app/components/primer/blankslate_component.rb
```diff
@@ -112,5 +124,17 @@ def initialize(
       @link_text = link_text
       @link_url = link_url
     end
+
+    class Spinner < Primer::Slot
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
```
joelhawksley, line 129: Might we include the arguments for SpinnerComponent as well?

app/components/primer/blankslate_component.html.erb
```diff
@@ -1,13 +1,13 @@
 <%= render Primer::BaseComponent.new(**@system_arguments) do %>
-  <% if @icon.present? %>
+  <% if spinner.present? %>
+    <%= render spinner.component %>
+  <% elsif @icon.present? %>
```
jonrohan, line 4: One small change here @joelhawksley, which I think is a bug fix. You can't have a `spinner`, `@icon`, and a `@image_src` at the same time.

joelhawksley (approved):
Looks great! Mind adding a CHANGELOG entry? We can then ship a release with this change.

app/components/primer/blankslate_component.html.erb
joelhawksley: 👍🏻  I like it.

PR #150, opened 2020-12-10
Fix Ruby warnings

joelhawksley:
This PR fixes a few Ruby warnings that were returned when running the test suite.

It also updates us to bundler `2.2.0`.

### DIFF
--- a/test/components/blankslate_component_test.rb
+++ b/test/components/blankslate_component_test.rb
@@ -36,7 +36,7 @@ class BlankslateComponentTest < Minitest::Test
   end
 
   def test_renders_a_narrow_large_and_spacious_blankslate_component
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -244,4 +245,4 @@ DEPENDENCIES
   yard (~> 0.9.25)
 
 BUNDLED WITH
-   1.17.3
+   2.2.0
```
BlakeWilliams, line 248: Does this change matter?

BlakeWilliams (approved):
one question, otherwise LGTM!

Gemfile.lock
joelhawksley: Good point. Removed.

PR #156, opened 2021-01-12
Update @primer/gatsby-theme-doctocat to use main branch

srt32:
The "Edit this page on GitHub" links in the footers were 404ing because they defaulted to master.

This config change should make the footers point at the right url using `main`

### DIFF
--- a/docs/gatsby-config.js
+++ b/docs/gatsby-config.js
@@ -10,7 +10,7 @@ module.exports = {
       resolve: '@primer/gatsby-theme-doctocat',
       options: {
         defaultBranch: 'main',
-        repoRootPath: 'docs',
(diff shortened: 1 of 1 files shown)


colebemis (approved):

joelhawksley:
@colebemis I actually wonder if we should remove these links altogether, as the docs are dynamically generated from code comments and should not be edited directly. Is that possible?

srt32:
whoop, this didn't quite do what I wanted. My apologies for not making it clear I had not completed testing yet...

The links are still broken but now broken in a new way

colebemis:
@joelhawksley We can add a config option to disable these links 👍

PR #158, opened 2021-01-12
Add flex as a supported display system argument

joshmgross:
Primer View Components can use `display:  :flex` to add the `d-flex` class. https://primer.style/css/utilities/flexbox#css


joshmgross:
Closing in favor of #157

PR #162, opened 2021-01-19
Add docs example for BoxComponent

srt32:
This PR adds a very basic example of using the `BoxComponent`.

### DIFF
--- a/app/components/primer/box_component.rb
+++ b/app/components/primer/box_component.rb
@@ -3,7 +3,7 @@
 module Primer
   # A basic wrapper component for most layout related needs.
   class BoxComponent < Primer::Component
-    # @example Default
+    # @example 40|Default
     #   <%= render(Primer::BoxComponent.new) %>
     #
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
--- a/app/components/primer/box_component.rb
+++ b/app/components/primer/box_component.rb
@@ -3,8 +3,8 @@
 module Primer
   # A basic wrapper component for most layout related needs.
   class BoxComponent < Primer::Component
-    # @example 40|Default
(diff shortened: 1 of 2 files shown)


srt32:
Not the prettiest example but I think better than nothing!

image

colebemis (approved):
🎉

colebemis:
I think an example using system arguments could also be helpful.

srt32:
> I think an example using system arguments could also be helpful.

I am still learning what those are ;). Do you have an example in particular that you like?

colebemis:
Maybe an example with a background color and padding?

```rb
<%= render(Primer::BoxComponent.new(bg: :gray, p: 3)) do %>
  Hello world
<% end %>
```

app/components/primer/box_component.rb
```diff
@@ -3,6 +3,12 @@
 module Primer
   # A basic wrapper component for most layout related needs.
   class BoxComponent < Primer::Component
+    # @example 20|Default
+    #   <%= render(Primer::BoxComponent.new) { "Your content here" } %>
+    #
+    # @example 20|Color and padding
```
colebemis, line 9: The height of the example will need to be larger because of the extra padding :)

```suggestion
    # @example 54|Color and padding
```

srt32:
@colebemis hmm, now things are scrolling:

image

colebemis:
@srt32 Hmm, I think you have to regenerate the docs. Looks like the `iframe` height wasn't updated: https://github.com/primer/view_components/pull/162/files#diff-271261c1a92ba2df62f80da7c9990a2508adb0547c698fa32d21d9c9c6fc1045R21

srt32:
@colebemis ready to go!

PR #164, opened 2021-01-25
Add log statement on docs:build when component has no examples

srt32:
We'd like to make sure all components have documented examples.

This change adds a log statement to the end of `docs:build` that will
list out the components missing examples.

If I remove the examples from `BoxComponent`, for example, I'll see this
statement:

```
Converting YARD documentation to Markdown files.
Markdown compiled.
The following components have no examples defined: Primer::BoxComponent. Consider adding an example?
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ rails_version = "#{ENV['RAILS_VERSION'] || '6.0.3.3'}"
 
 gem "rake", "~> 12.0"
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem "rails", rails_version == "main" ? { github: "rails/rails" } : rails_version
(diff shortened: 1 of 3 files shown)


joelhawksley (approved):

Gemfile
```diff
@@ -7,7 +7,7 @@ rails_version = "#{ENV['RAILS_VERSION'] || '6.0.3.3'}"
 
 gem "rake", "~> 12.0"
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem "rails", rails_version == "main" ? { github: "rails/rails" } : rails_version
```
srt32, line 10: I think `bundler` is defaulting to `master` instead of the default branch so we have to be explicit here. Changing to `git` and the full https path uses the secure approach so we can do away with the warning we were seeing.

Issue #155, opened 2021-01-11
Measure and reduce the number of allocations made by Primer::Classify


BlakeWilliams:
When investigating a page's performance related to allocations, I found that `Primer::Classify` was performing a significant number of allocations.

The places making the most allocations were:

- [Primer::Classify#extract_class](https://github.com/primer/view_components/blob/main/lib/primer/classify.rb#L155)
- `Hash#except`
- [Line 187 of Primer::Classify#extract_class](https://github.com/primer/view_components/blob/main/lib/primer/classify.rb#L187)

We should try to measure, then reduce the number of allocations made in that class since it can be a hot path for primer components.

manuelpuyol:
@BlakeWilliams with https://github.com/primer/view_components/pull/197 and https://github.com/primer/view_components/pull/166

should we close this issue?

BlakeWilliams:
I think we're probably good to close this for now, but can probably write some follow-up issues to prevent regressions, catch new keys, refactor, etc.

Thanks for the reminder!

PR #166, opened 2021-01-26
Add test to limit allocations of `Primer::Classify`

BlakeWilliams:
Reduce allocations from 97 to 38 when using a "basic" `Primer::Classify#call` implementation.

closes part of https://github.com/primer/view_components/issues/155

### DIFF
--- a/lib/primer/classify.rb
+++ b/lib/primer/classify.rb
@@ -154,7 +154,8 @@ module Primer
       # Example usage:
       # extract_hash({ mt: 4, py: 2 }) => "mt-4 py-2"
       def extract_hash(styles_hash)
-        out = styles_hash.each_with_object({ classes: [], styles: [] }) do |(key, value), memo|
+        memo = { classes: [], styles: [] }
+        styles_hash.each do |key, value|
           next unless VALID_KEYS.include?(key)
 
           if value.is_a?(Array) && !RESPONSIVE_KEYS.include?(key)
@@ -245,9 +246,9 @@ module Primer
         end
 
         {
-          classes: out[:classes].join(" "),
-          styles: out[:styles].join(" ")
-        }.merge(out.except(:classes, :styles))
+          classes: memo[:classes].join(" "),
+          styles: memo[:styles].join(" ")
+        }.merge(memo.except(:classes, :styles))
       end
     end
   end
--- a/lib/primer/classify.rb
+++ b/lib/primer/classify.rb
@@ -113,10 +113,12 @@ module Primer
       def call(classes: "", style: nil, **args)
         extracted_results = extract_hash(args)
 
-        {
-          class: [validated_class_names(classes), extracted_results[:classes]].compact.join(" ").presence,
-          style: [extracted_results[:styles], style].compact.join("").presence,
-        }.merge(extracted_results.except(:classes, :styles))
+        extracted_results[:class] = [validated_class_names(classes), extracted_results[:classes]].compact.join(" ").presence
+        extracted_results[:style] = [extracted_results[:styles], style].compact.join("").presence
+        extracted_results.delete(:classes)
+        extracted_results.delete(:styles)
+
+        extracted_results
       end
 
       private
--- a/lib/primer/classify.rb
+++ b/lib/primer/classify.rb
@@ -3,7 +3,6 @@
 module Primer
   class Classify
     MARGIN_DIRECTION_KEYS = [:mt, :ml, :mb, :mr]
-    MARGIN_DIRECTION_KEY_VALUES = [:mt, :ml, :mb, :mr].map { |k| [k, k.to_s.dasherize] }.to_h.freeze
     SPACING_KEYS = ([:m, :my, :mx, :p, :py, :px, :pt, :pl, :pb, :pr] + MARGIN_DIRECTION_KEYS).freeze
     DIRECTION_KEY = :direction
     JUSTIFY_CONTENT_KEY = :justify_content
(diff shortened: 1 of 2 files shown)


BlakeWilliams:
I ran some benchmarks testing the new classify vs the classify on `main` and it looks like we're ~1.6x faster too!

```
primer_view_components λ bundle exec rake benchmark
/Users/blake/.rbenv/versions/2.7.1/bin/ruby ./performance/benchmark.rb
            classify     5.291k i/100ms
        old classify     3.188k i/100ms
            classify     51.519k (± 6.5%) i/s -    259.259k in   5.053442s
        old classify     32.399k (± 4.7%) i/s -    162.588k in   5.029117s

Comparison:
            classify:    51519.4 i/s
        old classify:    32399.3 i/s - 1.59x  (± 0.00) slower
```

lib/primer/classify.rb
```diff
@@ -152,98 +159,102 @@ def validated_class_names(classes)
...
-              end
-            elsif key == VISIBILITY_KEY
-              memo[:classes] << "v-#{dasherized_val}"
-            else
-              memo[:classes] << "#{key.to_s.dasherize}#{breakpoint}-#{dasherized_val}"
-            end
+        memo
+      end
+
+      def extract_value(memo, key, val, breakpoint)
```
BlakeWilliams, line 182: I kept these names the same as from before the refactor, but happy to update them to be more descriptive.

iancanderson (approved):

lib/primer/classify.rb
```diff
@@ -152,98 +159,102 @@ def validated_class_names(classes)
       # Example usage:
       # extract_hash({ mt: 4, py: 2 }) => "mt-4 py-2"
       def extract_hash(styles_hash)
-        out = styles_hash.each_with_object({ classes: [], styles: [] }) do |(key, value), memo|
+        memo = { classes: [], styles: String.new }
```
manuelpuyol, line 162: any reason to use `String.new` instead of `""`?

lib/primer/classify.rb
```diff
@@ -152,98 +159,102 @@ def validated_class_names(classes)
...
-        }.merge(out.except(:classes, :styles))
+        if BOOLEAN_MAPPINGS.has_key?(key)
+          BOOLEAN_MAPPINGS[key][:mappings].map { |m| m[:css_class] if m[:value] == val }.compact.each do |css_class|
+            memo[:classes] << css_class
+          end
+        elsif key == BG_KEY
+          if val.to_s.starts_with?("#")
+            memo[:styles] << "background-color: #{val};"
+          else
+            memo[:classes] << "bg-#{val.to_s.dasherize}"
```
manuelpuyol, line 200: should we have a helper:
```rb
def dasherize(val)
  val.to_s.dasherize
end
```
?

test/primer/classify_test.rb
```diff
@@ -282,11 +282,35 @@ def test_does_not_raise_error_when_passing_in_a_primer_css_class_otherwise
     assert_generated_class("bg-blue text-center float-left ml-1 ",  { classes: "bg-blue text-center float-left ml-1" })
   end
 
+  def test_limits_allocations
+    # Warm up allocations
+    values = { align_self: :center, width: :fit, p: 4, m: 1, border: :top, box_shadow: true, color: :red, visibility: :hidden}
```
manuelpuyol, line 287: are we using **all** available keys here?

lib/primer/classify.rb
BlakeWilliams: `String.new` lets me mutate the string which we do below, but `""` is frozen due to `# frozen_string_literal: true`

test/primer/classify_test.rb
BlakeWilliams: I'm not, but that's not a bad idea.

I don't want to spend a ton more time on this specific PR, but that could be a good follow-up.

manuelpuyol: my idea here is that if we test all keys, we could catch regressions easier in the future :)

test/primer/classify_test.rb
```diff
@@ -282,11 +282,35 @@ def test_does_not_raise_error_when_passing_in_a_primer_css_class_otherwise
...
+    Primer::Classify.call(**values)
+
+    assert_allocations 38, within: 4 do
```
joelhawksley, line 290: Where'd this value of `4` come from? Does this value vary a little in local test vs. CI?

lib/primer/classify.rb
BlakeWilliams: Good question, I'm leaning slightly towards keeping it as-is, since it's calling `to_s.dasherize` and not doing a more complex operation.

test/primer/classify_test.rb
BlakeWilliams: You've convinced me, I was thinking of keeping it as-is so we could keep track of which aren't optimized and which are, but tracking regressions is more important. I'll add a bunch more keys to this call and bump the allocation values.

test/primer/classify_test.rb
BlakeWilliams: I should probably add a comment, but I believe the difference is between Ruby and Rails versions, since we run a matrix of versions in CI.

The other thought was to make a build specifically for testing allocations against 2.7, but I thought I'd submit this as-is and get thoughts/feedback first.

BlakeWilliams:
Updated the test to include all, or most of the keys we support. Running the test against `main` resulted in `204` allocations, this branch resulted in `87` allocations.

joelhawksley (approved):

PR #167, opened 2021-01-26
Document the status of a given component

srt32:
By default, a component has an experimental status and now we have the ability to
declare a status on each component, which gets rendered in the docs.

We can use this system to better inform consumers what they can expect
of a given component.

Screen Shot 2021-01-26 at 12 46 15 PM

### DIFF
--- a/app/components/primer/box_component.rb
+++ b/app/components/primer/box_component.rb
@@ -20,7 +20,7 @@ module Primer
     end
 
     def self.status
-      :released
+      Primer::Component::STATUSES[:released]
     end
   end
 end
--- a/app/components/primer/box_component.rb
+++ b/app/components/primer/box_component.rb
@@ -20,7 +20,7 @@ module Primer
     end
 
     def self.status
-      Primer::Component::STATUSES[:released]
+      Primer::Component::STATUSES[:stable]
     end
   end
 end
--- a/app/components/primer/box_component.rb
+++ b/app/components/primer/box_component.rb
@@ -18,5 +18,9 @@ module Primer
     def call
       render(Primer::BaseComponent.new(**@system_arguments)) { content }
     end
+
+    def self.status
+      :released
+    end
   end
 end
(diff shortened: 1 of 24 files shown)


colebemis:
👋 This is awesome! Fyi, Doctocat supports a `status` frontmatter variable that might come in handy here: https://primer.style/doctocat/usage/front-matter#status

app/components/primer/component.rb
```diff
@@ -7,5 +7,15 @@ class Component < ViewComponent::Base
     include FetchOrFallbackHelper
     include OcticonsHelper
     include JoinStyleArgumentsHelper
+
+    STATUSES = [
+      :experimental,
+      :beta,
+      :recommended,
+      :released,
+    ]
```
manuelpuyol, line 16: nit: I actually prefer using a hash here


```rb
STATUSES = {
  experimental: :experimental,
  ...
```

and in the class

```rb
def self.status
  Primer::Component::STATUSES[:experimental]
end
```

manuelpuyol: another idea would be transforming `status` into a DSL so we could add validations to ensure no invalid status are being used

srt32:
@manuelpuyol @colebemis ready for another 👁️

manuelpuyol (approved):
lgtm!

colebemis (approved):

app/components/primer/box_component.rb
```diff
@@ -18,5 +18,9 @@ def initialize(**system_arguments)
     def call
       render(Primer::BaseComponent.new(**@system_arguments)) { content }
     end
+
+    def self.status
```
joelhawksley, line 22: FWIW, I think we could achieve this via YARD as well, perhaps via the `version` tag?

See https://www.rubydoc.info/gems/yard/file/docs/Tags.md#adding-custom-tags.

That being said, I think it's cool to have this status available programmatically 👍🏻

srt32: I started with `tags` but got confused ;). Happy to revisit this if it's better / easier.

joelhawksley: 🤷🏻‍♂️ I think they're equivalent at this point. Just wanted to note this in case we have any issues with this approach.

PR #169, opened 2021-01-26
Add instructions for running the docs site locally

srt32:


CONTRIBUTING.md
```diff
@@ -22,11 +22,19 @@ Please note that this project is released with a [Contributor Code of Conduct](C
...
 - Write tests.
+- Keep your change as focused as possible. If there are multiple changes you would like to make that are not dependent upon each other, consider submitting them as separate pull requests.
 - Write [YARD comments](https://yardoc.org/) for component and slot initializers. Please document the purpose and general use of new components or slots you add, as well as the parameters they accept. See for example the comment style in app/components/primer/counter_component.rb.
 - Add new components to the `components` list in Rakefile in the `docs:build` task, so that Markdown documentation is generated for them within docs/content/components/.
-- Keep your change as focused as possible. If there are multiple changes you would like to make that are not dependent upon each other, consider submitting them as separate pull requests.
 - Write a [good commit message](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
 
+## Previewing docs changes
+
+- Run `bundle exec rake docs:build`
```
joelhawksley, line 32: I _think_ this is the same as `script/dev`. Right @manuelpuyol?

srt32: I see `bundle exec rake docs:livereload` in `Procfile` but don't see it actually spinning up the gatsby site. Maybe we update it do to that?

manuelpuyol: @srt32 `script/dev` should startup everything, including the docs:
https://github.com/primer/view_components/blob/de077dd7f7827f60e943f2e27b07450598d30b1c/Procfile#L5

you should be able to find them at `localhost:5400`

srt32: Maybe we add a log statement somewhere so this is more discoverable? I couldn't figure it out ;)

manuelpuyol: hmm, it should log! mine was like the last thing logged after I ran the script

PR #174, opened 2021-01-28
Truncate component

manuelpuyol:
Creates a `TruncateComponent` based on https://primer.style/css/components/truncate

image


Rakefile
```diff
@@ -110,6 +110,7 @@ namespace :docs do
       Primer::SpinnerComponent,
       Primer::SubheadComponent,
       Primer::TextComponent,
+      Primer::TruncateComponent,
       Primer::TimelineItemComponent
```
srt32, line 114:
```suggestion
      Primer::TimelineItemComponent,
      Primer::TruncateComponent
```

if we want :abc:

app/components/primer/truncate_component.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Primer
+  # The Truncate component is a wrapper component that will shorten text with an ellipsis.
```
joelhawksley, line 4:
```suggestion
  # Use TruncateComponent to shorten overflowing text with an ellipsis.
```

app/components/primer/truncate_component.rb
```diff
@@ -0,0 +1,36 @@
...
+    #   </div>
+    #
+    # @example 25|Inline
+    #   <%= render(Primer::TruncateComponent.new(tag: :span, inline: true)) { "branch-name-that-is-really-long" } %>
+    #
+    # @example 25|Expandable
+    #   <%= render(Primer::TruncateComponent.new(tag: :span, inline: true, expandable: true)) { "branch-name-that-is-really-long" } %>
+    #
+    # @param inline [Boolean] Whether the element is inline (or inline-block).
+    # @param expandable [Boolean] Whether the entire string should be revealed on hover. Can only be used in conjuction with inline.
```
joelhawksley, line 18:
```suggestion
    # @param expandable [Boolean] Whether the entire string should be revealed on hover. Can only be used in conjunction with `inline`.
```

app/components/primer/truncate_component.rb
```diff
@@ -0,0 +1,36 @@
...
+    #     <%= render(Primer::TruncateComponent.new(tag: :p)) { "branch-name-that-is-really-long" } %>
+    #   </div>
+    #
+    # @example 25|Inline
+    #   <%= render(Primer::TruncateComponent.new(tag: :span, inline: true)) { "branch-name-that-is-really-long" } %>
+    #
+    # @example 25|Expandable
+    #   <%= render(Primer::TruncateComponent.new(tag: :span, inline: true, expandable: true)) { "branch-name-that-is-really-long" } %>
+    #
+    # @param inline [Boolean] Whether the element is inline (or inline-block).
```
joelhawksley, line 17: What do you think about supporting `max_width`, as Primer React does? https://primer.style/components/Truncate

manuelpuyol: I thought about it but wasn't sure if we should have a `max_width` prop or let the user add a `style` tag

joelhawksley: Hmm. I think it might be nice to declare it as an argument so we can document the use case.

We have some rules that build up a `style` attribute: https://github.com/primer/view_components/blob/main/lib/primer/classify.rb#L200

manuelpuyol: done!

app/components/primer/truncate_component.rb
```diff
@@ -14,10 +14,14 @@ class TruncateComponent < Primer::Component
     # @example 25|Expandable
     #   <%= render(Primer::TruncateComponent.new(tag: :span, inline: true, expandable: true)) { "branch-name-that-is-really-long" } %>
     #
+    # @example 25|Custom size
+    #   <%= render(Primer::TruncateComponent.new(tag: :span, inline: true, expandable: true, max_width: 100)) { "branch-name-that-is-really-long" } %>
+    #
     # @param inline [Boolean] Whether the element is inline (or inline-block).
     # @param expandable [Boolean] Whether the entire string should be revealed on hover. Can only be used in conjunction with `inline`.
+    # @param max_width [Integer] Sets the max-width of the text
```
joelhawksley, line 22:
```suggestion
    # @param max_width [Integer] Sets the max-width of the text.
```

joelhawksley (approved):

PR #175, opened 2021-01-29
Add animation classes to Primer::Classify

manuelpuyol:
Adds option to animate components according to https://primer.style/css/utilities/animations !
Supports `:fade_in`, `:fade_out`, `:fade_up`, `:fade_down`, `:scale_in`, `:pulse`, `:grow_x`, and `:grow` (hover animation).

Example:

```rb
<%= render Primer::AvatarComponent(alt: "an avatar", src: "some src", animation: :fade_in) %>
```

Obs:

I'm working on a storybook to showcase our classify options, but https://github.com/jonspalmer/view_component_storybook/issues/19 needs to be fixed first


emplums (approved):
This looks good!

One thing to note: I am not 100% sure if we will be able to replicate adding animations via props in Primer React Components, we might need to make an `Animation` component or implement it via a Hook that returns a props getter. I would love to find a way to follow this API so that we can have consistency between implementations but, animating elements when they enter/leave the DOM in React is a bit tricky.

Issue #90, opened 2020-09-22
Handle `nil` cases in storybook


natashaU:
While working on the label component: https://github.com/primer/view_components/blob/main/app/components/primer/label_component.rb

I noticed that there was an error when implementing introducing a storybook story. '

The label component has the following:

```
SCHEME_MAPPINGS = {
      # gray
      gray: "Label--gray",
      dark_gray: "Label--gray-darker",

      # colored
      yellow: "Label--yellow",
      orange: "Label--orange",
      red: "Label--red",
      green: "Label--green",
      blue: "Label--blue",
      purple: "Label--purple",
      pink: "Label--pink",

      # Deprecated
      outline: "Label--outline",
      green_outline: "Label--outline-green",
    }.freeze
    SCHEME_OPTIONS = SCHEME_MAPPINGS.keys << nil
```

However, the following code in the story is unable to handle nil cases:
`select(:scheme, Primer::StoriesHelper.array_to_options(Primer::LabelComponent::SCHEME_MAPPINGS.keys), :blue)`

That's because nil shows up as ` :nil ` a hash without a key. 

cc/  @joelhawksley @aellispierce @jonspalmer

jonspalmer:
IMO we should just be using arrays as the options. https://github.com/jonspalmer/view_component_storybook supports this as of version 0.5.0 and eliminates the need for `Primer::StoriesHelper.array_to_options`

manuelpuyol:
as soon as https://github.com/primer/view_components/pull/89 merges I'll update stories definitions :)

manuelpuyol:
@jonspalmer arrays with `nil` as an option won't work

```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at stringifyObject (/Users/manuelpuyol/primer/view_components/demo/node_modules/@storybook/server/dist/lib/compiler/stringifier.js:130:16)
    at /Users/manuelpuyol/primer/view_components/demo/node_modules/@storybook/server/dist/lib/compiler/stringifier.js:120:14
    at Array.map (<anonymous>)
    at stringifyObject (/Users/manuelpuyol/primer/view_components/demo/node_modules/@storybook/server/dist/lib/compiler/stringifier.js:119:31)
    at /Users/manuelpuyol/primer/view_components/demo/node_modules/@storybook/server/dist/lib/compiler/stringifier.js:132:21
    at Array.map (<anonymous>)
[...]

PR #176, opened 2021-01-29
Add TooltipComponent

srt32:
This change adds a `TooltipComponent`, which is inspired by https://primer.style/css/components/tooltips.

image

image


app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+module Primer
+  # The Tooltip component is a wrapper component that will apply a tooltip
+  class TooltipComponent < Primer::Component
+    DIRECTION_DEFAULT = ""
+    ALIGNMENT_DEFAULT = ""
```
manuelpuyol, line 7: for empty defaults we tend to use `nil`

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,77 @@
...
+    MULTILINE_DEFAULT = false
+    DELAY_DEFAULT = true
+
+    DIRECTION_OPTIONS = [DIRECTION_DEFAULT] + %w(
```
manuelpuyol, line 11: why strings and not symbols?

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,77 @@
...
+    #   <%= render(Primer::TooltipComponent.new(label: "Even bolder", direction: "nw")) { "Bold Text" } %>
+    #
+    # @example 50|With an alignment
+    #   <%= render(Primer::TooltipComponent.new(label: "Even bolder", alignment: "right-1")) { "Bold Text" } %>
+    #
+    # @example 50|Without a delay
+    #   <%= render(Primer::TooltipComponent.new(label: "Even bolder", delay: false)) { "Bold Text" } %>
+    #
+    # @param label [String] the text to appear in the tooltip
+    # @param direction [String] Specify the direction of a tooltip with north, south, east, and west directions
```
manuelpuyol, line 46:
```suggestion
    # @param direction [String] Direction of the tooltip. <%= one_of(Primer:: TooltipComponent:: DIRECTION_OPTIONS) %>
```

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,77 @@
...
+    )
+      @system_arguments = system_arguments
+      @system_arguments[:tag] ||= :span
+      @system_arguments[:aria] = { label: label }
+
+      @system_arguments[:classes] = class_names(
+        @system_arguments[:classes],
+        "tooltipped",
+        "tooltipped-#{fetch_or_fallback(DIRECTION_OPTIONS, direction, DIRECTION_DEFAULT)}" => direction.present?,
+        "tooltipped-align-#{fetch_or_fallback(ALIGNMENT_OPTIONS, alignment, ALIGNMENT_DEFAULT)}" => alignment.present?,
```
manuelpuyol, line 67: since you check for presence, I think the defaults don't need to be empty, we could enforce a default

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,77 @@
...
+    # @param alignment [String] Align tooltips to the left or right of an element, combined with a directional class to specify north or south
+    # @param multiline [Boolean] Use this when you have long content
+    # @param delay [Boolean] By default the tooltips have a slight delay before appearing. Set false to override this
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    def initialize(
+      label:,
+      direction: DIRECTION_DEFAULT,
+      alignment: ALIGNMENT_DEFAULT,
+      multiline: ALIGNMENT_DEFAULT,
+      delay: DELAY_DEFAULT,
```
joelhawksley, line 56: What do you think about calling this `no_delay`, to match the Primer React component? https://primer.style/components/Tooltip#component-props

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,77 @@
...
+    # @param label [String] the text to appear in the tooltip
+    # @param direction [String] Specify the direction of a tooltip with north, south, east, and west directions
+    # @param alignment [String] Align tooltips to the left or right of an element, combined with a directional class to specify north or south
+    # @param multiline [Boolean] Use this when you have long content
+    # @param delay [Boolean] By default the tooltips have a slight delay before appearing. Set false to override this
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    def initialize(
+      label:,
+      direction: DIRECTION_DEFAULT,
+      alignment: ALIGNMENT_DEFAULT,
```
joelhawksley, line 54: Same comment, re: argument naming. Maybe we could use `align` here?

app/components/primer/tooltip_component.rb
srt32: It was easier to copy / paste ;) Happy to convert to strings if we'd like.

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,77 @@
...
+    #
+    # @example 50|With an alignment
+    #   <%= render(Primer::TooltipComponent.new(label: "Even bolder", alignment: "right-1")) { "Bold Text" } %>
+    #
+    # @example 50|Without a delay
+    #   <%= render(Primer::TooltipComponent.new(label: "Even bolder", delay: false)) { "Bold Text" } %>
+    #
+    # @param label [String] the text to appear in the tooltip
+    # @param direction [String] Direction of the tooltip. <%= one_of(Primer:: TooltipComponent:: DIRECTION_OPTIONS) %>
+    # @param alignment [String] Align tooltips to the left or right of an element, combined with a directional class to specify north or south. <%= one_of(Primer::TooltipComponent::ALIGNMENT_OPTIONS) %>
```
joelhawksley, line 47:
```suggestion
    # @param alignment [String] Align tooltips to the left or right of an element, combined with a `direction` to specify north or south. <%= one_of(Primer::TooltipComponent::ALIGNMENT_OPTIONS) %>
```

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,77 @@
...
+    #   <%= render(Primer::TooltipComponent.new(label: "Even bolder", direction: "nw")) { "Bold Text" } %>
+    #
+    # @example 50|With an alignment
+    #   <%= render(Primer::TooltipComponent.new(label: "Even bolder", alignment: "right-1")) { "Bold Text" } %>
+    #
+    # @example 50|Without a delay
+    #   <%= render(Primer::TooltipComponent.new(label: "Even bolder", delay: false)) { "Bold Text" } %>
+    #
+    # @param label [String] the text to appear in the tooltip
+    # @param direction [String] Direction of the tooltip. <%= one_of(Primer:: TooltipComponent:: DIRECTION_OPTIONS) %>
```
joelhawksley, line 46:
```suggestion
    # @param direction [String] Direction of the tooltip. <%= one_of(Primer::TooltipComponent::DIRECTION_OPTIONS) %>
```

app/components/primer/tooltip_component.rb
joelhawksley: I'd vote for symbols as well ❤️

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+module Primer
+  # The Tooltip component is a wrapper component that will apply a tooltip
```
joelhawksley, line 4:
```suggestion
  # The Tooltip component is a wrapper component that will apply a tooltip to the provided content.
```

app/components/primer/tooltip_component.rb
srt32: What would this look like? I'm still getting used to the fetch / fallback API

manuelpuyol: I'd say we could set `DIRECTION_DEFAULT` to `:n` for example

stories/primer/tooltip_component_stories.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Primer::TooltipComponentStories < ViewComponent::Storybook::Stories
+  layout "storybook_preview"
```
srt32, line 4: This is currently crashing locally, we think because of https://github.com/primer/view_components/issues/90. I'm looking into it.

app/components/primer/tooltip_component.rb
```diff
@@ -8,27 +8,27 @@ class TooltipComponent < Primer::Component
...
-    )
+    ALIGNMENT_OPTIONS = [ALIGNMENT_DEFAULT] + [
+      :"right-1",
+      :"left-1",
+      :"right-2",
+      :"left-2",
+      :"right-1",
+      :"left-1",
+      :"right-2",
+      :"left-2",
```
joelhawksley: Perhaps we could underscore these values to avoid the quotes?

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,81 @@
...
+    # @example 100|With an alignment
+    #   <div class="pt-5">
+    #     <%= render(Primer::TooltipComponent.new(label: "Even bolder", direction: :s, alignment: :right_1)) { "Bold Text With an Alignment" } %>
+    #   </div>
+    #
+    # @example 100|Without a delay
+    #   <div class="pt-5">
+    #     <%= render(Primer::TooltipComponent.new(label: "Even bolder", direction: :s, no_delay: true)) { "Bold Text without a delay" } %>
+    #   </div>
+    #
```
manuelpuyol, line 48: I'd like to see an example adding tooltips to things other than text 😉 

```suggestion
    #
    # @example 100|Wrapping another component
    #   <div class="pt-5">
    #     <%= render(Primer::TooltipComponent.new(label: "Even bolder")) do %>
    #       <%= render(Primer::ButtonComponent.new) { "Bold Button" } %>
    #     <% end %>
    #   </div>
    #
```

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,81 @@
...
+    # @param label [String] the text to appear in the tooltip
+    # @param direction [String] Direction of the tooltip. <%= one_of(Primer::TooltipComponent::DIRECTION_OPTIONS) %>
+    # @param align [String] Align tooltips to the left or right of an element, combined with a `direction` to specify north or south. <%= one_of(Primer::TooltipComponent::ALIGN_MAPPING.keys - [Primer::TooltipComponent::ALIGN_DEFAULT]) %>
```
manuelpuyol, line 51: I don't think we have to hide `Primer::TooltipComponent::ALIGN_DEFAULT`

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,81 @@
...
+    )
+      @system_arguments = system_arguments
+      @system_arguments[:tag] ||= :span
+      @system_arguments[:aria] = { label: label }
+
+      @system_arguments[:classes] = class_names(
+        @system_arguments[:classes],
+        "tooltipped",
+        "tooltipped-#{fetch_or_fallback(DIRECTION_OPTIONS, direction, DIRECTION_DEFAULT)}",
+        ALIGN_MAPPING[fetch_or_fallback(ALIGN_MAPPING.keys, align, :default)],
```
manuelpuyol, line 71:
```suggestion
        ALIGN_MAPPING[fetch_or_fallback(ALIGN_MAPPING.keys, align, ALIGN_DEFAULT)],
```

manuelpuyol (approved):
a few details, otherwise lgtm

app/components/primer/tooltip_component.rb
srt32: Added in 84871ff!

app/components/primer/tooltip_component.rb
```diff
@@ -0,0 +1,88 @@
...
+    #
+    # @example 100|With an alignment
+    #   <div class="pt-5">
+    #     <%= render(Primer::TooltipComponent.new(label: "Even bolder", direction: :s, alignment: :right_1)) { "Bold Text With an Alignment" } %>
+    #   </div>
+    #
+    # @example 100|Without a delay
+    #   <div class="pt-5">
+    #     <%= render(Primer::TooltipComponent.new(label: "Even bolder", direction: :s, no_delay: true)) { "Bold Text without a delay" } %>
+    #   </div>
```
joelhawksley, line 54: @srt32 as a quick follow-up, might we pass `pt: 5` to these components instead of having `pt-5` the wrapping div?

srt32: With that change, we still can't see the tooltip with the iframe as is.

```
     # @example 100|Default
-    #   <div class="pt-5">
-    #     <%= render(Primer::TooltipComponent.new(label: "Even bolder")) { "Default Bold Text" } %>
-    #   </div>
+    #   <%= render(Primer::TooltipComponent.new(label: "Even bolder", pt: 5)) { "Default Bold Text" } %>
     #
```


image

Am I missing something there?

manuelpuyol: maybe adding `mt-5` or even both `mt-5` and `pt-5` may solve it?

PR #178, opened 2021-02-01
Add docs for HeadingComponent and UnderlineNavComponent

srt32:
This change takes a first pass at adding docs to some of the components that are
missing them.

* HeadingComponent
* UnderlineNavComponent

image

image


manuelpuyol (approved):

docs/content/components/subhead.md
```diff
@@ -77,14 +77,20 @@ Use the Subhead component for page headings.
 | `danger` | `Boolean` | `false` | Whether to style the heading as dangerous. |
 | `system_arguments` | `Hash` | N/A | [System arguments](/system-arguments) |
 
+:nodoc
```
srt32, line 80: @manuelpuyol we expect all these nodoc strings?

manuelpuyol: we do, I added that comment because of `rubocop`
those are undocumented slots, so we should work on them

PR #180, opened 2021-02-02
Add docs and stories for DropdownMenu component

srt32:
This change adds docs and a storybook entry for the
DropdownMenuComponent.

We really should add a `DropdownComponent` that wraps this existing
component via a slot but for now we are targeting full docs so we will
commit this as at this point.

image

image


manuelpuyol (approved):
lgtm for now
once we implement a DropdownComponent we'll have to move some stuff around

PR #182, opened 2021-02-02
Add docs for DetailsComponent

srt32:
image


joelhawksley:
Screen Shot 2021-02-02 at 1 29 08 PM

🤔 not sure what's going on here, but at least on the Vercel preview, the examples here are not styled.

app/components/primer/details_component.rb
```diff
@@ -1,11 +1,7 @@
 # frozen_string_literal: true
 
 module Primer
-  #
-  # overlay - options are `none`, `default` and `dark`. Dictates the type of overlay to render with.
-  # button - options are `default` and `reset`. default will make the target a default primer ``.btn`
-  #          reset will remove all styles from the <summary> element.
-  #
+  # Details classes are created to enhance the native behaviors of the details element
```
joelhawksley, line 4: How about:

```suggestion
  # Use DetailsComponent to reveal content after clicking a button.
```

Or something similar?

srt32:
Interesting! It works as expected in chrome but not safari. I'm looking...

srt32:
I am seeing "shadow content" in safari, which I assume is the problem here but I'm not sure why it's happening. Is that what we expect?

image

joelhawksley:
> I am seeing "shadow content" in safari, which I assume is the problem here but I'm not sure why it's happening. Is that what we expect? 

I'm honestly not sure. @colebemis have you seen anything like this before?

keithamus:
> I am seeing "shadow content" in safari, which I assume is the problem here but I'm not sure why it's happening. Is that what we expect?

That’s the browsers’ shadowdom implementation of the summary element. Nothing to worry about, just Safari devtools exposing their own internals. This isn’t related to the underlying issue, I believe.

srt32:
b763ac1 removes the examples for now until we can resolve the safari rendering issues.

docs/content/components/timelineitem.md
```diff
@@ -41,15 +41,21 @@ Use `TimelineItem` to display items on a vertical timeline, connected by badge e
 | `square` | `Boolean` | `true` | Whether to round the edges of the image. |
 | `system_arguments` | `Hash` | N/A | [System arguments](/system-arguments) |
 
+:nodoc
```
joelhawksley, line 44: Might we omit these `nodocs` from the markdown output of the Rakefile?

srt32: @manuelpuyol ^ thoughts?

srt32: I resolved this problem for now by documenting the slots classes too ;)

srt32:
Ready for one more :eye: here

app/components/primer/details_component.rb
```diff
@@ -33,8 +32,10 @@ def render?
       summary.present? && body.present?
     end
 
-    # :nodoc
+    # Use the Summary slot  as the trigger button
```
joelhawksley, line 35:
```suggestion
    # Use the Summary slot as a trigger to reveal the content.
```

app/components/primer/details_component.rb
```diff
@@ -33,8 +32,10 @@ def render?
...
     class Summary < Primer::Slot
+      # @param button [Boolean] If there should be a button or not
```
joelhawksley, line 37:
```suggestion
      # @param button [Boolean] Whether to render the Summary as a button or not.
```

app/components/primer/details_component.rb
```diff
@@ -50,8 +51,9 @@ def component
       end
     end
 
-    # :nodoc
+    # Use the Body slot as the main content to be shown when triggered by the Summary
```
joelhawksley, line 54:
```suggestion
    # Use the Body slot as the main content to be shown when triggered by the Summary.
```

srt32: I was getting crashes including periods but I will try again!

keithamus (approved):
:+1: nice

PR #183, opened 2021-02-02
ButtonGroup component

manuelpuyol:
This PR creates a `ButtonGroupComponent` based on https://primer.style/css/components/buttons#button-groups and https://primer.style/components/Buttons#default-examples
I decided to make the buttons slots to make it harder to use this component incorrectly, so users don't add things that shouldn't be inside the BtnGroup

image


app/components/primer/button_group_component.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Primer
+  # Use ButtonGroupComponent to wrap a series of buttons.
+  # They will be rounded and spaced automatically
```
joelhawksley, line 5:
```suggestion
  # Use ButtonGroupComponent to render a series of buttons.
```

app/components/primer/button_group_component.rb
```diff
@@ -0,0 +1,35 @@
...
+  class ButtonGroupComponent < Primer::Component
+
```
joelhawksley, line 7:
```suggestion
```

stories/primer/button_group_component_stories.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class Primer::ButtonGroupComponentStories < ViewComponent::Storybook::Stories
+  layout "storybook_preview"
+
+  story(:button_group) do
+    content do |component|
+      component.button { "Button" }
+      component.button { "Button" }
+      component.button { "Button" }
+      component.button { "Button" }
```
joelhawksley, line 11: Might we use the example from the component class here that has the different button types?

test/components/button_group_component_test.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
```
joelhawksley, line 1: What should we do if no buttons are declared?

test/components/button_group_component_test.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def test_renders_button_items
+    render_inline(Primer::ButtonGroupComponent.new) { |c| c.button { "Button" } }
+
+    assert_selector("div.BtnGroup") do
+      assert_selector("button.btn.BtnGroup-item", text: "Button")
+    end
+  end
+
+  def test_renders_button_with_props
```
joelhawksley, line 22:
```suggestion
  def test_renders_button_with_arguments
```

test/components/button_group_component_test.rb
manuelpuyol: I guess not rendering is the best idea, will update

app/components/primer/button_group_component.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Primer
+  # Use ButtonGroupComponent to render a series of buttons.
+  class ButtonGroupComponent < Primer::Component
+    include ViewComponent::SlotableV2
+
+    renders_many :buttons, ->(classes: "", **kwargs) do
```
joelhawksley, line 8: What do you think is the best way to document that these slots take all of the arguments for ButtonComponent?

manuelpuyol: That's a good question. I don't think we aren't documenting slots in other components
@srt32 wdyt?

srt32: I've seen us document slots v1 behavior like this https://github.com/primer/view_components/pull/182/files#diff-a870f147518fb3db01d7aacb41bc479052f20a68f7e0f934fc78715d1851dd68R36-R39. We'd have to rewrite the Rakefile parser to handle slots v2?

srt32: Continued over in https://github.com/primer/view_components/pull/185#discussion_r569558545

PR #184, opened 2021-02-02
Update rubocop + lint

manuelpuyol:
Updated:
- rubocop (0.82.0)
- rubocop-github (0.16.0)
- rubocop-performance (1.7.1)
- rubocop-rails (2.5.2)

Linted errors and tuned some rules


test/primer/classify_test.rb
```diff
@@ -341,7 +341,7 @@ def refute_generated_class(input)
     assert_nil(Primer::Classify.call(**input)[:class])
   end
 
-  def assert_allocations(count, within:, &block)
```
joelhawksley, line 344: We don't need this argument? I would have thought we did. cc @BlakeWilliams

manuelpuyol: I think we don't need it since we use `yield`

Gemfile
```diff
@@ -3,7 +3,7 @@
 source "https://rubygems.org"
 
 gemspec
-rails_version = "#{ENV['RAILS_VERSION'] || '6.0.3.3'}"
+rails_version = (ENV["RAILS_VERSION"] || "6.0.3.3").to_s
```
joelhawksley, line 6: Mind updating this to `6.1.0`? We can also do so in another PR ❤️

joelhawksley (approved):

test/primer/classify_test.rb
BlakeWilliams: I forgot where I picked it up, but sometimes `&block` is added to method arguments to signal that they accept a block without requiring API consumers to look at the source.

I'm not opposed to removing it though, if that's what our linter recommends.

Gemfile
manuelpuyol: I can do it in a separate PR <3

PR #185, opened 2021-02-03
ButtonGroup component

manuelpuyol:
This PR creates a `ButtonGroupComponent` based on https://primer.style/css/components/buttons#button-groups and https://primer.style/components/Buttons#default-examples
I decided to make the buttons slots to make it harder to use this component incorrectly, so users don't add things that shouldn't be inside the BtnGroup

image


manuelpuyol:
Had to recreate because force-push messed some stuff up

app/components/primer/button_group_component.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Primer
+  # Use ButtonGroupComponent to render a series of buttons.
+  class ButtonGroupComponent < Primer::Component
+    include ViewComponent::SlotableV2
+
+    renders_many :buttons, ->(classes: "", **kwargs) do
```
manuelpuyol, line 8: From the other PR:

> @joelhawksley: What do you think is the best way to document that these slots take all of the arguments for ButtonComponent?


> @manuelpuyol: That's a good question. I don't think we aren't documenting slots in other components
@srt32 wdyt?

joelhawksley: @manuelpuyol sounds like it might be time to invent the right solution then 😄 

We can leave this PR as-is in that regard for now.

srt32: I've seen us document slots v1 behavior like this https://github.com/primer/view_components/pull/182/files#diff-a870f147518fb3db01d7aacb41bc479052f20a68f7e0f934fc78715d1851dd68R36-R39. We'd have to rewrite the Rakefile parser to handle slots v2?

.gitignore
```diff
@@ -12,6 +12,7 @@
 **/*.stories.json
 **/node_modules/
 *.log
+*.orig
```
srt32, line 15: What's this extension?

manuelpuyol: that's a file that sometimes is generated by my difftool

app/components/primer/button_group_component.rb
manuelpuyol: since I'm not defining a class as a slot here idk how YARD would deal with it

srt32: Open an issue for us track figuring this out? No need to block here.

.gitignore
joelhawksley: Ideally you'd handle this in a global .gitignore on your machine, but I'm cool merging this in ❤️

joelhawksley (approved):

joelhawksley:
Mind adding a CHANGELOG entry?

manuelpuyol:
> Mind adding a CHANGELOG entry?

I just noticed that we added a bunch of things without updating the CHANGELOG, I'll merge this as-is and open a PR with the updates!

PR #187, opened 2021-02-03
Marketing button component

manuelpuyol:
Not sure if this component should be named:

- `Primer::ButtonMarketingComponent` (following React's `ButtonDanger` pattern)
- `Primer::MarketingButtonComponent` (reads better)
- `Primer::Marketing::ButtonComponent` (grouping all marketing components together)

### Why not extend ButtonComponent?

They are really similar, but I decided to not add a `marketing:` argument to `ButtonComponent` since the options for its arguments differ

### Docs

image


joelhawksley (approved):
Code looks great! @srt32 thoughts on naming? I think this default of matching Primer React is fine.

srt32:
There don't seem to be too many Marketing related variants on Primer css so this seems 💯

PR #190, opened 2021-02-03
Add docs for {Flex,FlexItem}Component

srt32:
This change adds docs for `FlexComponent` and `FlexItemComponent`. There is a lot we could show here and I went with a handful of examples that exercises most of the knobs but not all the possible options.

image

image


srt32:
I feel like our current layout for PVC is not working well for something complex like flex. Comparing https://primer.style/css/utilities/flexbox#align-items to my WIP here on https://view-components-git-fork-srt32-st-docs-for-flex-item.primer.vercel.app/view-components/components/flex, I am not sure this page is going to scale nicely...

app/components/primer/flex_component.rb
```diff
@@ -32,6 +35,40 @@ class FlexComponent < Primer::Component
...
+    # @example 100|Direction column
+    #   <%= render(Primer::FlexComponent.new(direction: :column, bg: :gray)) do %>
+    #     <div class="p-5 border bg-gray-light">Item 1</div>
+    #     <div class="p-5 border bg-gray-light">Item 2</div>
+    #     <div class="p-5 border bg-gray-light">Item 3</div>
+    #   <% end %>
+    #
+    # @param justify_content [Symbol] Use this param to distribute space between and around flex items along the main axis of the container. <%= one_of(Primer::FlexComponent::JUSTIFY_CONTENT_OPTIONS) %>
+    # @param inline [Boolean] Defaults to false.
+    # @param flex_wrap [Boolean] Defaults to nil.
```
srt32, line 68: This is a boolean but also defaults to nil... what's the right thing here?

joelhawksley: @srt32 that's a good question. What do we do in Primer React?

srt32: Good question!

I also just noticed this on https://primer.style/components/Flex

> Previously, a Flex.Item component was used for flex item specific properties; Box now contains all those properties and should be used in its place.

Maybe we should deprecate FlexItem?

I am having a hard time grokking the PRC (I'm new to the codebase) but it looks to mainly delegate to `Box`, which delegates to styleSystem.

The API though is directly 

```
<Flex flexWrap="nowrap">
```

joelhawksley: @emplums what do you think? I'd generally lean towards deprecating all of the Flex-related components and just encourage the usage of Box with the flex-related system arguments.

app/components/primer/flex_component.rb
```diff
@@ -32,6 +35,40 @@ class FlexComponent < Primer::Component
     DEFAULT_DIRECTION = nil
     ALLOWED_DIRECTIONS = [DEFAULT_DIRECTION, :column, :column_reverse, :row, :row_reverse].freeze
 
+    # @example 100|Default
+    #   <%= render(Primer::FlexComponent.new(bg: :gray)) do %>
+    #     <div class="p-5 border bg-gray-light">Item 1</div>
+    #     <div class="p-5 border bg-gray-light">Item 2</div>
+    #     <div class="p-5 border bg-gray-light">Item 3</div>
```
joelhawksley, line 42: Might we use the Box component for these `div`s?

srt32: I tried using `FlexItem`'s previously but changed in https://github.com/primer/view_components/pull/190/commits/6da5b8380eac67cb64e09737d737dface5e5c96e because using plain ol divs solved the styling problem I brought up in https://github.slack.com/archives/CGJTTJ738/p1612458675188800

srt32: c32bb94 updates it to use `Box`!

srt32:
This PR is ready for another look. I'd love to get this merged so we can call everything existing documented as we sort out where we're going with Flex.

app/components/primer/flex_component.rb
```diff
@@ -37,30 +37,30 @@ class FlexComponent < Primer::Component
 
     # @example 100|Default
     #   <%= render(Primer::FlexComponent.new(bg: :gray)) do %>
-    #     <div class="p-5 border bg-gray-light">Item 1</div>
-    #     <div class="p-5 border bg-gray-light">Item 2</div>
-    #     <div class="p-5 border bg-gray-light">Item 3</div>
+    #     <%= render(Primer::BoxComponent.new(p: 5, classes: "border bg-gray-light")) { "Item 1" } %>
```
joelhawksley, line 40: Interesting. Are there not system arguments for the border/bg classes?

srt32: Attempting to not get bogged down here...

```
<%= render(Primer::BoxComponent.new(p: 5, bg: :gray_light, border: true)) { "Item 1" } %>
```

which is what I assumed would work, did not:

image

which we think is because of https://github.com/primer/view_components/blob/638d97c4a27d8fb3d993fdd617eda6cc704b2aa9/lib/primer/classify.rb#L216-L217

joelhawksley: Oh! Looks like we need to add support for `border: true` somehow. Mind filing a separate issue?

stories/primer/flex_component_stories.rb
```diff
@@ -0,0 +1,23 @@
...
+  story(:default) do
+    controls do
+      select(:justify_content, Primer::FlexComponent::JUSTIFY_CONTENT_OPTIONS, :center)
+      select(:align_items, Primer::FlexComponent::ALIGN_ITEMS_OPTIONS, :start)
+      select(:direction, Primer::FlexComponent::ALLOWED_DIRECTIONS, :row)
+      inline false
+      flex_wrap false
+    end
+
+    content do
```
joelhawksley, line 15: @manuelpuyol is there anything you think we could do to avoid the `html_safe` here?

joelhawksley (approved):

app/components/primer/flex_component.rb
emplums: @joelhawksley the proposed game plan for PRC is that `Box` and `Flex` will have all the same properties, the only difference is `Flex` will have `display: flex` set by default. The idea is that it will be handy shortcut from having to type `display: flex` over and over, and will still keep the ease of grokking the code to see exactly where flex is being used

PR #192, opened 2021-02-04
Dropdown component

manuelpuyol:
- Creates `DropdownComponent` based on https://primer.style/css/components/dropdown and https://primer.style/components/Dropdown
- Deprecates `DropdownMenuComponent` in favor of `Dropdown::MenuComponent`
- Menu now has `items` as slots instead of accepting raw HTML

image

image


app/components/primer/dropdown_component.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module Primer
+  # Dropdowns are lightweight context menus for housing navigation and actions.
```
srt32, line 4: I wonder if we could clarify when to use `Dropdown` and when to use `DropdownMenu`. It's not clear to me if we ever need a `DropdownMenu` now that we have this component...

manuelpuyol: I see `DropdownMenu` as a subcomponent of `Dropdown`, it is better to have them separate since it simplifies the logic in each one

joelhawksley: I agree with @srt32. Are there any cases in GitHub.com where we'd need both components?

manuelpuyol: They always go together, but I'm keeping them in separate files so we don't end up with a 300+ lines component.
As of why we need both is because it keeps the API cleaner, since I can customize the `Menu` with direction, alignment, paddings...

One thing I can do is change `Primer::DropdownMenuComponent` to `Primer::Dropdown::MenuComponent` like I did with SelectMenu to make it clearer that this is a subcomponent that shouldn't be used alone

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## main
 
+* Add `Dropdown` component.
+
+    *Manuel Puyol*
+
+* Deprecates `DropdownMenuComponent` in favor of `Dropdown::MenuComponent`. This component should not be used standalone, being part of `Dropdown`.
```
joelhawksley, line 9: If `Dropdown::MenuComponent` shouldn't be used directly, perhaps we can omit this mention of it? 

```suggestion
* Deprecate `DropdownMenuComponent`.
```

app/components/primer/dropdown/menu_component.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module Primer
+  module Dropdown
+    # DropdownMenus are lightweight context menus for housing navigation and actions.
+    # They're great for instances where you don't need the full power (and code)
+    # of the select menu.
```
joelhawksley, line 7: Do we need this description of we don't intend for this component to be consumed directly?

app/components/primer/dropdown/menu_component.rb
```diff
@@ -0,0 +1,64 @@
...
+      SCHEME_DEFAULT = :default
+      SCHEME_MAPPINGS = {
+        SCHEME_DEFAULT => "",
+        :dark => "dropdown-menu-dark"
+      }.freeze
+
+      DIRECTION_DEFAULT = :se
+      DIRECTION_OPTIONS = [DIRECTION_DEFAULT, :sw, :w, :e, :ne, :s].freeze
+
+      renders_many :items, lambda { |divider: false, **item_arguments|
```
joelhawksley, line 20: It's a bit of a nitpick, but what do you think about _always_ calling system arguments as such? (vs. `item_arguments` here)

app/components/primer/dropdown/menu_component.rb
```diff
@@ -0,0 +1,64 @@
...
+      #   <div class="position-relative mt-2">
+      #     <%= render(Primer::Dropdown::MenuComponent.new(header: "Options")) do |c|
+      #       c.item { "Item 1" }
+      #       c.item { "Item 2" }
+      #       c.item(divider: true)
+      #       c.item { "Item 3" }
+      #       c.item { "Item 4" }
+      #     end %>
+      #   </div>
+      #
```
joelhawksley, line 42: Do we want to have examples for a component that isn't used directly? (I could go either way)

app/components/primer/dropdown/menu_component.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module Primer
+  module Dropdown
```
joelhawksley, line 4: Can we define this inside the DropdownComponent? Like:

```suggestion
  class DropdownComponent
```

joelhawksley (commented):
👏🏻 looking good!

app/components/primer/dropdown/menu_component.rb
manuelpuyol: I wanted to avoid being so verbose `Primer::DropdownComponent::MenuComponent`, the `Component` repetition bothers me

app/components/primer/dropdown/menu_component.rb
manuelpuyol: I think the parent doc is fine too

app/components/primer/dropdown/menu_component.rb
joelhawksley: Hmm. I guess what concerns me is that we now have a new module here. While it _is_ verbose, the nesting of the component naming does reflect the relationship. This is definitely a case where I wish we didn't have `Component` in the name of our components...

app/components/primer/dropdown_menu_component.rb
```diff
@@ -0,0 +1,60 @@
...
+    SCHEME_MAPPINGS = {
+      SCHEME_DEFAULT => "",
+      :dark => "dropdown-menu-dark"
+    }.freeze
+
+    DIRECTION_DEFAULT = :se
+    DIRECTION_OPTIONS = [DIRECTION_DEFAULT, :sw, :w, :e, :ne, :s].freeze
+
+    def self.status
+      STATUSES[:deprecated]
```
srt32, line 18: 🎉

app/components/primer/dropdown/menu_component.rb
manuelpuyol: I think that having `Dropdown` (the parent component name) is enough to show the relationship. I really want to avoid being too verbose

app/components/primer/dropdown/menu_component.rb
```diff
@@ -0,0 +1,48 @@
...
+    # This component is part of `Primer::DropdownComponent` and should not be
```
srt32, line 5: 👍

srt32 (approved):

PR #193, opened 2021-02-05
Add support for functional color classes to `color`:

jshorty:
Prior to this change, color values supplied to the `color:` system argument are always classified with a `text-` prefix (`color: :my_color` becoming `text-my-color`).

This breaks with new functional color classes, where in some cases, a `color-` or other prefix is desired instead. This PR adds support for functional icon and text colors, a subset from the migration mapping at https://github.com/primer/css/blob/mkt/color-modes-docs/docs/content/support/v16-migration.mdx.

### DIFF
--- a/lib/primer/classify.rb
+++ b/lib/primer/classify.rb
@@ -202,11 +202,13 @@ module Primer
         elsif key == COLOR_KEY
           char_code = val[-1].ord
           # Does this string end in a character that is NOT a number?
-          memo[:classes] << if char_code >= 48 && char_code <= 57 # 48 is the charcode for 0; 57 is the charcode for 9
-                              "color-#{val.to_s.dasherize}"
-                            else
-                              "text-#{val.to_s.dasherize}"
-                            end
+          memo[:classes] <<
+            if (char_code >= 48 && char_code <= 57) || # 48 is the charcode for 0; 57 is the charcode for 9
+               val =~ /(primary|secondary|tertiary|link|success|warning|danger|info)/
+              "color-#{val.to_s.dasherize}"
+            else
+              "text-#{val.to_s.dasherize}"
+            end
         elsif key == DISPLAY_KEY
           memo[:classes] << "d#{breakpoint}-#{val.to_s.dasherize}"
         elsif key == VERTICAL_ALIGN_KEY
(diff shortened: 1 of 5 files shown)


lib/primer/classify.rb
```diff
@@ -202,11 +202,13 @@ def extract_value(memo, key, val, breakpoint)
...
           char_code = val[-1].ord
           # Does this string end in a character that is NOT a number?
-          memo[:classes] << if char_code >= 48 && char_code <= 57 # 48 is the charcode for 0; 57 is the charcode for 9
-                              "color-#{val.to_s.dasherize}"
-                            else
-                              "text-#{val.to_s.dasherize}"
-                            end
+          memo[:classes] <<
+            if (char_code >= 48 && char_code <= 57) || # 48 is the charcode for 0; 57 is the charcode for 9
+               val =~ /(primary|secondary|tertiary|link|success|warning|danger|info)/
```
BlakeWilliams, line 207: It looks like this is causing the allocations test to fail since this is introducing an allocation, I think we can fix it by moving `/(primary|secondary|tertiary|link|success|warning|danger|info)/` to a constant, like `COLOR_REGEX` and changing this line to:

```suggestion
               COLOR_REGEX.match?(val)
```

Thoughts?

jshorty: Thanks, that did the trick, though the same builds are now failing with something I don't think my change introduced?
```
activesupport-7.0.0.alpha requires ruby version >= 2.7.0, which is incompatible
with the current version, ruby 2.6.6p146
Error: Process completed with exit code 5.
```

BlakeWilliams: Hmmm, I'm not sure why that suddenly started happening. I can take a look shortly.

joelhawksley (approved):
Thanks for the PR @jshorty! Mind adding a CHANGELOG entry?

I also wonder if we might want to update the docs for the `color` argument. 🤔

lib/primer/classify.rb
jshorty: Looks like it resolved when Joel merged in main!

BlakeWilliams: Ah yeah, https://github.com/primer/view_components/pull/194 got merged which fixed the build :)

jshorty (commented):
Thanks for adding this, sorry I wasn't able to get to it yesterday!

joelhawksley:
@jshorty no problem! ❤️

PR #197, opened 2021-02-08
Cache common Primer values

BlakeWilliams:
* Add `Primer::Classify::Cache` class for preloading common primer classes
* Use `Primer::Classify::Cache` when available in `Primer::Classify`
* Add benchmark tests to avoid having to give imprecise counts due to
  differences in Ruby/Rails versions
* Add CI for running benchmarks against Rails main, Ruby 2.7
* Only run simplecov if `COVERAGE=1` in your environment

This takes us from 114 allocations when not cached, down to 40 allocations when cached. There's probably a lot more work that could be done here, like catching new keys that `Classify` supports and tracking that we're caching them correctly.

Co-authored with @kenyonj

### DIFF
--- a/lib/primer/classify/cache.rb
+++ b/lib/primer/classify/cache.rb
@@ -1,0 +1,125 @@
+# frozen_string_literal: true
+
+module Primer
+  class Classify
+    # :nodoc:
+    class Cache
+      # rubocop:disable Style/MutableConstant
+      LOOKUP = {}
+      # rubocop:enable Style/MutableConstant
+
+      class <<self
+        def read(memo, key, val, breakpoint)
+          value = LOOKUP.dig(breakpoint, key, val)
+          memo[:classes] << value if value
+        end
+
+        def clear!
+          LOOKUP.clear
+        end
+
+        def preload!
+          preload(
+            keys: Primer::Classify::MARGIN_DIRECTION_KEYS,
+            values: [-6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5, 6]
+          )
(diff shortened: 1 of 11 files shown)


BlakeWilliams:
Here's the benchmark results as well:

Screen Shot 2021-02-08 at 3 31 55 PM

lib/primer/classify/cache.rb
```diff
@@ -0,0 +1,125 @@
...
+          )
+
+          preload(
+            keys: Primer::Classify::DISPLAY_KEY,
+            values: [:flex, :block, :inline_block, :inline_flex, :none, :table, :table_cell]
+          )
+
+          preload(
+            keys: [Primer::Classify::COLOR_KEY, Primer::Classify::BG_KEY],
+            values: [:blue, :gray_dark, :gray, :gray_light, :red, :orange, :orange_light, :yellow, :green, :purple, :white, :pink]
```
joelhawksley, line 54: How did you arrive on these `values` here? (And for the rest of this file?) I'd love to have a process to ensure this optimization stays up to date 😄

joelhawksley (approved):
🤯 Next level work here, Blake and Justin! Dang!

lib/primer/classify/cache.rb
BlakeWilliams: It was very unscientific, we went through https://primer.style/css/ and tried to find colors to fill in here 😅.

There's probably a few better ways to handle it, like maybe parsing primer itself for color variables that we could use but we didn't have time to look into anything like that.

PR #199, opened 2021-02-08
Update storybook version

manuelpuyol:
We were using a RC for v6, but v6.1.7 is already out!


srt32 (approved):
Nice! How do we deploy / check it works?

PR #201, opened 2021-02-08
Simplify hash argument

joelhawksley:
Just a quick cleanup to simplify how we pass a hash to content_tag.

### DIFF
--- a/app/components/primer/base_component.rb
+++ b/app/components/primer/base_component.rb
@@ -116,7 +116,7 @@ module Primer
     end
 
     def call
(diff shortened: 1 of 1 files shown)


srt32 (approved):
Nice! Does this reduce an allocation as well?

joelhawksley:
> Nice! Does this reduce an allocation as well?

I don't believe so, but I'm not sure.

BlakeWilliams (approved):
I think that might be one less allocation, nice! 🎉

PR #202, opened 2021-02-09
Update border box example to use ERB blocks

joelhawksley:
While the existing example here is correct, I think the suggested example is more useful to developers using this pattern, as it shows how to pass ERB to slots.


manuelpuyol (approved):

docs/content/components/borderbox.md
```diff
@@ -15,14 +15,26 @@ BorderBox is a Box component with a border.
...
+    Body
+  <% end %>
+  <% component.slot(:row) do %>
+    Row one
+  <% end %>
+  <% component.slot(:row) do %>
+    Row two
+  <% end %>
+  <% component.slot(:row) do %>
+    Row three
```
waldnzwrld, line 32: I think it might be useful to show that ERB is fully welcomed in this block. Something like:

```suggestion
    <% if true %> 
      <p>Row three</p>
     <% else %>
       <p>No Row</p>
    <% end %>
```

Issue #188, opened 2021-02-03
How to document SlotV2 arguments


manuelpuyol:
Related: https://github.com/primer/view_components/pull/185#discussion_r569558545

We need to document arguments for our component's slots. Currently we document directly on their `Slot` class, but with the introduction of SlotV2 the syntax has changed and we need to decide how to document them.

joelhawksley:
@manuelpuyol thanks for raising this issue.

My main desires for this change is that the documentation for slots is presented consistently regardless of whether they are declared as classes or lambdas.

Do you have any ideas in mind?

PR #209, opened 2021-02-09
YARD extension for slots

manuelpuyol:
Closes https://github.com/primer/view_components/issues/188

- Adds a way to document slots by creating handlers for `renders_one` and `renders_many` calls.
- Adds a method to create links between components in the doc.
- Adds documentation to `ButtonGroup#buttons` slot

image

### DIFF
--- a/app/components/primer/button_group_component.rb
+++ b/app/components/primer/button_group_component.rb
@@ -7,7 +7,7 @@ module Primer
 
     # Required list of buttons to be rendered.
     #
-    # @param kwargs [Hash] the same arguments as <%= link_to_component(Primer::ButtonComponent) %>
+    # @param kwargs [Hash] The same arguments as <%= link_to_component(Primer::ButtonComponent) %>.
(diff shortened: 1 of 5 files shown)


app/components/primer/button_group_component.rb
```diff
@@ -5,6 +5,9 @@ module Primer
   class ButtonGroupComponent < Primer::Component
     include ViewComponent::SlotableV2
 
+    # Required list of buttons to be rendered.
+    #
+    # @param kwargs [Hash] the same arguments as <%= link_to_component(Primer::ButtonComponent) %>
```
joelhawksley, line 10: What do you think having this import all of the arguments from `ButtonComponent`, so we don't have to link to it?

manuelpuyol: we can do it as a next step! I don't want this PR to become overcomplicated

joelhawksley, line 10:
```suggestion
    # @param kwargs [Hash] The same arguments as <%= link_to_component(Primer::ButtonComponent) %>.
```

joelhawksley (approved):
Cool use of YARD here!

PR #214, opened 2021-02-10
Add static dump of component statuses 

srt32:
We'd like a way to inspect the statuses of components in projects where
this library is used without having to load the entire library (which
requires loading Rails and can be slow for things like linters).

This PR adds some code and a rake task to dump the components' statuses
as JSON into a file that can be read and parsed without loading the
environment.

We also add specs to make sure the file is up to date when developers
change the status of a component. If they change a status and don't
update this file, CI will fail and tell them to run the rake task and
commit the diff.

I am very open to alternative approaches here ;)


lib/primer/view_components.rb
```diff
@@ -6,5 +6,39 @@
 module Primer
   # :nodoc
   module ViewComponents
+    DEFAULT_STATUSES_PATH = File.expand_path("static")
```
srt32, line 9: Is this a good location?

lib/primer/view_components.rb
```diff
@@ -6,5 +6,39 @@
...
+    # dump_statuses generates the status hash and then serializes
+    # it as json at the given path
+    def self.dump_statuses(path: DEFAULT_STATUSES_PATH)
+      require "json"
+
+      statuses = generate_statuses
+
+      File.open(File.join(path, DEFAULT_STATUS_FILE_NAME), "w") do |f|
+        f.write(statuses.to_json)
+        f.write($INPUT_RECORD_SEPARATOR)
```
srt32, line 34: This is so that we don't get the "no newline" warning in the GitHub UI ;)

lib/primer/view_components.rb
```diff
@@ -6,5 +6,39 @@
...
+    # the component's status
+    def self.generate_statuses
+      statuses = {}
+      components = Primer::Component.descendants
+
+      components.each do |component|
+        statuses[component.to_s] = component.status.to_s
+      end
+
+      statuses
```
manuelpuyol, line 22:
```suggestion
      Primer::Component.descendants.each_with_object({}) do |component, mem|
        mem[component.to_s] = component.status.to_s
      end
```

manuelpuyol (approved):
lgtm

PR #217, opened 2021-02-10
Promote BlankSlateComponent to beta

srt32:


joelhawksley (approved):

manuelpuyol:
I don't recommend this since Blankslate uses SlotsV1, which will create a breaking change once we update it to V2

PR #218, opened 2021-02-11
Allow using a local view_component gem for test/local development

manuelpuyol:
Related to https://github.com/github/view_component/issues/525

If `VIEW_COMPONENT_PATH` env var is set, it will install the gem in that location instead of using RubyGems


Gemfile
```diff
@@ -16,6 +16,7 @@ gem "webpacker", "~> 5.0"
 # Reduces boot times through caching; required in config/boot.rb
 gem "bootsnap", ">= 1.4.2", require: false
 
+gem "view_component", path: ENV["VIEW_COMPONENT_PATH"] if ENV["VIEW_COMPONENT_PATH"]
```
BlakeWilliams, line 19:
```suggestion
gem "view_component", github: "github/view_component", branch: "main"
```

Would this work while also allowing https://bundler.io/man/bundle-config.1.html#LOCAL-GIT-REPOS to work?

srt32: were we getting view_components for free from Rails before this change?

BlakeWilliams: I think it may have been coming from the `gemspec` before this addition.

manuelpuyol: I'm not sure that we want to **always** override the `view_component` defined in `.gemspec`
I don't know what the implications would be

manuelpuyol:
@BlakeWilliams may I have a 👍  here?

BlakeWilliams (approved):

PR #219, opened 2021-02-11
Markdown component

manuelpuyol:
Creates a `MarkdownComponent` based on https://primer.style/css/components/markdown

image


app/components/primer/markdown_component.rb
```diff
@@ -0,0 +1,293 @@
+# frozen_string_literal: true
+
+module Primer
+  # Use MarkdownComponent to weap markdown content
```
srt32, line 4:
```suggestion
  # Use MarkdownComponent to wrap markdown content
```

srt32 (approved):

PR #220, opened 2021-02-11
Improve iframe on docs

manuelpuyol:
- iframes are now wrapped in a `Box p-3` class to make it similar to the CSS docs

image

- iframes can now automatically set their height using the `auto` option while documenting
  - this happens on load, so resizing the window won't resize the iframe

image

- iframes can still have hard-coded height
  - This is helpful for components with hidden content, such as `Dropdown` and `Tooltip`

image


Rakefile
```diff
@@ -180,6 +180,7 @@ namespace :docs do
         end
 
         initialize_method.tags(:example).each do |tag|
+          padding = 34 # accounts for p-3 and border
```
srt32, line 183: Not blocking but maybe we're getting to the point where could / should pull this markup and config into a erb template (or view component!)

Rakefile
```diff
@@ -192,7 +193,12 @@ namespace :docs do
           f.puts
           html = view_context.render(inline: tag.text)
 
-          f.puts("<iframe style=\"width: 100%; border: 0px; height: #{iframe_height}px;\" srcdoc=\"<html><head><link href=\'https://unpkg.com/@primer/css/dist/primer.css\' rel=\'stylesheet\'></head><body>#{html.tr('"', "\'").delete("\n")}</body></html>\"></iframe>")
+          iframe_attrs = if iframe_height == "auto"
+                           "onLoad={(e) => e.target.style.height = e.target.contentWindow.document.body.scrollHeight + #{padding} + 'px'} style=\"width: 100%; border: 0px;\""
```
srt32, line 197: I shudder a little about maintaining this but I have no specific objections...

srt32 (approved):

Rakefile
manuelpuyol: we have to revisit this whole pipeline, the Rakefile is getting unmaintainable

Issue #196, opened 2021-02-08
Add support for `border: true`


srt32:
Oh! Looks like we need to add support for `border: true` somehow. Mind filing a separate issue?

_Originally posted by @joelhawksley in https://github.com/primer/view_components/pull/190#discussion_r572259980_

PR #221, opened 2021-02-11
Support border: true

srt32:
closes https://github.com/primer/view_components/issues/196

Now we can pass `border: true` to apply the `border` class. Previously,
this would turn into `border-true`, which was silly.


joelhawksley (approved):
Looks good. Might we update the docs?

PR #233, opened 2021-02-16
Promote CounterComponent and LinkComponent to beta status

srt32:
We believe these two components are ready for beta in that they don't
use deprecated platform features, documentation is complete, and the API
is expected to be stable.


test/components/counter_component_test.rb
```diff
@@ -128,4 +128,8 @@ def test_render_arbitrary_string
 
     assert_selector("[title='?']", text: "?")
   end
+
+  def test_status
+    assert_equal Primer::CounterComponent.status, Primer::Component::STATUSES[:beta]
```
manuelpuyol, line 133: wdyt of having a custom selector `assert_state_X`?

srt32: I'll open a separate PR replacing the current uses. Cool?

srt32: WDYT of this?

```
diff --git a/test/components/link_component_test.rb b/test/components/link_component_test.rb
index 9e80ab0..ad9d4cd 100644
--- a/test/components/link_component_test.rb
+++ b/test/components/link_component_test.rb
@@ -39,6 +39,6 @@ class PrimerLinkComponentTest < Minitest::Test
   end

   def test_status
-    assert_equal Primer::LinkComponent.status, Primer::Component::STATUSES[:beta]
+    assert_component_is_state(Primer::LinkComponent, :beta)
   end
 end
diff --git a/test/test_helpers/component_test_helpers.rb b/test/test_helpers/component_test_helpers.rb
index 069c0a2..69c4f17 100644
--- a/test/test_helpers/component_test_helpers.rb
+++ b/test/test_helpers/component_test_helpers.rb
@@ -15,5 +15,9 @@ module Primer
     ensure
       FetchOrFallbackHelper.fallback_raises = true
     end
+
+    def assert_component_is_state(component, state)
+      assert_equal component.status, Primer::Component::STATUSES[state]
+    end
   end
 end
```

joelhawksley: How about `assert_component_state`?

srt32: Done in https://github.com/primer/view_components/pull/233/commits/d5bbfb0ac5c2eea3035ae1bb86669b80b3d7ab0a

manuelpuyol (approved):

PR #239, opened 2021-02-16
Convert FlashComponent to slots v2

srt32:
This PR converts `FlashComponent` to use slots v2.

This change is breaking in that the API is now similar but different enough to require consumers to make changes when upgrading.

The changes required look like this:

Previously:

```
<%= render(Primer::FlashComponent.new) do |component| %>
  This is a flash message with actions!
  <% component.slot(:actions) do %>
    <%= render(Primer::ButtonComponent.new(variant: :small)) { "Take action" } %>
  <% end %>
<% end %>
```

and now

```
<%= render(Primer::FlashComponent.new) do |component| %>
  This is a flash message with actions!
  <% component.action do %>
    <%= render(Primer::ButtonComponent.new(variant: :small)) { "Take action" } %>
  <% end %>
<% end %>
```


app/components/primer/flash_component.rb
```diff
@@ -61,16 +61,16 @@ def self.status
       Primer::Component::STATUSES[:beta]
     end
 
-    # :nodoc
-    class Actions < Primer::Slot
-      attr_reader :system_arguments
-
-      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    class ActionComponent < ViewComponent::Base
```
manuelpuyol, line 64:
```suggestion
    class ActionComponent < Primer::Component
```

app/components/primer/flash_component.rb
```diff
@@ -61,16 +61,16 @@ def self.status
...
       def initialize(**system_arguments)
         @system_arguments = system_arguments
         @system_arguments[:tag] = :div
         @system_arguments[:classes] = class_names(@system_arguments[:classes], "flash-action")
       end
+
+      def call
+        content
+      end
```
manuelpuyol, line 73:
```suggestion
      def call
        render(Primer::BaseComponent.new(**@system_arguments)) { content }
      end
```

manuelpuyol:
Since this is a breaking change, I wonder if we should simply have both slots for now (deprecating the V1) before removing the V1 definition completely 🤔 
I _think_ they can coexist without any major issues

srt32:
I'd love to not have to juggle both versions (and track coming back and cleaning up). I think the change is relatively simple for consumers but I'm not sure how we want to message the change so users are not surprised.

joelhawksley:
I could go either way re: backwards compatibility, but would prefer we default to not implementing both APIs. There is enough of a difference between the two that I'd be wary of mixing them.

manuelpuyol (approved):

PR #240, opened 2021-02-17
Avatar stack

manuelpuyol:
Creates AvatarStackComponent based on https://primer.style/css/components/avatars#avatar-stack

image


app/components/primer/avatar_stack_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= render Primer::BaseComponent.new(**@system_arguments) do %>
+  <%= render body_component do %>
+    <% avatars.each_with_index do |avatar, i| %>
+      <% if i == 2 %>
```
srt32, line 4: Do we want to make this value configurable?

manuelpuyol: I don't think Primer supports other values:

image

app/components/primer/avatar_stack_component.rb
```diff
@@ -0,0 +1,81 @@
...
+    #
+    # @example auto|With tooltip
+    #   <%= render(Primer::AvatarStackComponent.new(tooltipped: true, body_arguments: { label: 'This is a tooltip!' })) do |c| %>
+    #     <%= c.avatar(src: "http://placekitten.com/200/200", alt: "@kittenuser") %>
+    #     <%= c.avatar(src: "http://placekitten.com/200/200", alt: "@kittenuser") %>
+    #     <%= c.avatar(src: "http://placekitten.com/200/200", alt: "@kittenuser") %>
+    #   <% end  %>
+    #
+    # @param align [Symbol] <%= one_of(Primer::AvatarStackComponent::ALIGN_OPTIONS) %>
+    # @param tooltipped [Boolean] Whether to add a tooltip to the stack or not
```
srt32, line 38:
```suggestion
    # @param tooltipped [Boolean] Whether to add a tooltip to the stack or not.
```

app/components/primer/avatar_stack_component.rb
```diff
@@ -0,0 +1,81 @@
...
+    # @param body_arguments [Hash] Parameters to add to the Body. If `tooltipped` is set, has the same arguments as <%= link_to_component(Primer::TooltipComponent) %>.
```
srt32, line 39: Could we confirm the behavior is not winky if `tooltipped: true` is set but no body params are passed?

manuelpuyol: The component will raise, because if `tooltipped == true`, `body_params` requires `label:`
I can add a test checking this behavior

srt32: Raising seems unfortunate, in my opinion. I'd expect to end up with no tooltip at all. WDYT?

manuelpuyol: I think it is ok to raise. If you call a `TooltipComponent` without `label:` it will also raise
That makes it easier for the user to know what is wrong instead of simply not rendering anything

srt32: I was thinking more about the production use case where the tooltip value might be dynamically generated and there's some bug in the user code where it's missing. I'd not expect my page to blow up but I appreciate that passing the wrong args makes sense to blow up in dev. Does the library have a philosophy on that kind of question?

app/components/primer/avatar_stack_component.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    def body_component
+      if @tooltipped
+        Primer::TooltipComponent.new(**@body_arguments)
+      else
+        Primer::BaseComponent.new(**@body_arguments)
+      end
+    end
+
+    def before_render
```
srt32, line 69: For my own learning, why does this need to be in `before_render` and not in `initialize`?

srt32 (approved):

app/components/primer/avatar_stack_component.rb
manuelpuyol: Because `slots` aren't available in the `initialize`

srt32: Aha, interesting!

app/components/primer/avatar_stack_component.rb
manuelpuyol: I don't think we have a philosophy for that case. That's something we can discuss more!
Right now, if an user does not provide a **required** argument to a component, it will break

manuelpuyol: Maybe that's a smell on this component API, that's why I initially wanted `body` to be a slot, but couldn't because this specific use is not supported by the framework

PR #246, opened 2021-02-18
docs(adr): add 001 Record Architecture Decisions

keithamus:
This adds a basic document that describes how we might make architectural decisions, and record such decisions for future reference.


srt32:
I'm a 👍  to this idea but /docs is used to build our docs site (via gatsby). I wonder if we should use a different directory for more library developer focused docs?

srt32 (approved):

joelhawksley:
🙇🏻 @keithamus might you be willing to file a similar document on `view_component`? I'd love for us to use this framework there as well.

PR #250, opened 2021-02-18
docs(adr): add Developing and Publishing ClientSide Behaviours

keithamus:
This records some architectural decisions on how we will add client side (JS) behaviours to PVC.


adr/000x-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,46 @@
+# 000X. Developing and Publishing ClientSide Behaviours
```
srt32, line 1: What do you think about numbering the ADR's based off the PR number (250, in this case). That way the values are monotonically increasing and we should avoid merge conflicts with concurrent PR's.

adr/000x-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,46 @@
...
+ - The setup process should be uncomplicated and familiar to consumers; by using existing Rails paradigms, or JavaScript ecosystem paradigms or tools such as npm.
+ - Client side behaviours should be well tested and offer the same level of quality the existing non-behavioural components do.
+
+## Decision
+
+### TypeScript
+
+We will write behaviours in TypeScript, which is a compile-to-JavaScript language that offers stronger type guarantees leading to a higher quality level of code. TypeScript is already being used for [`primer/components`](https://github.com/primer/components) which is also a motivating factor.
+
+Behaviours will live in TypeScript files in the `app/components` directory next to their respective `.rb` files. By convention files will be named with the same basename as the Ruby counterpart. An `app/components/primer.ts` file will also exist which must import all other `.ts` files, acting as an index file. A single `.js` file with no external dependencies, which will live in `app/assets/javascripts/primer.js`. This single file can then be loaded through the Rails Asset pipeline, allowing for consumers to call `javascript_include_tag("primer")`.
```
manuelpuyol, line 26: Since we are publishing to NPM, does it make sense to have something Rails-specific? Rails could easily 
```js
import '@primer/view_components'
```

adr/000x-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Testing
+
+We will test our behaviours in the Browser environment (using System tests), as this is where they will be consumed. This will require using Browser APIs to automate testing. We will use test automation libraries that automate browsers via the [Devtools Protocol](https://chromedevtools.github.io/devtools-protocol/) such as [Puppeteer](https://pptr.dev/), [Playwright](https://playwright.dev/) or [Ferrum](https://ferrum.rocks/). Using a Devtools Protocol based tool gives us the best level of flexibility in testing, while also being compatible with current browsers.
+
+### Documentation
+
+We will document which components are behavioural, so that consumers are aware they are required to load the client side bundle code for these components to work properly. We will offer a documentation page on how to use the `javascript_include_tag` helper to load the bundle using the Asset Bundle, or otherwise import the npm package within their codebases.
```
manuelpuyol, line 38: Will it be possible to load single components if my project does not use all of them?

adr/000x-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Consequences
```
srt32, line 40: This write up seems great! Are there any major alternatives we explored that are worth documenting?

adr/000x-developing-and-publishing-clientside-behaviours.md
keithamus: It will be possible using the npm package.

adr/000x-developing-and-publishing-clientside-behaviours.md
keithamus: I'm open to the idea. What do others think? @manuelpuyol @joelhawksley?

adr/000x-developing-and-publishing-clientside-behaviours.md
keithamus: This document is attempting to describe offering both rails & not rails specific options. The `javascript_include_tag` allows consumers to add the primer behaviours without touching and JS, and only adding that tag to their `application.html.erb`. This means, however, that whatever is in `app/assets/javascripts/primer.js` will be loaded which may or may not include polyfills, additional libraries, so on and so forth.

In cases where more nuance is required, they can import the JS using the npm package by doing exactly what you describe, but this means they need to add their own JS file with the import and do all the necessary work to load _that_ JS file. This will use npm idiomatic conventions allowing for selective imports or bundle splitting.

I've clarified this by updating the docs.

adr/000x-developing-and-publishing-clientside-behaviours.md
keithamus: Through exploration I think the only two tenable solutions are to offer a sidecar npm package, or to align with Rails Engine features and offer a `app/assets/javascript/*` file for inclusion. Both of these are easy enough for us to do which is why I've documented us doing both. 

Other solutions are effectively "roll your own X". I'm not sure exactly how to describe this in the document as the scope is effectively infinite? 

Essentially: there's really two possible ways to do this, and this document is saying "let's do both".

adr/000x-developing-and-publishing-clientside-behaviours.md
manuelpuyol: I'm :+1: with that

adr/0246-record-architecture-decisions.md
```diff
@@ -1,4 +1,4 @@
-# 1. Record architecture decisions
+# 0246. Record architecture decisions
```
manuelpuyol, line 1: Mind changing the file name? 😄

adr/0250-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,48 @@
...
+ - The setup process should be uncomplicated and familiar to consumers; by using existing Rails paradigms, or JavaScript ecosystem paradigms or tools such as npm.
+ - Client side behaviours should be well tested and offer the same level of quality the existing non-behavioural components do.
+
+## Decision
+
+### TypeScript
+
+We will write behaviours in TypeScript, which is a compile-to-JavaScript language that offers stronger type guarantees leading to a higher quality level of code. TypeScript is already being used for [`primer/components`](https://github.com/primer/components) which is also a motivating factor.
+
+Behaviours will live in TypeScript files in the `app/components` directory next to their respective `.rb` files. By convention files will be named with the same basename as the Ruby counterpart. An `app/components/primer.ts` file will also exist which must import all other `.ts` files, acting as an index file. A single `.js` file with no external dependencies, which will live in `app/assets/javascripts/primer.js`. This single file can then be loaded through the Rails Asset pipeline, allowing for consumers to call `javascript_include_tag("primer")`.
```
joelhawksley, line 26: Note: test this architecture in the embedded application in this repository to confirm this works as described.

adr/0250-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Publishing to npm
+
+Additionally, we will produce an [npm package](https://www.npmjs.com/) which contains all the compiled down JavaScript both as individual files and as the single bundle in `app/assets/javascripts/primer.js`. This npm package will allow for consumption of the JS outside of using the traditional Rails Asset Pipeline. Developers will be able to call `import "@primer/view_components"` within their JS, which means their JS toolchain will be able to perform necessary transformations or optimisations. This will also allow for selective component imports, for example `import "@primer/view_components/time_ago_component"` could import just the "TimeAgoComponent".
```
joelhawksley, line 30:
```suggestion
Additionally, we will produce an [npm package](https://www.npmjs.com/) which contains all the compiled JavaScript both as individual files and as the single bundle in `app/assets/javascripts/primer.js`. This npm package will allow for consumption of the JS outside of using the traditional Rails Asset Pipeline. Developers will be able to call `import "@primer/view_components"` within their JS, which means their JS toolchain will be able to perform necessary transformations or optimisations. This will also allow for selective component imports, for example `import "@primer/view_components/time_ago_component"` could import just `TimeAgoComponent`.
```

adr/0250-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,48 @@
...
+
+Publishing a side-car package like this moderately increases complexity for this library, but drastically _increases_ flexibility and decreases the burden on the library to provide flexible paths for the variety of JavaScript compilers in use today. Good precedent exists for offering sidecar client side libraries, for example the [Turbo Framework offers a sidecar package](https://github.com/hotwired/turbo-rails#installation) if not using the built in Asset Pipeline.
```
joelhawksley, line 32:
```suggestion
Publishing a sidecar package like this moderately increases _complexity_ for this library, but drastically increases _flexibility_ and decreases the burden on the library to provide flexible paths for the variety of JavaScript compilers in use today. Good precedent exists for offering sidecar client side libraries, for example the [Turbo Framework offers a sidecar package](https://github.com/hotwired/turbo-rails#installation) if not using the built in Asset Pipeline.
```

joelhawksley (commented):
👏🏻

adr/0250-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,48 @@
+# 0250. Developing and Publishing ClientSide Behaviours
```
joelhawksley, line 1: 👋🏻 one other thought here: have we considered if we might want to provide some sort of encapsulation of the Javascript we write for ViewComponents? I could see us coupling the two together in similar fashion to our explorations of CSS modules, generating a selector specifically for a component and using it in the JS.

keithamus: Our current architecture is leveraging Web Components, which are encapsulated by tag name. Attempting to define an already defined web component causes an exception on the front-end. 

Web Components are encapsulated in that they are only activated for/on the pre-defined tags. In other words behaviour is not executed globally, only when an element enters the dom.

Not in the scope of this ADR but something I want to explore further is adding a suite of meta-tests which assert that the JS follows good conventions, such as:

 - The naming of JS files matches the naming of components.
 - The JS avoids global behaviours where applicable.
 - Any component with a JS file must have an integration test.

adr/0250-developing-and-publishing-clientside-behaviours.md
keithamus: We've now tested this and confirmed it works as described :+1:

keithamus (commented):
I think this is ready to go. This mostly now reflects reality, as we've implemented most of what this ADR documents.

adr/0250-developing-and-publishing-clientside-behaviours.md
joelhawksley: > Attempting to define an already defined web component causes an exception on the front-end.

This is the aspect I'm curious about. Can we provide an architecture where this exception is raised at compile time or as part of a test suite?

(This isn't a blocker at all)

adr/0250-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,48 @@
...
+
+Date: 2021-02-19
+
+## Status
+
+Proposed
+
```
joelhawksley, line 8: What do you think about only merging accepted ADRs? Then we could remove this field.

```suggestion
## Status

Proposed

```

adr/0250-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,48 @@
...
+Publishing a sidecar package like this moderately increases _complexity_ for this library, but drastically increases _flexibility_ and decreases the burden on the library to provide flexible paths for the variety of JavaScript compilers in use today. Good precedent exists for offering sidecar client side libraries, for example the [Turbo Framework offers a sidecar package](https://github.com/hotwired/turbo-rails#installation) if not using the built in Asset Pipeline.
+
+### Testing
+
+We will test our behaviours in the Browser environment (using System tests), as this is where they will be consumed. This will require using Browser APIs to automate testing. We will use test automation libraries that automate browsers via the [Devtools Protocol](https://chromedevtools.github.io/devtools-protocol/) such as [Puppeteer](https://pptr.dev/), [Playwright](https://playwright.dev/) or [Ferrum](https://ferrum.rocks/). Using a Devtools Protocol based tool gives us the best level of flexibility in testing, while also being compatible with current browsers.
```
joelhawksley, line 36:
```suggestion
We will test our behaviours in the Browser environment (using Rails System tests), as this is where they will be consumed. This will require using Browser APIs to automate testing. We will use test automation libraries that automate browsers via the [Devtools Protocol](https://chromedevtools.github.io/devtools-protocol/) such as [Puppeteer](https://pptr.dev/), [Playwright](https://playwright.dev/) or [Ferrum](https://ferrum.rocks/). Using a Devtools Protocol based tool gives us the best level of flexibility in testing, while also being compatible with current browsers.
```

adr/0250-developing-and-publishing-clientside-behaviours.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Documentation
+
+We will document which components are behavioural, so that consumers are aware they are required to load the client side bundle code for these components to work properly. We will offer a documentation page on how to use the `javascript_include_tag` helper to load the bundle using the Asset Bundle, or otherwise import the npm package within their codebases.
```
joelhawksley, line 40: Did you mean:

```suggestion
We will document which components are behavioural, so that consumers are aware they are required to load the client side bundle code for these components to work properly. We will offer a documentation page on how to use the `javascript_include_tag` helper to load the bundle using the Asset Pipeline, or otherwise import the npm package within their codebases.
```

?

joelhawksley (approved):

adr/0250-developing-and-publishing-clientside-behaviours.md
keithamus: The field is useful for `Superseded`/`Deprecated` decisions. I also feel like there are good motivations for intentionally documenting an ADR as `Rejected` - to document decisions which we don't want to implement (with reasons as to why).

keithamus:
```suggestion
## Status

Accepted

```

adr/0250-developing-and-publishing-clientside-behaviours.md
keithamus: This is already implicitly part of our test suite as we include the compiled bundle in the system tests!

Issue #247, opened 2021-02-18
Automate building docs on PR's


srt32:
We can use https://github.com/github/catalyst/blob/main/.github/workflows/docs-build.yml as inspiration.

We can also dump statuses as well.

PR #252, opened 2021-02-18
Build docs on Actions

srt32:
closes https://github.com/primer/view_components/issues/247

This change adds a workflow that will build the docs and statuses.json file on each push to `main`. It will then commit and push the changes directly to main.

This means that devs no longer have to build the docs (if they don't want to) since we'll automatically keep them up to date once a PR merges.

This change also removes the previously added CI check for the docs being stale.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,33 +87,6 @@ jobs:
       with:
         name: simplecov-resultset-rails${{matrix.rails_version}}-ruby${{matrix.ruby_version}}
         path: coverage
-  docs:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@master
-    - name: Setup Ruby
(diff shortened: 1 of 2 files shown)


srt32:
@manuelpuyol I think this is ready but I am not sure how best to test it (given I'm working off my fork and can't push to this repo. I still don't know why...). Have any ideas?

manuelpuyol:
@joelhawksley any idea why @srt32 can't push to the repo?

manuelpuyol (approved):

joelhawksley:
> @joelhawksley any idea why @srt32 can't push to the repo?

I see @srt32 is an Admin when I look at the repository settings 🤔

srt32:
I can do all the things in the UI but not from my CLI, which is very confusing. I'll keep trying things.

Issue #251, opened 2021-02-18
Classify breaks when empty string is passed to `color:`


manuelpuyol:
problem is here: https://github.com/primer/view_components/blob/386ac0cd188455ad3b7d5733a5291d7fa81fa402/lib/primer/classify.rb#L206

PR #255, opened 2021-02-19
Early return when classify value is an empty string

manuelpuyol:
Closes https://github.com/primer/view_components/issues/251

### DIFF
--- a/lib/primer/classify.rb
+++ b/lib/primer/classify.rb
@@ -182,7 +182,7 @@ module Primer
       end
 
       def extract_value(memo, key, val, breakpoint)
-        return if val.nil?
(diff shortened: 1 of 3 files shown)


lib/primer/classify.rb
```diff
@@ -182,7 +182,7 @@ def extract_hash(styles_hash)
       end
 
       def extract_value(memo, key, val, breakpoint)
-        return if val.nil?
+        return if val.nil? || val == ""
```
manuelpuyol, line 185: `.blank?` can't be used here because `false` is a valid value

BlakeWilliams (approved):
Looks great! ✨

PR #257, opened 2021-02-19
Make docs-build action work and push to the PR branch

srt32:
Follow up to https://github.com/primer/view_components/pull/252

The previous attempt was broken.  To avoid branch protection rule errors on main, now we commit the changes to the current PR branch (on push). We can choose to make this build required once we're happy with it so devs can't forget to commit the docs and also can't write broken docs.

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,10 +1,6 @@
 name: Docs Build
 
-on:
-  push:
-    branches:
-      - main
-      - srt32-patch-1
+on: [push, pull_request]
 
 jobs:
   build-docs:
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -33,4 +33,4 @@ jobs:
           git config --local user.name "Actions Auto Build"
           git add -f docs static/statuses.json
           git commit -m "docs: build docs" || true
-          git push --force origin HEAD:refs/heads/main
+          git push --force origin HEAD
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -29,4 +29,4 @@ jobs:
           git config --local user.name "Actions Auto Build"
           git add -f docs static/statuses.json
           git commit -m "docs: build docs" || true
-          git push --force origin HEAD
+          git push
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: Setup Ruby
         uses: actions/setup-ruby@v1
         with:
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,6 +1,6 @@
 name: Docs Build
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   build-docs:
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,6 +1,10 @@
 name: Docs Build
 
-on: [push]
+on:
+  push:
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-build.yml
```diff
@@ -6,24 +6,26 @@ on:
       - main
```
srt32, line 6:
```suggestion
      - main
      - srt32-patch-1
```

srt32:
Aha, protected branches are going to be a problem with this approach:

```

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Gemfile.lock

no changes added to commit (use "git add" and/or "git commit -a")
remote: error: GH006: Protected branch update failed for refs/heads/main.        
remote: error: 9 of 10 required status checks have not succeeded: 1 expected. At least 1 approving review is required by reviewers with write access.      
```

srt32:
We could build the docs on a different branch (docs, for example) and push there?

srt32:
cc/ @manuelpuyol for thoughts? ^

manuelpuyol:
Is it possible to commit to the branch that triggered the workflow? That way we could make it run on PRs instead of pushes on main

srt32:
> Is it possible to commit to the branch that triggered the workflow? That way we could make it run on PRs instead of pushes on main

We could totally do this but it means devs will need to `git pull` before adding more commits, which might be fine but surprising. Let's start there!

.github/workflows/docs-build.yml
```diff
@@ -1,33 +1,32 @@
...
-    - name: Generate docs and statuses
-      run: |
-        gem install bundler:2.2.9
-        bundle config path vendor/bundle
-        bundle update
-        bundle exec rake docs:build
-        bundle exec rake statuses:dump
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
srt32, line 10:
```suggestion
      - uses: actions/checkout@v2
```

.github/workflows/docs-build.yml
```diff
@@ -1,33 +1,32 @@
...
+          bundle exec rake docs:build
+          bundle exec rake statuses:dump
       - name: Commit & Push Docs Data
         run: |
           git config --local user.email "actions@github.com"
           git config --local user.name "Actions Auto Build"
           git add -f docs static/statuses.json
           git commit -m "docs: build docs" || true
-          git push --force origin HEAD:refs/heads/main
+          git push --force origin HEAD
```
srt32, line 32:
```suggestion
          git push
```

srt32:
It's alive!

image

.github/workflows/docs-build.yml
```diff
@@ -1,33 +1,34 @@
...
 
-on:
-  push:
-    branches:
-      - main
+on: [push]
 
 jobs:
+  build-docs:
+    name: Build Docs
```
srt32, line 7:
```suggestion
    name: Build Docs
```

srt32:
@manuelpuyol ready for another 👁️

.github/workflows/docs-build.yml
```diff
@@ -1,33 +1,34 @@
 name: Docs Build
 
-on:
-  push:
-    branches:
-      - main
+on: [push]
```
manuelpuyol, line 3: should be restrict to **not** run on `main`? otherwise the build will always be broken

srt32: I think we can skip main. Why would it be broken though? I think it might be wasteful but it should noop on main since it ran on the branhc.

manuelpuyol: Won't the `push` on main break?

srt32: Not sure! I find actions pretty confusing ;)

srt32: Updated in 4c7121e0db6755e4d1b00c7aa5ab5b35fc7763f6

manuelpuyol (approved):

Issue #259, opened 2021-02-19
Create helpers to render components


manuelpuyol:
Rendering simple components such as `Primer::HeadingComponent` is verbose and not user friendly.
e.g:
```erb
<%= render Primer::HeadingComponent.new(tag: :h1) { "My Heading" } %>
```

We should provide helpers to simplify the API, something like:

```erb
<%= primer_heading(:h1, "My Heading") %>
```

manuelpuyol:
Related https://github.com/primer/view_components/pull/58

manuelpuyol:
closing because of https://github.com/primer/view_components/pull/260
For the moment, we want to experiment only with `primer(:octicon`and `primer(:heading` and see how it feels

PR #260, opened 2021-02-19
Proposal: ViewHelpers for components

manuelpuyol:
Proposal for https://github.com/primer/view_components/issues/259

This PR provides a simpler way to render PVC components, providing a `primer(:name, **kwargs) { block }` view helper.

### How it works

Created a DSL `Primer::DSL::ViewHelper` that enables components to register their own helpers using:

```rb
view_helper :my_helper
```

Also provides `Primer::ViewHelper` that creates the `primer(name, **kwargs, &block)`. The helper will look for registered helpers and raise an error if not found. This is the module needs to be manually included to be used outside of `Primer::Component`


joelhawksley (commented):
👋🏻 I'd be curious to hear your thoughts on going with `primer(:octicon...` vs. `primer_octicon`, etc. Did you consider the latter?

manuelpuyol:
I did consider both! The difference is how we would register the components in each option

To have helpers like `primer_octicon` I'd need a static mapping of `helper -> component` like 
```rb
MAPPNGS = {
  octicon: Primer::OcticonComponent
}
```

And then, when the `ViewHelper` module was included, I could define methods for each helper using `define_method` and so on.

On the other hand, `primer(:octicon)` allowed the creation of a DSL so each component could register itself instead of creating yet another static list that I have to keep adding components.

From a user point of view, calling `primer_octicon icon: :some_icon` vs `primer :octicon, icon: :some_icon` is the same thing. So I opted for building the solution with the DSL

lib/primer/dsl/view_helper.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require "active_support/concern"
+
+module Primer
+  # :nodoc:
+  module DSL
```
BlakeWilliams, line 7: Since this is closely related to `Primer::ViewHelper` what are your thoughts on defining this as a child module (`Primer::ViewHelper::DSL`) of `ViewHelper` to avoid having two modules named `ViewHelper`?

lib/primer/dsl/view_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      class ViewHelperAlreadyDefined < StandardError; end
+
+      included do
+        class_attribute :helpers, instance_writer: false, default: {}
+      end
+
+      class_methods do
+        def view_helper(name)
+          raise ViewHelperAlreadyDefined if helpers[name].present?
```
BlakeWilliams, line 27:
```suggestion
          raise ViewHelperAlreadyDefined.new("#{name} is already defined") if helpers[name].present?
```

I think we need to call `.new` here, thoughts on adding some details to the error string as well?

lib/primer/dsl/view_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+    #   include Primer::DSL::ViewHelper
+    #   view_helper :my_component
+    # end
+    module ViewHelper
+      extend ActiveSupport::Concern
+
+      class ViewHelperAlreadyDefined < StandardError; end
+
+      included do
+        class_attribute :helpers, instance_writer: false, default: {}
```
BlakeWilliams, line 22:
```suggestion
        class_attribute :primer_helpers, instance_writer: false, default: {}
```

Super minor, but thoughts on "namespacing" the attribute to help avoid conflicts with other libraries/gems?

lib/primer/view_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+  # `render Primer::NameComponent.new(**kwargs) { block }`
+  module ViewHelper
+    extend ActiveSupport::Concern
+
+    class ViewHelperNotFound < StandardError; end
+
+    def primer(name, **component_args, &block)
+      component = Primer::Component.helpers[name]
+
+      raise ViewHelperNotFound if component.blank?
```
BlakeWilliams, line 17:
```suggestion
      raise ViewHelperNotFound.new("no component defined for helper #{name}") if component.blank?
```

test/primer/dsl/view_helper_test.rb
```diff
@@ -0,0 +1,21 @@
...
+  include Primer::DSL::ViewHelper
+end
+
+class Primer::DSL::ViewHelperTest < Minitest::Test
+  def test_register_view_helpers
+    assert_respond_to(TestComponent, :view_helper)
+
+    TestComponent.view_helper :test
+
+    assert_raises Primer::DSL::ViewHelper::ViewHelperAlreadyDefined do
```
BlakeWilliams, line 15: Thoughts on making a separate, explicit test for potential error conditions?

test/primer/view_helper_test.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class Primer::ViewHelperTest < Minitest::Test
+  include Primer::ViewHelper
+  include Primer::ComponentTestHelpers
+
+  # The helper calls #render, but it is not available in tests
+  alias render render_inline
```
BlakeWilliams, line 10: Do we have an integration test that uses the full Rails stack that we could utilize for these tests?

I don't think it's worth adding one in this iteration if we don't, but it could be beneficial if the tests went through the full rendering pipeline instead of through `render_inline`.

BlakeWilliams (commented):
I like this a lot!

I'm really excited to see how we utilize this since I think there's a lot we can build on here. Like automatic generation of helpers based on component name, utilizing this pattern upstream, etc.

test/primer/view_helper_test.rb
manuelpuyol: we don't right now, but we are working on it!

app/lib/primer/classify.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require_relative "classify/cache"
```
BlakeWilliams, line 3: Tests are passing so I assume it's fine, but is this already included elsewhere?

BlakeWilliams (approved):
🎉

app/lib/primer/classify.rb
manuelpuyol: Since I moved the helpers from `lib` to `app/lib`, the engine now requires everything by itself (once), so we don't have to worry about manually including files

PR #262, opened 2021-02-22
Link component storybook on docs

manuelpuyol:
Closes https://github.com/primer/view_components/issues/281

Following the [customization](https://primer.style/doctocat/usage/customization) docs on doctocat, they suggest using [shadowing](https://www.gatsbyjs.com/blog/2019-04-29-component-shadowing/) to customize the theme.

In this PR, I customized the main `Layout` from `gatsby-theme-doctocat` so we can add a `View storybook` link below `View source`.  I also created a custom `Example` component to remove the need of `iframes` for examples! For this to work properly, I also changed the theme's `Head` component to include `PrimerCSS`.

Since we aren't using `iframes` anymore, I updated our docs from `@example [height]|[name]|[description]` to `@example [name]|[description]`

image


docs/src/@primer/gatsby-theme-doctocat/components/iframe.js
```diff
@@ -0,0 +1,27 @@
...
+  };
+  if (height === "auto") {
+    rest.onLoad = (e) => {
+      e.target.style.height = `${e.target.contentWindow.document.body.scrollHeight + PADDING}px`
+    }
+  } else {
+    rest.style.height = `${Number(height) + PADDING}px`
+  }
+
+  console.log(content)
```
srt32, line 20:
```suggestion
```

docs/src/@primer/gatsby-theme-doctocat/components/layout.js
```diff
@@ -0,0 +1,154 @@
+import {
```
srt32, line 1: It feels really unfortunate to  duplicate this whole component. Maybe @colebemis could upstream the idea of supporting multiple links?

srt32:
Way out there idea... could we embed the specific component's storybook into the docs page directly?

docs/src/@primer/gatsby-theme-doctocat/components/layout.js
srt32: One other, in the middle idea. What if the section where we currently have the "View source example" was refactored to be a `LinksList` component. Then we could override just that more targeted component to render our links.

manuelpuyol: I'd ❤️  to have this directly in doctocat! But I don't think it is a big issue to have it here while it isn't supported

manuelpuyol:
I don't remember ever seeing an embedded Storybook 🤔 
I wonder how knobs would work

Also, a component may have multiple stories, so we'd still need a sidebar for navigation

srt32 (approved):

joelhawksley (changes requested):
👋🏻 I'd like to see us resolve the question about where we plan to host the PVC storybook domain-wise before landing this PR.

docs/content/components/avatar.md
```diff
@@ -2,7 +2,7 @@
 title: Avatar
 status: Beta
 source: https://github.com/primer/view_components/tree/main/app/components/primer/avatar_component.rb
-storybook: https://primer-view-components.herokuapp.com/?path=/story/primer-avatar-component
+storybook: https://primer.style/view-components/stories/?path=/story/primer-avatar-component
```
srt32, line 5: https://primer.style/view-components/stories is 404ing for me. Do we expect it to work at this point?

manuelpuyol: Yeah, I'm waiting on https://github.com/primer/primer.style/pull/216

manuelpuyol:
https://primer.style/view-components/stories/ is working!

joelhawksley:
Screen Shot 2021-02-23 at 1 26 08 PM

The vertical spacing between these two links feels a little tight. What do you think @colebemis?

docs/src/@primer/gatsby-theme-doctocat/components/example.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+function Example({src}) {
+  return <div className="Box p-3" dangerouslySetInnerHTML={{ __html: src}}></div>
```
manuelpuyol, line 4: Here I had to use `dangerouslySetInnerHTML` instead of `{children}` because the theme was somehow messing with our example tags.

E.g:
`<details-menu class="x">` became `<details-menu classname="x">` and `<u>` became `<u class="List-some-md5">`

Even though this is _dangerous_ we control what is being set as the inner HTML, so it is fine to use it

maybe @colebemis has an idea on how to fix this issue another way

docs/src/@primer/gatsby-theme-doctocat/components/layout.js
colebemis: 👋 I opened a PR to upstream this into Doctocat: https://github.com/primer/doctocat/pull/220

I also included some design updates.

docs/src/@primer/gatsby-theme-doctocat/components/head.js
```diff
@@ -0,0 +1,25 @@
...
+
+  return (
+    <Helmet>
+      <title>{title}</title>
+      <meta name="description" content={description} />
+      <meta property="og:title" content={title} />
+      <meta property="og:description" content={description} />
+      <meta property="og:image" content={siteMetadata.imageUrl} />
+      <meta property="twitter:card" content="summary_large_image" />
+      <link href="https://unpkg.com/@primer/css-next@canary/dist/primer.css" rel="stylesheet" />
```
colebemis, line 20: Looks like including this CSS file changes the font size of all the text on the page from `16px` to `14px`. It be great if we could prevent that somehow.

docs/src/@primer/gatsby-theme-doctocat/components/example.js
colebemis: `dangerouslySetHTML` seems fine to me 👍

PR #263, opened 2021-02-22
TimelineItem Slots v2

manuelpuyol:
before:

```erb
<%= render(Primer::TimelineItemComponent.new) do |component| %>
  <% component.slot(:avatar, src: "https://github.com/github.png", alt: "github") %>
  <% component.slot(:badge, bg: :green, color: :white, icon: :check) %>
  <% component.slot(:body) { "Success!" } %>
<% end %>
```

after:

```erb
<%= render(Primer::TimelineItemComponent.new) do |component| %>
  <% component.avatar(src: "https://github.com/github.png", alt: "github") %>
  <% component.badge(bg: :green, color: :white, icon: :check) %>
  <% component.body { "Success!" } %>
<% end %>
```


app/components/primer/timeline_item_component.html.erb
```diff
@@ -1,17 +1,5 @@
-<%= render Primer::BaseComponent.new(**system_arguments) do %>
-  <% if avatar %>
-    <%= render Primer::AvatarComponent.new(alt: avatar.alt, src: avatar.src, size: avatar.size, square: avatar.square, **avatar.system_arguments) %>
-  <% end %>
-
-  <% if badge %>
```
srt32, line 6: Do we need to maintain the `if` logic in the new approach?

app/components/primer/timeline_item_component.rb
```diff
@@ -1,22 +1,49 @@
...
+    }
 
-    with_slot :avatar, class_name: "Avatar"
-    with_slot :badge, class_name: "Badge"
-    with_slot :body, class_name: "Body"
+    # Badge that will be connected to other TimelineItems.
+    #
+    # @param icon [String] Name of [Octicon](https://primer.style/octicons/) to use.
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    renders_one :badge, Primer::TimelineItem::BadgeComponent
```
srt32, line 26: For my own learning, why did you choose to make badge its own class instead of using a lambda like with the other slots?

app/components/primer/timeline_item_component.html.erb
manuelpuyol: we don't! because if the user does not call the slot it will be `nil` and then `<%= nil %>` is noop

srt32 (approved):
A question or two but 👍

app/components/primer/timeline_item_component.rb
manuelpuyol: because it was more complex. the `badge` slot also defines its own content (an octicon) so I thought it was better to keep it as a separate component.

I don't think we can declare content for the component in a lambda, correct me if I'm wrong @BlakeWilliams

joelhawksley: If `BadgeComponent` is only ever used in this context, IMO we should try to embed it inside this component. I'd also be curious to hear @BlakeWilliams' thoughts 😄

manuelpuyol: I moved `BadgeComponent` to live inside `TimelineItemComponent` :)

PR #266, opened 2021-02-23
Add integration test harness using Ferrum

joelhawksley:
### What

This PR introduces a proof-of-concept integration harness using Ferrum, leveraging ViewComponent previews to provide the example routes.

### Why

In order to test javascript behaviors, we need to test our ViewComponents in integration, in a browser environment. 

See https://github.com/primer/view_components/pull/250 for more detail.


manuelpuyol:
@joelhawksley Does storybook and `yarn build` still work without webpack in `demo/`?

demo/package.json
```diff
@@ -5,7 +5,6 @@
   "dependencies": {
     "@rails/actioncable": "^6.0.0",
     "@rails/ujs": "^6.0.0",
```
manuelpuyol, line 7: since we are here... do we need these packages?

joelhawksley:
> @joelhawksley Does storybook and yarn build still work without webpack in demo/?

Just checked locally, and both still work 👍🏻

test/application_system_test_case.rb
```diff
@@ -0,0 +1,18 @@
...
+
+require "test_helper"
+require "capybara/rails"
+require "capybara/minitest"
+require "capybara/cuprite"
+
+Ferrum::Browser.new(process_timeout: 5)
+
+class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
+  driven_by :cuprite, using: :chrome, screen_size: [1400, 1400]
```
manuelpuyol, line 11: any reasoning behind 1400x1400?

test/application_system_test_case.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def with_preview(preview_name)
+    component_uri = self.class.name.split("Integration").first.underscore
+
+    visit("/rails/view_components/#{component_uri}/#{preview_name}")
```
manuelpuyol, line 16: we also have the stories routes available here! wdyt of using them? 
e.g: `/rails/stories/primer/avatar_stack_component/avatar_stack`

test/application_system_test_case.rb
joelhawksley: Nope, this was default from examples we saw.

test/application_system_test_case.rb
joelhawksley: Interesting. I could go either way. I think we'll want to define specific previews for specific integration tests, but I could also see us reusing the Storybook infrastructure. Ultimately they are similar concepts that should probably share more logic than they do in this PR. 

Perhaps we can land what we have here and look into this further at a later time?

manuelpuyol (approved):
We can talk more about previews vs storybook later 🚢

Issue #231, opened 2021-02-16
Promote LabelComponent to beta


srt32:
TODO:

* We need to remove the deprecated color options before promoting.

PR #270, opened 2021-02-23
Add functional functional color support to LabelComponent

srt32:
The color mapping is pulled from https://github.com/primer/css/blob/mkt/color-modes-docs/docs/content/support/v16-migration.mdx#labels

supports https://github.com/primer/view_components/issues/231

This change updates the internal CSS we use to use the existing classes
while full functional color support is implemented in Primer CSS.

### DIFF
--- a/app/components/primer/label_component.rb
+++ b/app/components/primer/label_component.rb
@@ -4,12 +4,12 @@ module Primer
   # Use labels to add contextual metadata to a design.
   class LabelComponent < Primer::Component
     NEW_SCHEME_MAPPINGS = {
-      primary: "Label--primary",
-      secondary: "Label--secondary",
-      info: "Label--info",
-      success: "Label--success",
-      warning: "Label--warning",
-      danger: "Label--danger"
+      primary: "Label--gray-darker",
+      secondary: "Label--gray",
+      info: "Label--blue",
+      success: "Label--green",
+      warning: "Label--yellow",
+      danger: "Label--red"
     }.freeze
 
     DEPRECATED_SCHEME_MAPPINGS = {
--- a/docs/content/components/state.md
+++ b/docs/content/components/state.md
@@ -43,7 +43,7 @@ Component for rendering the status of an item.
 | Name | Type | Default | Description |
 | :- | :- | :- | :- |
 | `title` | `String` | N/A | `title` HTML attribute. |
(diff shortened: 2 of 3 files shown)


srt32:
@joelhawksley @colinkeany this PR partially undoes https://github.com/primer/view_components/pull/135. Will this change break anything you're working on?

app/components/primer/label_component.rb
```diff
@@ -4,12 +4,12 @@ module Primer
...
-      danger: "Label--danger"
+      primary: "Label--gray-darker",
+      secondary: "Label--gray",
+      info: "Label--blue",
+      success: "Label--green",
+      warning: "Label--yellow",
+      danger: "Label--red"
     }.freeze
 
     DEPRECATED_SCHEME_MAPPINGS = {
```
joelhawksley, line 15: I _believe_ we should instead be deleting this constant and leaving `NEW_SCHEME_MAPPINGS` in place, perhaps renaming it to `SCHEME_MAPPINGS`?

srt32: I'm honestly confused at this point ;)

Deleting this mapping would be a breaking change. That's what we want?

joelhawksley: I think that's what we want in order to move this component to `Beta`. Perhaps we could start by raising some sort of warning in development/test if the deprecated mappings are used?

cc @manuelpuyol as I think he's worked on deprecating component options.

manuelpuyol: `Primer::LabelComponent` has 179 uses in dotcom, maybe we should:
1. deprecate old options
2. open PRs in dotcom updating the uses
3. remove old options

joelhawksley: @manuelpuyol that sounds good. Might it be possible to use CI to do a runtime check to ensure we've removed all uses of the deprecated options?

srt32: Assuming https://github.com/primer/view_components/pull/278 works, I think we want to close out this PR. We'd then work on removing the deprecated use cases in dotcom and open a follow up PR here removing the deprecated options. Sound right?

manuelpuyol: yeah, we have to find a better deprecation method. I worked with this before https://github.com/primer/view_components/pull/40 but I'm not happy with how it turned out

manuelpuyol:
Closing because of https://github.com/primer/view_components/pull/278

PR #273, opened 2021-02-23
NPM package setup

manuelpuyol:
cherry-picked changes from https://github.com/primer/view_components/compare/ferrum-plus-timeago to allow better work parallelization

This setup will allow us to add JS dependencies and build our own compiled assets. With that, we'll be able to create and ship components with behaviors such as `TimeAgo` and `TabContainer`.

see https://github.com/primer/view_components/pull/273#issuecomment-785348594 for more context


app/assets/javascripts/primer.js
```diff
@@ -0,0 +1 @@
+
```
keithamus, line 1: I wonder whether or not we need to commit this file?

It will be needed for publishing the npm package and the gem, but as it’s an artefact generated from other source, maybe we should not check it in?

demo/app/assets/javascripts/primer.js
```diff
@@ -0,0 +1 @@
+import '../../../../app/assets/javascripts/primer.js';
```
keithamus, line 1: Does this work? I think this needs to be a symlink, or we need the demo package to depend on the root package as a gem

manuelpuyol: so, the demo does depend on `primer_view_components` gem, but yeah it does not look like it is working as-is 🤔

app/assets/javascripts/primer.js
manuelpuyol: I think we need to commit it, otherwise it won't be available when vendoring the gem

manuelpuyol:
after pairing with @keithamus and some exploration, some big changes happened 😅 

Looks like the way we were loading things was unconventional, so we used `turbo-rails` as an example on how to configure our paths https://github.com/hotwired/turbo-rails/blob/main/lib/turbo/engine.rb

This led to a change where I moved our helpers, such as `Classify` from `lib/primer` to `app/lib/primer`.

With the new loading approach, we were able to get rid of the `app/components/primer/view_components.rb` entrypoint, since the Engine knows how to load itself!

We also added an `assets` initializer, pointing where our JS is so other apps can add it to their pipeline.

keithamus (approved):
This LGTM - I wonder if it’s worth breaking up this into smaller prs that we could ship/iterate on faster?

Either way this is good

joelhawksley (approved):
😅 I can tell this took a lot of hard work! Thanks for figuring it out ❤️

PR #275, opened 2021-02-23
Reduce time spent in Primer::BaseComponent test selector

srt32:
When we initialize `Primer::BaseComponent`, we are calling `except` on the
args hash twice. Once is to remove a test selector and the other is to
remove classify related keys.

This change experiments with calling except only once. We allocate a new
array to generate the list to exclude but should allocate only one
additional hash due to calling except once instead of twice. We will allocate the same number of objects (we are trading a hash for an array) but we should spend less time processing the hash since we'll only call `except` on it once.

We also save an allocation by `s/merge/merge!` when we pass the params to `content_tag`.

### DIFF
--- a/test/benchmarks/bench_classify.rb
+++ b/test/benchmarks/bench_classify.rb
@@ -57,7 +57,7 @@ class BenchClassify < Minitest::Benchmark
   end
 
   def bench_allocations_for_base_component
-    assert_allocations 103 do
+    assert_allocations 100 do
       component = Primer::BaseComponent.new(tag: :div, classes: @values)
 
       component.call
--- a/test/benchmarks/bench_classify.rb
+++ b/test/benchmarks/bench_classify.rb
@@ -56,6 +56,14 @@ class BenchClassify < Minitest::Benchmark
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/benchmarks/bench_classify.rb
```diff
@@ -56,6 +56,14 @@ def bench_allocations_with_cache
     end
   end
 
+  def bench_allocations_for_base_component
+    assert_allocations 100 do
```
srt32, line 60:
```suggestion
    assert_allocations 98 do
```

srt32:
@BlakeWilliams @joelhawksley could you help me understand some next steps to determine further if this change is worth pursuing?

BlakeWilliams:
@srt32 Happy to!

I think we could add a benchmark test for base component and see a before/after of total # of allocations similar to [what we do for classify](https://github.com/primer/view_components/blob/main/test/benchmarks/bench_classify.rb#L50-L57).

I also sometimes pull in https://github.com/ko1/allocation_tracer to correlate lines of code to allocations.

Feel free to reach out on Slack if you'd like to pair on any of that.

Issue #280, opened 2021-02-25
Setting label background


artemsaveliev:
Trying to replicate labels with background
https://primer.style/css/components/labels#issue-labels 
![image](https://user-images.githubusercontent.com/15679218/109174020-4e928b00-7739-11eb-821b-a55e968354ee.png)

I used this in erb:
```
render Primer::LabelComponent.new(scheme: :red, color: :white, bg: :red)
```
which shows this:
![image](https://user-images.githubusercontent.com/15679218/109175422-a087e080-773a-11eb-8d91-a030c8c134bc.png)

The reason seems to be the "initial!important" override:
![image](https://user-images.githubusercontent.com/15679218/109175648-dcbb4100-773a-11eb-91a8-75c7a483082f.png)

PR #282, opened 2021-02-25
TabNav and TabContainer components

manuelpuyol:
This PR moves `TabNav` and `TabContainer` components from dotcom.
One change this PR makes from the dotcom components is that I dropped `TabNavContainerComponent` in favor of using `TabNavComponent.new(with_panel: true)`.

image

gif

**Note: The behavior does not work on the docs yet, but it does in Storybook!**


test/application_system_test_case.rb
```diff
@@ -11,7 +11,7 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   driven_by :cuprite, using: :chrome, screen_size: [1400, 1400]
 
   def with_preview(preview_name)
-    component_uri = self.class.name.split("Integration").first.underscore
+    component_uri = self.class.name.gsub("Test", "").gsub("Integration", "").underscore
```
srt32, line 14: Could you help me understand this change?

stories/primer/tab_nav_component_stories.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  story(:tab_nav) do
+    controls do
+      with_panel true
+    end
+
+    content do |c|
+      c.tab(selected: true, title: "Tab 1") { "Panel 1" }
+      c.tab(title: "Tab 2") { "Panel 2" }
+      c.tab(title: "Tab 3") { "Panel 3" }
```
srt32, line 14: Is there a way to add an open text field so users could pass in arbitrary `classes`? It would make testing issues like https://github.com/primer/view_components/issues/280 much easier!

test/application_system_test_case.rb
manuelpuyol: I moved the integration tests from `test/components/my_component_integration_test.rb` to `test/integration/my_component_test.rb`
so the class name was `MyComponentIntegrationTest` and now is `IntegrationMyComponentTest`, which changes how we extract the component name

stories/primer/tab_nav_component_stories.rb
manuelpuyol: unfortunately we can only control the **parent** component in Storybook :(
It's a `view_component_storybook` limitation, but I also don't know if component previews allow controlling a component's slot
@BlakeWilliams thoughts?

app/components/primer/tab_container_component.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Primer
+  # Use TabContainer to create tabbed content with keyboard support.
```
srt32, line 4:
```suggestion
  # Use TabContainer to create tabbed content with keyboard support. This component requires javascript.
```

Maybe a good enough note for now about the JS dep?

app/assets/javascripts/primer_view_components.js
```diff
@@ -1 +1 @@
...
...return;const a=t.target.closest('[role="tab"]');if(!a||!a.closest('[role="tablist"]'))return;e(this,r.indexOf(a))}))}connectedCallback(){for(const t of this.querySelectorAll('[role="tablist"] [role="tab"]'))t.hasAttribute("aria-selected")||t.setAttribute("aria-selected","false"),t.hasAttribute("tabindex")||("true"===t.getAttribute("aria-selected")?t.setAttribute("tabindex","0"):t.setAttribute("tabindex","-1"))}}function e(t,e){const r=t.querySelectorAll('[role="tablist"] [role="tab"]'),a=t.querySelectorAll('[role="tabpanel"]'),n=r[e],i=a[e];if(!!t.dispatchEvent(new CustomEvent("tab-container-change",{bubbles:!0,cancelable:!0,detail:{relatedTarget:i}}))){for(const t of r)t.setAttribute("aria-selected","false"),t.setAttribute("tabindex","-1");for(const t of a)t.hidden=!0,t.hasAttribute("tabindex")||t.hasAttribute("data-tab-container-no-tabstop")||t.setAttribute("tabindex","0");n.setAttribute("aria-selected","true"),n.setAttribute("tabindex","0"),n.focus(),i.hidden=!1,t.dispatchEvent(new CustomEvent("tab-container-changed",{bubbles:!0,detail:{relatedTarget:i}}))}}window.customElements.get("tab-container")||(window.TabContainerElement=t,window.customElements.define("tab-container",t));
```
srt32, line 1: I'm trying to think through any problems we might run into when we introduce this change to dotcom. Loading this script will increase the amount of JS loaded and will possibly load this same tab behavior twice on a page. Is that right? Anything else we could / should think about?

manuelpuyol: This JS is not loaded every time the component is called, it is loaded either when the user adds `<%= javascript_include_tag "primer_view_components", type: "module" %>` to their view or (after we publish to NPM) adds `import @primer/view_components`

srt32: I'm at the edge of my understanding but if a project (dotcom?) already includes `'@github/tab-container-element'`, we're still okay?

manuelpuyol: oh, in that case it **will** be required twice
when upgrading the gem in dotcom we have to update imports to those packages

keithamus: dotcom won't be consuming this file directly, instead it will go the _other_ available route, and consume the npm package.

This file, and the `javascript_include_tag` pattern is available for those who _quickly_ want to get PVC up and running without bringing their own JS toolchain.

For those with their own JS toolchain (like dotcom), we will _also_ publish the `@primer/view_components` package. The `@primer/view_components` package will load each dependency as a module so toolchains can apply tree-shaking (aka dead/duplicate code elimination). This package can be imported with `import "@primer/view_components"` where-ever the project sees fit.

In the specific case of dotcom, we'll put it the import statement into our `behaviours` file, and our toolchain will tree-shake out any duplicate imports. It will effectively be a no-op as our `vendor` bundle already contains `@github/tab-container-element`.

keithamus (approved):

app/components/primer/tab_nav_component.rb
```diff
@@ -0,0 +1,92 @@
...
+        @system_arguments[:tag] ||= :a
+        @system_arguments[:role] = :tab
+
+        if selected
+          if @system_arguments[:tag] == :a
+            @system_arguments[:"aria-current"] = :page
+          else
+            @system_arguments[:"aria-selected"] = true
+          end
+        end
```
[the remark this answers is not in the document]
manuelpuyol: updated here https://github.com/primer/view_components/pull/282/commits/399fadd2b9fbe37a752d716b0728120f0da49941

@BlakeWilliams do you have suggestions on how to validate components? I used `before_render` but not sure if that's the correct place to do it

keithamus (approved):
Amazing work :tada:

PR #289, opened 2021-02-25
Update BorderBoxComponent to use SlotsV2

manuelpuyol:
before:

```erb
<%= render(Primer::BorderBoxComponent.new) do |component| %>
  <% component.slot(:header) do %>
    Header
  <% end %>
  <% component.slot(:body) do %>
    Body
  <% end %>
  <% component.slot(:row) do %>
    <% if true %>
      Row one
    <% end %>
  <% end %>
  <% component.slot(:row) do %>
    Row two
  <% end %>
  <% component.slot(:footer) do %>
    Footer
  <% end %>
<% end %>
```

after:

```erb
<%= render(Primer::BorderBoxComponent.new) do |component| %>
  <% component.header do %>
    Header
  <% end %>
  <% component.body do %>
    Body
  <% end %>
  <% component.row do %>
    <% if true %>
      Row one
    <% end %>
  <% end %>
  <% component.row do %>
    Row two
  <% end %>
  <% component.footer do %>
    Footer
  <% end %>
<% end %>
```


srt32:
Could you update the PR body (and maybe the changelog entry) to show the before and after to it's easier to find?

srt32 (approved):

PR #290, opened 2021-02-25
Update Breadcrumb to use SlotsV2

manuelpuyol:
before:

```erb
<%= render(Primer::BreadcrumbComponent.new) do |component| %>
  <% component.slot(:item, href: "/") do %>Home<% end %>
  <% component.slot(:item, href: "/about") do %>About<% end %>
  <% component.slot(:item, selected: true) do %>Team<% end %>
<% end %>
```

after:

```erb
<%= render(Primer::BreadcrumbComponent.new) do |component| %>
  <% component.item(href: "/") do %>Home<% end %>
  <% component.item(href: "/about") do %>About<% end %>
  <% component.item(selected: true) do %>Team<% end %>
<% end %>
```


app/components/primer/breadcrumb/item_component.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Primer
+  module Breadcrumb
+    # This component is part of `Primer::BreadcrumbComponent` and should not be
+    # used as a standalone component.
```
srt32, line 6: Where did we land on definging this class within the parent to avoid this kind of comment?

manuelpuyol: In this case, a separate class is **required** since I need to define a `.erb` template!

srt32: Why can't we do an inline `call`?

app/components/primer/breadcrumb_component.rb
```diff
@@ -3,15 +3,20 @@
...
+    renders_many :items, Primer::Breadcrumb::ItemComponent
 
     # @example auto|Basic
     #   <%= render(Primer::BreadcrumbComponent.new) do |component| %>
-    #     <% component.slot(:item, href: "/") do %>Home<% end %>
-    #     <% component.slot(:item, href: "/about") do %>About<% end %>
-    #     <% component.slot(:item, selected: true) do %>Team<% end %>
+    #     <% component.item(href: "/") do %>Home<% end %>
+    #     <% component.item(href: "/about") do %>About<% end %>
+    #     <% component.item(selected: true) do %>Team<% end %>
```
srt32, line 19: For my own understanding... what's a valid use case where an item that is not the last item would be `selected: true`?

srt32 (approved):
A couple questions but LGTM!

app/components/primer/breadcrumb/item_component.rb
manuelpuyol: the template has html tags, and I don't think we can call them from `call` without calling `html_safe`
I want to avoid calling `html_safe` since it's a smell for me

Also, not sure how `call` would deal with https://github.com/primer/view_components/blob/4ae907d308b955598630430c9982dcdc7b1eafe1/app/components/primer/breadcrumb/item_component.html.erb#L1

srt32: Not blocking. More for fun, to handle the elements, could we use `content_tag` methods to generate the elements? I'm not sure I understand the white spacing issue.

manuelpuyol: >  I understand the white spacing issue.

Me neither, maybe that's something related with how ERB generates the html when there are `ifs` involved.
I'm going to try some stuff here

app/components/primer/breadcrumb_component.rb
manuelpuyol: that... I'm not sure

manuelpuyol: but I don't think slots know about each other

PR #291, opened 2021-02-26
Timeago

keithamus:
### What?

This adds a new `TimeAgo` component based on the `TimeAgoComponent` from dotcom. Small changes have been made to get it to comply with the Primer tests but its 99.9% the same component.

### Why?

This is a useful component to have in the Primer namespace, and should live in Primer ViewComponents where it can better serve users of PVC.

This is also one of the first behavioural components which will help validate JS behaviours in PVC. Similar to https://github.com/primer/view_components/pull/282


test/components/time_ago_component_test.js
```diff
@@ -0,0 +1,12 @@
...
+
+componentTest('time_ago_component', it => {
+
+  it('works', async page => {
+    await expect(page).to.have.selector('body')
+    const el = await page.$('body')
+    await expect(el).to.be.visible()
+  })
+
+})
```
manuelpuyol, line 12: does this file do something? 
I couldn't find `./helpers.js` and don't see `chai` as a dependency

keithamus: Oops this was from an old iteration of the tests!

app/components/time_ago_component.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Primer
+  # A basic wrapper for the "<time-ago>" element.
+  class TimeAgoComponent < Primer::Component
+    def initialize(time:, micro: false, **system_arguments)
```
manuelpuyol, line 6: mind adding some docs here?
e.g: https://github.com/primer/view_components/blob/776d4ecee07f102ef16e41b3245441449d0963db/app/components/primer/avatar_component.rb#L10-L24

app/components/time_ago_component.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Primer
+  # A basic wrapper for the "<time-ago>" element.
```
manuelpuyol, line 4: I think we need something more descriptive here. we normally follow something like `Use TimeAgo to ...`

app/components/time_ago_component.rb
```diff
@@ -0,0 +1,41 @@
...
+  class TimeAgoComponent < Primer::Component
+    def initialize(time:, micro: false, **system_arguments)
+      @system_arguments = system_arguments
+      @system_arguments[:datetime] = time.utc.iso8601
+      @system_arguments[:classes] = class_names("no-wrap", @system_arguments[:classes])
+      @system_arguments[:tag] = "time-ago"
+      @system_arguments[:format] = "micro" if micro
+      @content = if micro
```
manuelpuyol, line 12: nit: might we call this `@time`?
`@content` is an internal ivar from the view_component framework

app/components/time_ago_component.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Primer
+  # A basic wrapper for the "<time-ago>" element.
+  class TimeAgoComponent < Primer::Component
+    def initialize(time:, micro: false, **system_arguments)
+      @system_arguments = system_arguments
+      @system_arguments[:datetime] = time.utc.iso8601
+      @system_arguments[:classes] = class_names("no-wrap", @system_arguments[:classes])
```
manuelpuyol, line 9: do we **always** want `no-wrap`?

app/components/time_ago_component.ts
```diff
@@ -0,0 +1 @@
+import {TimeAgoElement} from '@github/time-elements'
```
manuelpuyol, line 1: for the future, shall we setup eslint for this project?

app/components/time_ago_component.rb
```diff
@@ -0,0 +1,48 @@
...
+module Primer
+  # Use Primer::TimeAgoComponent to display a time relative to how long ago it was
+  class TimeAgoComponent < Primer::Component
+
+    #
+    # @example auto|Default
+    #   <%= render(Primer::TimeAgoComponent.new(time: Time.zone.now)) %>
+    #
+    # @param time [Time] The time to be formatted 
+    # @param micro [Boolean] If true then the text will be formatted in "micro" mode, using as few characters as possible
```
manuelpuyol, line 12:
```suggestion
    # @param micro [Boolean] If true then the text will be formatted in "micro" mode, using as few characters as possible
    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
```

app/components/time_ago_component.ts
keithamus: eslint would be a fantastic idea!

app/components/time_ago_component.rb
keithamus: We had this in the helper/component on GitHub. I think we do. I can't think of a good reason to remove it.

app/components/time_ago_component.rb
keithamus: I've refactored this out more so the initialiser just assigns variables an does not apply logic, the logic is then in a `time_in_words` method.

app/components/time_ago_component.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Primer
+  # Use Primer::TimeAgoComponent to display a time relative to how long ago it was
```
srt32, line 4:
```suggestion
  # Use Primer::TimeAgoComponent to display a time relative to how long ago it was. This component requires JavaScript`.
```

static/statuses.json
```diff
@@ -1 +1,5 @@
...
+>>>>>>> origin/main
```
srt32, line 5: Your next commit should force the docs action to fix this for you.

CHANGELOG.md
```diff
@@ -55,6 +55,10 @@
 
     *Simon Taranto*
 
+* Add `TimeAgo` component.
```
manuelpuyol, line 58: may we move this to be below `# main`?

app/components/primer/time_ago_component.ts
```diff
@@ -0,0 +1 @@
+import '@github/time-elements'
```
manuelpuyol, line 1: shouldn't we only import the time-ago element? `import {TimeAgoElement} from '@github/time-elements'`

app/components/primer/time_ago_component.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Primer
+  # Use Primer::TimeAgoComponent to display a time relative to how long ago it was. This component requires JavaScript.
```
joelhawksley, line 4: Might we want to link to some sort of documentation on how to load the JS?

joelhawksley (approved):
👏🏻

app/components/primer/time_ago_component.ts
keithamus: We should but:
 
 - Our linter complains
 - Or compiler doesn't actually tree shake this as the whole file has side effects meaning no dead code elimination can take place

This is a piece of tech debt we should address upstream and patch in a follow up PR I think.

app/components/primer/time_ago_component.rb
keithamus: Let's tackle this in a follow up PR for all elements requiring JS

srt32 (approved):

PR #292, opened 2021-02-26
Update release scripts taking NPM into consideration

manuelpuyol:
`script/release` will now create the branch and update both the gem and NPM versions
`script/publish` will publish to NPM and RubyGems


CONTRIBUTING.md
```diff
@@ -31,6 +31,7 @@ Here are a few things you can do that will increase the likelihood of your pull
 
 If you are the current maintainer of this gem:
 
+1. Checkout to `main`
```
srt32, line 34:
```suggestion
1. Checkout to `main` and `git pull`
```

script/publish
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# Releasing gem
+bundle exec rake release
+
+# Releasing NPM package
+npm publish --access public
```
srt32, line 7: Will this require the dev to have a token or something locally?

manuelpuyol: This requires the dev to be authorized to NPM locally (and the user needs to have access to `@primer` org)
When running `npm publish` the script tell you what to do if you are missing something

script/release-npm
```diff
@@ -0,0 +1,18 @@
...
+echo "####################################################"
+echo "NPM version:"
+echo "####################################################"
+
+read -p "MAJOR: " major
+read -p "MINOR: " minor
+read -p "PATCH: " patch
+
+# Update package.json version
+sed -E -i.bak "s/\"version\": \"(.*)?\"/\"version\": \"$major.$minor.$patch\"/g" package.json
```
keithamus, line 12: This looks quite fragile. Could we look at shelling out to Ruby or perhaps using something like `jq` here?

keithamus (commented):
What do you think about ensuring the versions between the Ruby gem and the npm package move in lockstep?

manuelpuyol:
> What do you think about ensuring the versions between the Ruby gem and the npm package move in lockstep?

I like this idea! My first attempt was asking if you want to bump major/minor/patch instead of asking the version number. But that required reading the version files to extract the current version and my scripting skills didn't know how to manage it 😅

keithamus:
Oops sorry I've been trashing all over your branch :grimacing:

manuelpuyol:
close in favor of https://github.com/primer/view_components/pull/296

PR #296, opened 2021-03-01
build: rewrite release script

keithamus:
@manuelpuyol this updates #292, rewriting the release script a little, such that you can pick to bump via Major, Minor, Version. It also ensures npm is in lockstep with the ruby gem, which we want to happen, I think.


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@primer/view-components",
-  "version": "0.0.1",
+  "name": "@primer/view_components",
+  "version": "0.0.20",
```
manuelpuyol, line 3:
```suggestion
  "version": "0.0.21",
```

script/release
```diff
@@ -1,43 +1,105 @@
...
+    demo/Gemfile.lock \
+    lib/primer/view_components/version.rb \
+    app/assets/javascripts \
+    package.json \
+    yarn.lock
+}
+
+commit() {
+  git commit -m "release $1.$2.$3"
+  git tag "v$1.$2.$3"
```
manuelpuyol, line 51:
```suggestion
```

Tagging will be handled after merging when publishing the gem

CONTRIBUTING.md
```diff
@@ -31,6 +31,7 @@ Here are a few things you can do that will increase the likelihood of your pull
 
 If you are the current maintainer of this gem:
 
+1. Checkout to `main` and `git pull`
```
srt32, line 34: We could also make the script do this part? Make sure it's on `main` and then `git pull` and confirm no conflicts?

script/release
```diff
@@ -1,43 +1,105 @@
...
+}
+
+push() {
```
manuelpuyol, line 54:
```suggestion
push() {
  git push origin release

```

script/publish
```diff
@@ -0,0 +1,23 @@
...
+# Publish NPM
+read -p "Publish NPM (y/N)? " answer
+case ${answer:0:1} in
+    y|Y )
+        npm publish --access public
+    ;;
+    * )
+        # noop
+    ;;
+esac
```
manuelpuyol, line 23:
```suggestion
bundle exec rake release

# Publish NPM
npm publish --access public
```

manuelpuyol (commented):
The only "issue" I see with this approach is that we'll create a bunch of empty JS releases, since most of our gem releases don't change any JS

script/release
```diff
@@ -1,43 +1,105 @@
...
+  echo "####################################################"
+  echo "Now, open a PR with this branch and merge it to main"
+  echo "Then, run script/publish on main to release the gem and the NPM package"
+  echo "Finally, create a GitHub release https://github.com/primer/view_components/releases/new with the changes from CHANGELOG"
```
srt32, line 57: I wonder if we could automate even more of this if we were okay relying the `gh` CLI?

manuelpuyol:
@srt32 I wonder if we could create an Action similar to the docs build that runs this (adapted) script for us

script/release
```diff
@@ -1,43 +1,105 @@
...
+    echo new version $major $minor $patch
+
+    create_release_branch
+    update_readme $major $minor $patch
+    update_ruby_version $major $minor $patch
+    update_gemfiles $major $minor $patch
+    update_npm $major $minor $patch
+    add_changed_files
+    commit
+    push
```
srt32, line 102: Do we not need `--tags` here?

keithamus: Just removed the tagging behaviour, so not any more.

script/release
manuelpuyol: Yes! That's a next step in my opinion
we could use gh API to create the release and everything

script/release
manuelpuyol: https://github.com/primer/view_components/pull/296#discussion_r584891031

srt32:
> @srt32 I wonder if we could create an Action similar to the docs build that runs this (adapted) script for us

We certainly could! We'd need to be comfy putting creds in there though, right?

manuelpuyol:
> > @srt32 I wonder if we could create an Action similar to the docs build that runs this (adapted) script for us
> 
> We certainly could! We'd need to be comfy putting creds in there though, right?

oh yeah, forgot about that
also we need 2FA to publish the gem 😅 
I'll look more into this later, I want to be as close as possible to how https://github.com/primer/doctocat releases their changes

lib/primer/view_components/version.rb
```diff
@@ -11,3 +11,7 @@ module VERSION
     end
   end
 end
+
+if __FILE__ == $0
+  puts Primer::ViewComponents::VERSION::STRING
+end
```
manuelpuyol, line 17:
```suggestion
puts Primer::ViewComponents::VERSION::STRING if __FILE__ == $PROGRAM_NAME
```

let's make rubocop happy 😄

srt32:
> > > @srt32 I wonder if we could create an Action similar to the docs build that runs this (adapted) script for us
> > 
> > 
> > We certainly could! We'd need to be comfy putting creds in there though, right?
> 
> oh yeah, forgot about that
> also we need 2FA to publish the gem 😅
> I'll look more into this later, I want to be as close as possible to how https://github.com/primer/doctocat releases their changes

Is https://github.com/primer/doctocat/blob/master/.github/workflows/release.yml#L29 not using 2FA? I think we want it to be...

manuelpuyol:
> > > > @srt32 I wonder if we could create an Action similar to the docs build that runs this (adapted) script for us
> > > 
> > > 
> > > We certainly could! We'd need to be comfy putting creds in there though, right?
> > 
> > 
> > oh yeah, forgot about that
> > also we need 2FA to publish the gem 😅
> > I'll look more into this later, I want to be as close as possible to how https://github.com/primer/doctocat releases their changes
> 
> Is https://github.com/primer/doctocat/blob/master/.github/workflows/release.yml#L29 not using 2FA? I think we want it to be...

I'm not sure how they release, but doesn't look like they are using 2FA there

manuelpuyol (approved):
amazing @keithamus !

script/release
```diff
@@ -1,43 +1,165 @@
...
+  fetch
+  if ! remote_history_is_clean; then 
+    echo "Error: changes exist on origin not pulled into this branch. Please pull"
+    exit 1
+  fi 
+
+  if ! local_history_is_clean; then 
+    echo "Error: changes exist that haven't been pushed to origin. Please pull"
+    exit 1
+  fi
```
keithamus, line 120: So it now checks:

 - You're on `main`
 - You have a clean tree, no erroneous files
 - You have pushed up everything
 - You've pulled down everything
 - You're not going to overwrite an existing tag

I've tried to test these as best I can but I guess we'll try running this for real when merged and see how it goes ;)

manuelpuyol: Having some errors in this script is fine, we can start using it and fix it as it goes

PR #297, opened 2021-03-01
build(release): always create a new branch

keithamus:
This will allow us to always push up to a new release branch rather than pushing over the same old one.


manuelpuyol:
@keithamus we also need to update the `push` to match the new branch name

keithamus:
Good point. Do you want to tackle that and get this branch merged? I just clocked off 😂

manuelpuyol (approved):

PR #299, opened 2021-03-01
release 0.0.22

manuelpuyol:


package.json
```diff
@@ -18,7 +18,9 @@
   "bugs": {
     "url": "https://github.com/primer/view_components/issues"
   },
-  "files": ["app/components/primer/*.js"],
+  "files": [
+    "app/components/primer/*.js"
+  ],
```
srt32, line 23: Do we know why this format keeps moving around?

srt32 (approved):

package.json
manuelpuyol: This is `npm version --no-git-tag-version ...` fault, so I think it is better to leave it this way 😅

PR #304, opened 2021-03-02
Fix zeitwerk autoload integration

manuelpuyol:
zeitwerk expected `app/lib/primer/view_helper/dsl.rb` to declare `Primer::ViewHelper::Dsl` instead of `Primer::ViewHelper::DSL`.

In this PR I rename the module and activate zeitwerk in the demo app to catch other errors like this one earlier

### DIFF
--- a/demo/config/application.rb
+++ b/demo/config/application.rb
@@ -16,7 +16,7 @@ Bundler.require(*Rails.groups)
 
 module Demo
   class Application < Rails::Application
-    config.load_defaults 6.0
+    config.load_defaults 6.0 if Rails.version.to_i >= 6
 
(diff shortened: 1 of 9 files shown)


demo/config/application.rb
```diff
@@ -16,6 +16,8 @@
 
 module Demo
   class Application < Rails::Application
+    config.load_defaults 6.0 if Rails.version.to_i >= 6
```
srt32, line 19: What is this change?

manuelpuyol: basically activates `zeitwerk` mode for the demo app
it only works for rails 6+ tho

srt32 (approved):

PR #308, opened 2021-03-02
Promote Base and BlankSlate components to beta

srt32:


manuelpuyol:
I wonder id we want to expose `BaseComponent` to the outside world

srt32:
It gets used in dotcom to make internal components so I think we need it at least for that purpose. What did you have in mind?

manuelpuyol:
hmm I didn't expect it to be used

srt32:
Still okay to promote it?

joelhawksley (approved):
👍🏻 to move this component to `beta`. We can (and should) ensure our documentation guides folks towards correct usages of this component.

PR #311, opened 2021-03-02
update broken stories

manuelpuyol:


stories/primer/dropdown/menu_component_stories.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "primer/dropdown/menu_component"
```
manuelpuyol, line 3: for some reason, the story can't find `Primer::Dropdown::MenuComponent` without this require

joelhawksley (commented):
Is there a failing test we could write that would have caught this problem?

stories/primer/label_component_stories.rb
```diff
@@ -7,7 +7,7 @@ class Primer::LabelComponentStories < ViewComponent::Storybook::Stories
     controls do
       title "this is a label"
       # TODO: Update to use functional scheme mappings
```
srt32, line 9:
```suggestion
```

remove this now?

manuelpuyol:
> Is there a failing test we could write that would have caught this problem?

we could run `bin/rails view_component_storybook:write_stories_json` on CI and see if it breaks?

srt32 (approved):

PR #315, opened 2021-03-03
docs(index): add JS install instructions

keithamus:
This adds some documentation on how to use the JavaScript behaviours that come with Primer ViewComponents.

### DIFF
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -31,7 +31,7 @@ In `config/application.rb`, add **after the application definition**:
 require "primer/view_components/engine"
 ```
 
-To add the JavaScript behaviours, in your `application.html.erb` in the `<head>` tag add:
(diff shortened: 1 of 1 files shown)


docs/content/index.md
```diff
@@ -31,6 +31,18 @@ In `config/application.rb`, add **after the application definition**:
 require "primer/view_components/engine"
 ```
 
+To add the JavaScript behaviours, in your `application.html.erb` in the `<head>` tag add:
```
srt32, line 34:
```suggestion
Optionally, to add the JavaScript behaviours, in your `application.html.erb` in the `<head>` tag add:
```

Could we call out this is only necessary for certain components? I wonder if this is confusing without more explanation. WDYT?

srt32 (approved):

docs/content/index.md
keithamus: I think this is good :+1:

manuelpuyol (approved):

Issue #321, opened 2021-03-03
Primer::OcticonComponent is not respecting `test_selector`


manuelpuyol:
Instead of rendering `data-test-selector='given value'` it renders `test_selector='given value'`

PR #322, opened 2021-03-03
Update OcticonComponent to correctly deal with test_selector

manuelpuyol:
Closes https://github.com/primer/view_components/issues/321

Extracts `TestSelectorHelper` to be used by `Octicon` and `Base`

### DIFF
--- a/app/components/primer/octicon_component.rb
+++ b/app/components/primer/octicon_component.rb
@@ -5,7 +5,8 @@ module Primer
   class OcticonComponent < Primer::Component
     view_helper :octicon
 
-    include Primer::ClassNameHelper
+    include ClassNameHelper
+    include TestSelectorHelper
     include OcticonsHelper
 
     SIZE_DEFAULT = :small
@@ -38,7 +39,7 @@ module Primer
       # Filter out classify options to prevent them from becoming invalid html attributes.
       # Note height and width are both classify options and valid html attributes.
       octicon_helper_options = @system_arguments.slice(:height, :width)
-      @system_arguments = @system_arguments.except(*Primer::Classify::VALID_KEYS, :classes).merge(octicon_helper_options)
+      @system_arguments = add_test_selector(@system_arguments).except(*Primer::Classify::VALID_KEYS, :classes).merge(octicon_helper_options)
     end
 
     def call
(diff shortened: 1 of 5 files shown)


app/lib/primer/test_selector_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+  module TestSelectorHelper
+    TEST_SELECTOR_TAG = :test_selector
+
+    def add_test_selector(args)
+      if args.key?(TEST_SELECTOR_TAG) && !Rails.env.production?
+        args[:data] ||= {}
+        args[:data][TEST_SELECTOR_TAG] = args[TEST_SELECTOR_TAG]
+      end
+
+      args.except(TEST_SELECTOR_TAG)
```
srt32, line 17: I've found this method confusing in that it's called `add_test_selector` but it actually removes it in all envs. I had started poking at this in https://github.com/primer/view_components/pull/275. I don't have a particular recommendation here but curious if any of these sparks thoughts for you.

app/components/primer/octicon_component.rb
```diff
@@ -38,7 +39,7 @@ def initialize(icon:, size: SIZE_DEFAULT, **system_arguments)
       # Filter out classify options to prevent them from becoming invalid html attributes.
       # Note height and width are both classify options and valid html attributes.
       octicon_helper_options = @system_arguments.slice(:height, :width)
-      @system_arguments = @system_arguments.except(*Primer::Classify::VALID_KEYS, :classes).merge(octicon_helper_options)
```
srt32, line 41: Could you help me understand why this is an issue only with octicon?

manuelpuyol: because octicon does not use `Primer::BaseComponent`

srt32: aha! Why do we do that?

srt32 (approved):

app/components/primer/octicon_component.rb
manuelpuyol: because we want to use the `OcticonHelper` to handle the `svg`generation

srt32: Why can't we include that helper were we to inherit from BaseComponent instead?

PR #323, opened 2021-03-03
Status DSL

manuelpuyol:
Extracted our current `self.status` logic to a DSL to make the components cleaner


joelhawksley (approved):
Nice! Love the DSL ❤️

keithamus (approved):
This looks like a great improvement :+1:

app/lib/primer/status/dsl.rb
```diff
@@ -0,0 +1,43 @@
...
+      }.freeze
+
+      class UnknownStatusError < StandardError; end
+
+      included do
+        class_attribute :component_status, instance_writer: false, default: STATUSES[:alpha]
+      end
+
+      class_methods do
+        def status(status = nil)
```
srt32, line 33: Is it typical to have this method be both the writer and reader (depending on if the arg is passed)?

manuelpuyol: I borrowed this idea from the `view_component` framework. They do the same for SlotsV2 where `.slot_name` returns the slot and `.slot_name(xyz)` sets the slot

I thought it was a good way to not break the read API so I didn't have to change tests and dumps

PR #327, opened 2021-03-04
Docs: Add view component js to docs

keithamus:
Currently the docs site does not have any interactive demos for the JS behavioural components. For example if you visit https://primer.style/view-components/components/tabcontainer and try to click one of the tabs, nothing happens.

This adds the JS via the npm package to the docs site, which will enable the JS behaviours.


docs/src/@primer/gatsby-theme-doctocat/components/head.js
```diff
@@ -18,8 +18,9 @@ function Head(props) {
       <meta property="og:image" content={siteMetadata.imageUrl} />
       <meta property="twitter:card" content="summary_large_image" />
       <link href="https://unpkg.com/@primer/css-next@canary/dist/primer.css" rel="stylesheet" />
+      <script src="https://unpkg.com/@primer/view-components@0.0.26/app/assets/javascripts/primer_view_components.js"></script>
```
manuelpuyol, line 21: not sure how I feel about it since we'll have to keep updating the version here.
it also doesn't fix the local development issue

keithamus: Changed it to latest so we dont have to update. Doesn't fix local development but its a step in the right direction :smiley_cat:

manuelpuyol (approved):

PR #329, opened 2021-03-04
Convert Popover to use slotsV2

manuelpuyol:
We want to preserve the current API

```erb
<%= render Primer::PopoverComponent.new do |component| %>
  <% component.heading do %>
    Activity feed
  <% end %>
  <% component.body do %>
    This is the Popover body.
  <% end %>
<% end %>
```

but SlotsV2 does not allow us to manipulate the slot's content in the parent and `heading` is actually rendered inside `body`
because of that, I found a hacky way to inject content while still preserving the API


app/components/primer/popover_component.rb
```diff
@@ -5,37 +5,83 @@ module Primer
...
+      )
+      system_arguments[:p] ||= 4
+      system_arguments[:mt] ||= 2
+      system_arguments[:mx] ||= :auto
+      system_arguments[:text_align] ||= :left
+      system_arguments[:box_shadow] ||= :large
+
+      # This is a hack to allow the parent to set the slot's content
+      @body_arguments = system_arguments
+      ContentComponent.new
```
manuelpuyol, line 56: @BlakeWilliams I'd like your thoughts on this 😄 

What I did here is use the slot to provide the parent `Popover` the arguments passed by the user (used to style the body) and then used `ContentComponent` to only render the `content` provided

Meanwhile, the `Popover` defines the wrapping `body_component` that's going to be used to render the body tag with all the options provided

BlakeWilliams: Interesting! 🤔 

Do we need a `body` slot? Could this be moved to the popover initializer?

BlakeWilliams, line 56:
```suggestion
    renders_one :body, lambda { |caret: CARET_DEFAULT, large: false, **system_arguments, &block|
      system_arguments[:classes] = class_names(
        system_arguments[:classes],
        "Popover-message Box",
        CARET_MAPPINGS[fetch_or_fallback(CARET_MAPPINGS.keys, caret, CARET_DEFAULT)],
        "Popover-message--large" => large
      )
      system_arguments[:p] ||= 4
      system_arguments[:mt] ||= 2
      system_arguments[:mx] ||= :auto
      system_arguments[:text_align] ||= :left
      system_arguments[:box_shadow] ||= :large

      # This is a hack to allow the parent to set the slot's content
      @body_arguments = system_arguments
      block.call
```

I *think* that would work, and be equivalent?

manuelpuyol: will try

manuelpuyol: we do, since users customize the slot with classes, system args and whatnot

BlakeWilliams: I think I'm having a hard time seeing the distinction, since `content` is effectively a component's `body`. So now the component has both a `body`, and `content`.

I still question if the slot makes sense. Maybe the component could accept two different kind of system args? One for the wrapper, and one for the content/body? 🤔

manuelpuyol: we _could_ have the `Popover` component receive a `body` hash argument, which is essentially the same as what the slot is doing.
But I think that the `slot` API is friendlier

app/components/primer/popover_component.rb
```diff
@@ -5,37 +5,83 @@ module Primer
...
+
+      Primer::HeadingComponent.new(**system_arguments)
+    }
+
+    # The body
+    #
+    # @param caret [Symbol] <%= one_of(Primer::PopoverComponent::CARET_MAPPINGS.keys) %>
+    # @param large [Boolean] Whether to use the large version of the component.
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    renders_one :body, lambda { |caret: CARET_DEFAULT, large: false, **system_arguments, &block|
```
srt32, line 41: A lambda accepting a block 🤯

app/components/primer/popover_component.rb
BlakeWilliams: It's the same, but omitting the body slot does introduce less indirection.

I think slots are great for rendering more content on the page, but I'm not sure if they're a good mechanism for accepting and storing arguments.

From an API perspective, I think I would imagine something like the following for the default case:

```erb
<%= render Primer::PopoverComponent.new do |component| %>
 <% component.heading do %>
   Activity feed
 <% end %>

 This is the Popover body.
<% end %>
```

Then modifying the wrapper arguments (maybe):

```erb
<%= render Primer::PopoverComponent.new(wrapper_args: { px: 2, py: 3 }) do |component| %>
 <% component.heading do %>
   Activity feed
 <% end %>

 This is the Popover body.
<% end %>
```

It's more verbosity in the initializer, but that's also partly due to us having two different configurable components rolled into a single component.

I could also see something like this being a pattern:

```erb
<%= render Primer::PopoverComponent
  .new(px: 2, py: 3)
  .wrapper_system_arguments(px: 2, py: 3)
  do |component| %>
 <% component.heading do %>
   Activity feed
 <% end %>

 This is the Popover body.
<% end %>
```

None of these may be the right answer, but I think having the `body` slot replace `content` does change the expectation of what passing a block to a component means in terms of `content`.

manuelpuyol: @srt32 opinions in the API here? I feel like we discussed this option

srt32: My only concern with 

```
<%= render Primer::PopoverComponent.new do |component| %>
 <% component.heading do %>
   Activity feed
 <% end %>

 This is the Popover body.
<% end %>
```

is that it gives us less flexibility to add / change slots later since body is now implicit.

Not blocking but passing the args for the body into the parent does seem surprising

```
Primer::PopoverComponent.new(wrapper_args: { px: 2, py: 3 })
```

docs can solve this problem though.

I am fine with this approach if that lends to a more reasonable implementation.

BlakeWilliams: FWIW I don't think any of my comments should be blocking. Reading/understanding it a bit more, I think my concern might be the HTML output and the component API not being aligned.

as-is we have this component API:

```erb
<%= render Primer::PopoverComponent.new do |component| %>
 <% component.heading do %>
   Activity feed
 <% end %>

 <% component.body %>
   This is the Popover body.
  <% end %>
<% end %>
```

But the HTML output looks will end up like (tags named after where they originate, not the actual output):

```html
<wrapper component-options> <!-- options/tag comes from PopoverComponent.new -->
  <dynamic-component body-options> <!-- This is an explicit `Primer::BaseComponent.new` generated using options from the `body` slot, but does not use body content -->
    <header header-options></header> <!-- This is the header slot, using both content and options -->
    <body-content></body-content> <!-- This is the content from the body, but its options have no impact on it -->
  </dynamic-component>
</wrapper>
```

Given that output, I think I'd expect something like:

```erb
<%= render Primer::PopoverComponent.new do |component| %>
 <% component.popover_body |popover_body| %>
   <% popover_body.heading do %>
     Activity feed
   <% end %>

    This is the Popover body.
  <% end %>
<% end %>
```

It's possible that I'm putting too much emphasis on the component structure mapping to the output structure. I do think there may be some value in the two lining up since that might be the expectation from those implementing these components who don't know the implementation details.

manuelpuyol: We have a lot of options here, I'd say that keeping the API we have now is the safest _for now_, if that feels wrong, we can update it later!

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main
 
+* **Breaking change**: Updates `PopoverComponent` to use Slots V2.
+
+    *Simon Taranto*
```
srt32, line 7: I think this one is yours ;)

srt32 (approved):

PR #330, opened 2021-03-04
Support dark theme on docs

manuelpuyol:
Thanks @colebemis for the help!

This PR adds a way of toggling dark theme in the docs using the `cmd + ctrl + t` shortcut


colebemis:
Woo! I'm glad this worked.

 Might be nice to add some border to the Box:
image

srt32 (approved):
👍 💯 I wonder if we could add an issue to track adding a UI element to toggle the theme. Discovering the keyboard shortcut is very not obvious IME

manuelpuyol:
> 👍 💯 I wonder if we could add an issue to track adding a UI element to toggle the theme. Discovering the keyboard shortcut is very not obvious IME

I think that's intentional since it isn't fully shipped yet?

colebemis:
@srt32 From my PR in Primer React:

> The keyboard shortcut is intentionally undiscoverable because it should only be used to visually test components as we migrate to functional color variables (part of the [theming epic](https://github.com/github/design-systems/issues/1219)). 
> This is a temporary solution. When Primer React Components fully supports color modes, we'll be able to implement dark mode across the entire docs site instead of just the code examples.

https://github.com/primer/components/pull/1095

PR #331, opened 2021-03-05
Update Primer::Classify to convert `color:` to use functional colors

manuelpuyol:
This PR creates a way of remapping colors to their functional counterpart without breaking the API. It's an opt-in configuration so it doesn't break users that still are migrating their PrimerCSS version.

### Implementation

Created the `Primer::Classify::FunctionalColors` base class that implements all the mapping logic, providing the method `functional_color`.

Then, separated a child class for each color property (text and border_color for now) to avoid `FunctionalColors` from having thousands of lines.

Each child class defines its mappings and deprecations, then declared a `color` method that calls `functional_color` with the correct arguments.


app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,58 @@
...
+          return "color-#{val.to_s.dasherize}" if FUNCTIONAL_COLOR_REGEX.match?(val) || ends_with_number?(val)
+          # if the app still allows non functional colors
+          return "text-#{val.to_s.dasherize}" unless force_functional_colors?
+
+          if TEXT_COLOR_MAPPINGS.key?(val)
+            functional_color = TEXT_COLOR_MAPPINGS[val]
+            # colors without functional mapping stay the same
+            return "text-#{val.to_s.dasherize}" if functional_color.blank?
+
+            ActiveSupport::Deprecation.warn("Color #{val} is deprecated. Please use #{functional_color} instead.") if Rails.env.test?
```
srt32, line 39:
```suggestion
            ActiveSupport::Deprecation.warn("Color #{val} is deprecated. Please use #{functional_color} instead.") if !Rails.env.production?
```

maybe this to make more noise outside of CI?

lib/primer/view_components/engine.rb
```diff
@@ -12,6 +12,9 @@ class Engine < ::Rails::Engine
         #{root}/app/lib
       ]
 
+      config.primer_view_component = ActiveSupport::OrderedOptions.new
```
joelhawksley, line 15:
```suggestion
      config.primer_view_components = ActiveSupport::OrderedOptions.new
```

lib/primer/view_components/engine.rb
```diff
@@ -12,6 +12,9 @@ class Engine < ::Rails::Engine
...
+      config.primer_view_component.force_functional_colors = false
```
joelhawksley, line 16:
```suggestion
      config.primer_view_components.force_functional_colors = false
```

app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,58 @@
...
+          raise ArgumentError, "Color #{val} does not exist."
+        end
+
+        def ends_with_number?(val)
+          char_code = val[-1].ord
+          char_code >= 48 && char_code <= 57
+        end
+
+        def force_functional_colors?
+          Rails.application.config.primer_view_component.force_functional_colors
```
joelhawksley, line 53:
```suggestion
          Rails.application.config.primer_view_components.force_functional_colors
```

app/lib/primer/classify/functional_colors.rb
manuelpuyol: I kept it in `test` for now because I wanted to avoid noise at all costs 😅 
but if you think that deprecation warning in `development` is ok, I'm good too

joelhawksley: I think warnings in dev would be nice to have 👍🏻

test/test_helpers/component_test_helpers.rb
```diff
@@ -16,6 +16,13 @@ def without_fetch_or_fallback_raises
       FetchOrFallbackHelper.fallback_raises = true
     end
 
+    def with_force_functional_colors(new_value)
+      old_value = Rails.application.config.primer_view_component.force_functional_colors
+      Rails.application.config.primer_view_component.force_functional_colors = new_value
+      yield
+      Rails.application.config.primer_view_component.force_functional_colors = old_value
```
manuelpuyol, line 23:
```suggestion
      Rails.application.config.primer_view_components.force_functional_colors = new_value
      yield
      Rails.application.config.primer_view_components.force_functional_colors = old_value
```

app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+module Primer
+  class Classify
+    # https://primer-css-git-mkt-color-modes-docs-primer.vercel.app/css/support/v16-migration
+    class FunctionalColors
```
manuelpuyol, line 6: @BlakeWilliams may I get some help with allocations added here? I see the test failing 😬

app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,58 @@
...
+        orange_light: nil,
+        purple: nil,
+        pink: nil,
+        inherit: nil
+      }.freeze
+
+      class << self
+        def text_color(val)
+          # the value is a functional color
+          return "color-#{val.to_s.dasherize}" if FUNCTIONAL_COLOR_REGEX.match?(val) || ends_with_number?(val)
```
BlakeWilliams, line 30:
```suggestion
          return "color-#{val.to_s.dasherize}" if ends_with_number?(val) || FUNCTIONAL_COLOR_REGEX.match?(val) 
```

I think the number check is faster/cheaper than the regex check, thoughts on swapping the conditional around?

app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,58 @@
...
+          return "text-#{val.to_s.dasherize}" unless force_functional_colors?
+
+          if TEXT_COLOR_MAPPINGS.key?(val)
+            functional_color = TEXT_COLOR_MAPPINGS[val]
+            # colors without functional mapping stay the same
+            return "text-#{val.to_s.dasherize}" if functional_color.blank?
+
+            ActiveSupport::Deprecation.warn("Color #{val} is deprecated. Please use #{functional_color} instead.") unless Rails.env.production?
+
+            return text_color(functional_color)
```
BlakeWilliams, line 41: Is it possible to avoid recursion here? Since this is a hot path and the code is changing often, I'd like to reduce the potential for changes to introduce an infinite loop.

app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,58 @@
...
+          end
+
+          raise ArgumentError, "Color #{val} does not exist."
```
BlakeWilliams, line 44: We haven't raised before when given invalid values. Should we log a warning instead?

app/lib/primer/classify/functional_colors.rb
BlakeWilliams: I think the regex is where the additional allocations are coming from, so it's probably fine to bump up the count in the benchmark as long as it isn't increasing the cached version count.

I'd be interested in seeing what a benchmark between this and the current version of classify looks like since we now have to compare each value against a regex.

app/lib/primer/classify/functional_colors.rb
manuelpuyol: it will only reach here if the user opted in to `force_functional_colors`, so I think we are fine raising

app/lib/primer/classify/functional_colors.rb
BlakeWilliams: Actually it looks like we were already running that regex, so we're probably fine benchmark wise. 👍

manuelpuyol: I think we always compared https://github.com/primer/view_components/blob/5a71b6d309a5fba5a684cfeb52e507bed1810d00/app/lib/primer/classify.rb#L207-L208
I just moved that guard here

manuelpuyol: > so it's probably fine to bump up the count in the benchmark as long as it isn't increasing the cached version count

I actually increased by 1, but only locally?

image

BlakeWilliams: Hmm, what version of Ruby?

manuelpuyol: oh, I think I'm on 2.6.3 😅

BlakeWilliams: Ah, I'm on 2.7! That's probably why.

Maybe the benchmark test should raise/not run if you aren't on the "correct" version of Ruby. 🤔

app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,84 @@
...
+        gray: FUNCTIONAL_TEXT_OPTIONS[:secondary],
+        gray_light: FUNCTIONAL_TEXT_OPTIONS[:tertiary],
+        blue: FUNCTIONAL_TEXT_OPTIONS[:link],
+        green: FUNCTIONAL_TEXT_OPTIONS[:success],
+        yellow: FUNCTIONAL_TEXT_OPTIONS[:warning],
+        red: FUNCTIONAL_TEXT_OPTIONS[:danger],
+        white: FUNCTIONAL_TEXT_OPTIONS[:white],
+        # still unsure what will happen with these colors
+        black: nil,
+        orange: nil,
```
srt32, line 30: If someone passes in `orange` (or another unsupported color) what happens? Could we maybe explain the answer in the PR body or in the changelog?

app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,84 @@
...
+          raise ArgumentError, "Color #{val} does not exist."
+        end
+
+        def ends_with_number?(val)
+          char_code = val[-1].ord
+          char_code >= 48 && char_code <= 57
+        end
+
+        def force_functional_colors?
+          Rails.application.config.primer_view_components.force_functional_colors
```
srt32, line 75: I wonder if it'd be better for consumers to be able to opt in at the component level instead of at the app level. For example, in dotcom, it'd be nice to able to opt in for components instead of having to flip a switch for the app. Maybe I'm misunderstanding the goal here?

lib/primer/view_components/engine.rb
```diff
@@ -12,6 +12,10 @@ class Engine < ::Rails::Engine
         #{root}/app/lib
       ]
 
+      config.primer_view_components = ActiveSupport::OrderedOptions.new
+      config.primer_view_components.force_functional_colors = true
```
srt32, line 16: Does this mean we force functional colors by default?

app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,84 @@
...
+        yellow: FUNCTIONAL_TEXT_OPTIONS[:warning],
+        red: FUNCTIONAL_TEXT_OPTIONS[:danger],
+        white: FUNCTIONAL_TEXT_OPTIONS[:white],
+        # still unsure what will happen with these colors
+        black: nil,
+        orange: nil,
+        orange_light: nil,
+        purple: nil,
+        pink: nil,
+        inherit: nil
```
simurai, line 34: Currently Primer CSS still offers the [`.text-inherit`](https://github.com/primer/css/blob/474f27fca4ffd9cbf78e057923648e0b90f3101a/src/utilities/colors.scss#L51-L52). I was debating about removing it, but I think it's still useful and would be good to keep it.

Something we could consider is renaming it from `.text-inherit` to `.color-text-inherit`. But it's not really a color that comes from `primer/primitives` but rather just a way to "reset" a color. So maybe fine to keep as is.

app/lib/primer/classify/functional_colors.rb
manuelpuyol: hmm I guess that'd make it a lot more complex, since `Classify` is used by **all** components.

lib/primer/view_components/engine.rb
manuelpuyol: yes, I think we should always follow Primer's latest implementation, which is functional color in this case

app/lib/primer/classify/functional_colors.rb
joelhawksley: I share Blake's concern. Might we only raise this warning if we aren't in production?

app/lib/primer/classify/functional_colors.rb
joelhawksley: I wonder if we might want to find a way to run this code in tests and then rewrite the values in the codebase, vs. converting these on the fly 🤔

manuelpuyol: I wanted to avoid a breaking change here since they must be extensively used in dotcom. The deprecation warning should help us identify the old uses so we can update it as we go

app/lib/primer/classify/functional_colors.rb
```diff
@@ -0,0 +1,86 @@
...
+            functional_color = TEXT_COLOR_MAPPINGS[val]
+            # colors without functional mapping stay the same
+            return "text-#{val.to_s.dasherize}" if functional_color.blank?
+
+            ActiveSupport::Deprecation.warn("Color #{val} is deprecated. Please use #{functional_color} instead.") unless Rails.env.production? || silence_color_deprecations?
+
+            return "color-#{functional_color.to_s.dasherize}"
+          end
+
+          raise ArgumentError, "Color #{val} does not exist."
```
manuelpuyol, line 68:
```suggestion
          raise ArgumentError, "Color #{val} does not exist." unless Rails.env.production?
```

app/lib/primer/classify/functional_colors.rb
jonrohan: I believe that is the correct behavior, using the presentational class when a functional one doesn't exist.

app/lib/primer/classify/functional_colors.rb
manuelpuyol: done!

app/lib/primer/classify/functional_border_colors.rb
```diff
@@ -0,0 +1,54 @@
...
+        gray: FUNCTIONAL_OPTIONS[:primary],
+        gray_light: FUNCTIONAL_OPTIONS[:secondary],
+        gray_dark: FUNCTIONAL_OPTIONS[:tertiary],
+        blue: FUNCTIONAL_OPTIONS[:info],
+        green: FUNCTIONAL_OPTIONS[:success],
+        yellow: FUNCTIONAL_OPTIONS[:warning],
+        red: FUNCTIONAL_OPTIONS[:danger],
+        white: FUNCTIONAL_OPTIONS[:inverse],
+        # still unsure what will happen with these colors
+        gray_darker: nil,
```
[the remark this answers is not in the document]
manuelpuyol: Removed `nil` mappings here https://github.com/primer/view_components/pull/331/commits/685e009ff53ded3c324c8378179b795c470b4848

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main
 
+* Add `force_functional_colors` option to convert colors to functional.
```
srt32, line 5:
```suggestion
* Add `force_functional_colors` option to convert colors to functional. This change includes a deprecation warning in non-production environments that warns about non functional color usage.
```

PR #336, opened 2021-03-08
Fix deprecated colors in the docs

simurai:
In https://github.com/primer/css/pull/1235 all the deprecated colors got removed to prepare for a Primer CSS `16.0.0` release. To make sure that the docs for Primer ViewComponents still look ok, this PR changes Primer CSS to a pinned version (that still includes the deprecated colors).

Before | After
--- | ---
Screen Shot 2021-03-08 at 15 57 40 | Screen Shot 2021-03-08 at 15 58 18

Example: [Deprecated Labels](https://primer.style/view-components/components/label#deprecated-schemes)

### Alternative

Some of the deprecated classes (like `.Label--orange`) will still [exist on dotcom](https://github.com/github/github/pull/172699) until we find a good replacement. So an alternative to this PR could also be to keep `@primer/css-next@canary` but remove the mentions of the deprecated colors. For example remove this part:

https://github.com/primer/view_components/blob/3bc834471a005819e85b22442449e36e71d0ab0b/app/components/primer/label_component.rb#L40-L42

So it can still be used on dotcom (for now) but it's undocumented.

### DIFF
--- a/docs/src/@primer/gatsby-theme-doctocat/components/head.js
+++ b/docs/src/@primer/gatsby-theme-doctocat/components/head.js
@@ -17,7 +17,7 @@ function Head(props) {
       <meta property="og:description" content={description} />
       <meta property="og:image" content={siteMetadata.imageUrl} />
       <meta property="twitter:card" content="summary_large_image" />
-      <link href="https://unpkg.com/@primer/css-next@canary/dist/primer.css" rel="stylesheet" />
+      <link href="https://unpkg.com/@primer/css-next@0.0.0-7e7b125/dist/primer.css" rel="stylesheet" />
     </Helmet>
   )
 }


manuelpuyol (commented):
I'm ok with pinning the version, but would like to hear what @joelhawksley and @srt32 think about it

manuelpuyol (approved):

joelhawksley:
I'd rather we remove references to deprecated colors from the library entirely, allowing us to track to v16.0.0 of Primer CSS. We can leave the references in dotcom and work to remove them there.

srt32:
Do we need a private (to the github org) set of docs for "here's all the dotcom exceptions to PVC at the moment"?

srt32:
This PR got auto closed. Is that what we wanted?

manuelpuyol:
> ...into a separate DEPRECATED_SCHEME_MAPPINGS or similar?

I've got some ideas on how to deal with deprecated options, gonna work on it soon :)

PR #349, opened 2021-03-09
Simplify ViewHelpers removing the DSL

manuelpuyol:
The DSL approach needed the component to be loaded for the helper to work, which caused issues when the app was lazy loading classes.

## Error example:

`Primer::FlashComponent` used the `primer(:octicon)` view helper. But, in development/test, if you tried to load a page that used `Flash` without referencing `Primer::OcticonComponent` in any other place, it would raise an error `no component defined for helper octicon`.

 ## Now

By removing the DSL, we can manually register helpers that don't depend on the class to be loaded, so they can call it themselves.

## Changes

Also took the opportunity to change `primer(:component)` to `primer_component` @joelhawksley 👀


app/components/primer/blankslate_component.html.erb
```diff
@@ -2,11 +2,7 @@
   <% if spinner.present? %>
     <%= spinner %>
   <% elsif @icon.present? %>
-    <%= primer(:octicon,
-      icon: @icon,
-      size: @icon_size,
-      classes: "blankslate-icon"
-    ) %>
+    <%= primer_octicon icon: @icon, size: @icon_size, classes: "blankslate-icon" %>
```
joelhawksley, line 5: 😍 As you might expect, I'm a fan of `primer_octicon` 😄

joelhawksley (approved):

PR #353, opened 2021-03-09
Link update options

manuelpuyol:
Link will now accept `variant`, `underline` and also `tag: :span`.
Updates made according to https://primer-css-git-mkt-color-modes-docs-primer.vercel.app/css/components/links

image


app/components/primer/link_component.rb
```diff
@@ -5,27 +5,57 @@ module Primer
...
+        "no-underline" => !underline
       )
     end
 
     def call
       render(Primer::BaseComponent.new(**@system_arguments)) { content }
     end
+
+    def before_render
+      raise ArgumentError, "href is required when using <a> tag" if @system_arguments[:tag] == :a && @system_arguments[:href].nil? && !Rails.env.production?
```
srt32, line 58: I wonder if we could call this out more explicitly in the docs string above?

demo/app/views/layouts/application.html.erb
```diff
@@ -4,7 +4,7 @@
     <title>Demo</title>
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
-    <link href="https://unpkg.com/@primer/css-next/dist/primer.css" rel="stylesheet" />
+    <link href="https://unpkg.com/@primer/css-next@canary/dist/primer.css" rel="stylesheet" />
```
srt32, line 7: Could you help me understand why we want this change?

app/components/primer/link_component.rb
srt32: It's not clear to me we'd raise and then raise only in non prod environments. Maybe this?

> # @param href [String] URL to be used for the Link. Required if tag is :a. If the requirements are not met an error will be raised in non production environments. In production, an empty link element will be rendered.

srt32 (approved):

app/components/primer/link_component.rb
```diff
@@ -5,27 +5,57 @@ module Primer
...
+    # @example Without underline
+    #   <%= render(Primer::LinkComponent.new(href: "#", underline: false)) { "Link" } %>
     #
-    # @param href [String] URL to be used for the Link
-    # @param muted [Boolean] Uses light gray for Link color, and blue on hover
+    # @example Span as link
+    #   <%= render(Primer::LinkComponent.new(tag: :span)) { "Span as a link" } %>
+    #
+    # @param tag [String]  <%= one_of(Primer::LinkComponent::TAG_OPTIONS) %>
+    # @param href [String] URL to be used for the Link. Required if tag is `:a`.
```
manuelpuyol, line 35:
```suggestion
    # @param href [String] URL to be used for the Link. Required if tag is `:a`. If the requirements are not met an error will be raised in non production environments. In production, an empty link element will be rendered.
```

PR #354, opened 2021-03-10
build: add type declarations to package.json

keithamus:
### What?

This configures typescript to build the type declaration files. It also configures the package.json to include those files in the published version.

### Why?

@smockle had pointed out we don't actually ship types with the view_components bundle, so this fixes that.


.gitignore
```diff
@@ -19,3 +19,4 @@ demo/config/master.key
 primer_view_components*.gem
 .DS_Store
 app/components/**/*.js
+app/components/**/*.d.ts
```
manuelpuyol, line 22: if we don't commit them, how will publish work?
Will it rely on the local files from the person that ran `script/publish`?

smockle: Yes, the local type declaration files will be published, as the local JS files are currently. 👍

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@
 
     *Manuel Puyol*
 
+* Publish types with npm package.
```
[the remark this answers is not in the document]
smockle: Sure! Fixed in https://github.com/primer/view_components/pull/354/commits/454a1fbf23da7a40c1cece1031e567b05dab2ed6

manuelpuyol:
does `npm publish` run the `prepare` script before publishing? If not, may we update `script/publish` to run `yarn` before `npm publish`?

manuelpuyol (approved):

PR #357, opened 2021-03-10
Update `box_shadow:` to use functional classes

manuelpuyol:
according to https://primer-css-git-mkt-color-modes-docs-primer.vercel.app/css/support/v16-migration#shadow the mapping is 1:1, so we can simply change the classes

### DIFF
--- a/app/lib/primer/classify.rb
+++ b/app/lib/primer/classify.rb
@@ -252,7 +252,7 @@ module Primer
           memo[:classes] << "#{key.to_s.dasherize}#{breakpoint}-n#{val.abs}"
         elsif key == BOX_SHADOW_KEY
           memo[:classes] << if val == true
-                              "box-shadow"
+                              "color-shadow-small"
                             elsif val == :none
                               "box-shadow-none"
                             else
--- a/app/lib/primer/classify.rb
+++ b/app/lib/primer/classify.rb
@@ -253,8 +253,10 @@ module Primer
         elsif key == BOX_SHADOW_KEY
           memo[:classes] << if val == true
                               "box-shadow"
+                            elsif val == :none
+                              "box-shadow-none"
                             else
-                              "box-shadow-#{val.to_s.dasherize}"
+                              "color-shadow-#{val.to_s.dasherize}"
                             end
         elsif key == VISIBILITY_KEY
           memo[:classes] << "v-#{val.to_s.dasherize}"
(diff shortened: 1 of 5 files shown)


app/lib/primer/classify.rb
```diff
@@ -252,9 +252,11 @@ def extract_value(memo, key, val, breakpoint)
           memo[:classes] << "#{key.to_s.dasherize}#{breakpoint}-n#{val.abs}"
         elsif key == BOX_SHADOW_KEY
           memo[:classes] << if val == true
-                              "box-shadow"
+                              "color-shadow-small"
+                            elsif val == :none
```
srt32, line 256: Should we check for `val == false` too?

app/lib/primer/classify.rb
```diff
@@ -252,9 +252,11 @@ def extract_value(memo, key, val, breakpoint)
...
+                              "box-shadow-none"
                             else
-                              "box-shadow-#{val.to_s.dasherize}"
+                              "color-shadow-#{val.to_s.dasherize}"
```
srt32, line 259: I would have expected some spec to change or fail. Is that something we could add?

manuelpuyol: it probably will, I totally forgot about the spec 😅

app/lib/primer/classify.rb
manuelpuyol: hmm we never used it, but don't see why not

test/lib/classify_test.rb
```diff
@@ -322,11 +322,13 @@ def test_font_weight
   end
 
   def test_box_shadow
-    assert_generated_class("box-shadow",             { box_shadow: true })
-    assert_generated_class("box-shadow-medium",      { box_shadow: :medium })
-    assert_generated_class("box-shadow-large",       { box_shadow: :large })
-    assert_generated_class("box-shadow-extra-large", { box_shadow: :extra_large })
-    assert_generated_class("box-shadow-none",        { box_shadow: :none })
+    assert_generated_class("color-shadow-small",       { box_shadow: true })
+    assert_generated_class("color-shadow-small",       { box_shadow: true })
```
srt32, line 326:
```suggestion
    assert_generated_class("color-shadow-small",       { box_shadow: true })
```

dup line?

srt32 (approved):

PR #359, opened 2021-03-11
Apply all avatar classes to the link wrapper if applicable

laserlemon:
If an `href` argument is specified, the `AvatarComponent` wraps the `img` element in an `a` element, applying the `avatar` class to the `a` instead. These changes make sure that the other optional avatar-related classes like `avatar--small` and `circle` are also applied to the `a` (if present) rather than the `img`. Otherwise, you get a square `a` around a circle `img`.

| Before | After |
|:---:|:---:|
| Screen Shot 2021-03-11 at 9 24 04 AM | Screen Shot 2021-03-11 at 9 24 43 AM |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,10 +30,6 @@
 
     *Keith Cirkel* & *Clay Miller*
 
-* **Breaking change**: Updates `PopoverComponent` to use Slots V2.
-
-    *Manuel Puyol*
-
 * Fix `AvatarComponent` to apply classes to the link wrapper if present.
 
     *Steve Richert*
@@ -41,6 +37,10 @@
 * Fix `AvatarComponent` to apply the `avatar-small` class rather than `avatar--small`.
 
     *Steve Richert*
+
+* **Breaking change**: Updates `PopoverComponent` to use Slots V2.
+
+    *Manuel Puyol*
 
 ## 0.0.27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@
 
     *Manuel Puyol*
 
+* Fix `AvatarComponent` to apply classes to the link wrapper if present.
+
+    *Steve Richert*
+
 ## 0.0.27
 
 * Promote `BreadcrumbComponent` and `ProgressBarComponent` to beta status.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@
 
     *Steve Richert*
 
+* Fix `AvatarComponent` to apply the `avatar-small` class rather than `avatar--small`.
+
+    *Steve Richert*
+
 ## 0.0.27
 
 * Promote `BreadcrumbComponent` and `ProgressBarComponent` to beta status.
--- a/app/components/primer/avatar_component.rb
+++ b/app/components/primer/avatar_component.rb
@@ -34,17 +34,25 @@ module Primer
       @system_arguments[:height] = size
       @system_arguments[:width] = size
 
-      @system_arguments[:classes] = class_names(
-        system_arguments[:classes],
-        "avatar" => !href,
+      avatar_class_names = class_names(
+        "avatar",
         "avatar--small" => size < SMALL_THRESHOLD,
         "circle" => !square
       )
+
+      if @href
+        @link_arguments = { href: href, classes: avatar_class_names }
+      else
+        @system_arguments[:classes] = class_names(
+          system_arguments[:classes],
+          avatar_class_names
+        )
+      end
     end
 
     def call
       if @href
-        render(Primer::LinkComponent.new(href: @href, classes: "avatar")) do
+        render(Primer::LinkComponent.new(**@link_arguments)) do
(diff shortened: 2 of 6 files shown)


test/components/avatar_component_test.rb
```diff
@@ -46,7 +46,7 @@ def test_squared_avatar
     render_inline(Primer::AvatarComponent.new(src: "https://github.com/github.png", alt: "github", square: true))
 
     assert_selector("img.avatar")
-    refute_selector(".CircleBadge")
+    refute_selector(".circle")
```
laserlemon, line 49: It appeared that this should have been changed when AvatarComponent moved from using the CircleBadge class to circle.

app/components/primer/avatar_component.rb
```diff
@@ -34,17 +34,25 @@ def initialize(src:, alt:, size: 20, square: false, href: nil, **system_argument
       @system_arguments[:height] = size
       @system_arguments[:width] = size
 
-      @system_arguments[:classes] = class_names(
-        system_arguments[:classes],
-        "avatar" => !href,
+      avatar_class_names = class_names(
```
manuelpuyol, line 37: I think we can keep  `system_arguments[:classes]` here, so If an user wants to add a custom class, it adds the class to the `<a>` instead of ``.
At least it looks like that'd be the correct behavior since we are adding all classes to `<a>`

manuelpuyol (commented):
Thanks for the PR! I only have one suggestion there, but otherwise this looks great 😄 
Mind adding a CHANGELOG entry?

app/components/primer/avatar_component.rb
laserlemon: I wondered about that. I'll make the change and update the `CHANGELOG`. Thanks for the quick review! 💛

app/components/primer/avatar_component.rb
```diff
@@ -34,17 +34,25 @@ def initialize(src:, alt:, size: 20, square: false, href: nil, **system_argument
...
       )
+
+      if @href
+        @link_arguments = { href: href, classes: avatar_class_names }
+      else
+        @system_arguments[:classes] = class_names(
+          system_arguments[:classes],
+          avatar_class_names
+        )
+      end
```
manuelpuyol, line 50:
```suggestion
      avatar_class_names = class_names(
        "avatar",
        system_arguments[:classes],
        "avatar--small" => size < SMALL_THRESHOLD,
        "circle" => !square
      )

      if @href
        @link_arguments = { href: href, classes: avatar_class_names }
      else
        @system_arguments[:classes] = avatar_class_names
      end
```

this is what I had in mind

app/components/primer/avatar_component.rb
manuelpuyol: this is roughly what I had in mind

```rb
avatar_class_names = class_names(
  "avatar",
  system_arguments[:classes],
  "avatar--small" => size < SMALL_THRESHOLD,
  "circle" => !square
)

if @href
  @link_arguments = { href: href, classes: avatar_class_names }
  @system_arguments[:classes] = nil # otherwhise it will apply the custom classes twice
else
  @system_arguments[:classes] = avatar_class_names
end
```

manuelpuyol: if we go this route, I'd ❤️  to see a test showing the custom class behavior too! :)

laserlemon:
@manuelpuyol I addressed your feedback and tweaked the implementation to be non-destructive to the given arguments. I also added a bit of test coverage around setting link wrapper's `href`.

app/components/primer/avatar_component.rb
```diff
@@ -33,22 +33,21 @@ def initialize(src:, alt:, size: 20, square: false, href: nil, **system_argument
...
     def call
       if @href
-        render(Primer::LinkComponent.new(href: @href, classes: "avatar")) do
-          render(Primer::BaseComponent.new(**@system_arguments)) { content }
+        render(Primer::LinkComponent.new(href: @href, classes: @system_arguments[:classes])) do
+          render(Primer::BaseComponent.new(**@system_arguments.except(:classes))) { content }
         end
       else
-        render(Primer::BaseComponent.new(**@system_arguments)) { content }
+        render(Primer::BaseComponent.new(classes: @classes, **@system_arguments)) { content }
```
laserlemon, line 50: Shoot, I missed this in my refactor. Must add a test the exposes this!

laserlemon: Fixed!

CHANGELOG.md
```diff
@@ -34,6 +34,10 @@
 
     *Manuel Puyol*
 
+* Fix `AvatarComponent` to apply classes to the link wrapper if present.
```
manuelpuyol, line 37: mind moving this up? we tend to group breaking changes in the bottom

manuelpuyol (approved):
✨

CHANGELOG.md
laserlemon: 👍

manuelpuyol (approved):

PR #365, opened 2021-03-11
Create codescanning workflow

srt32:
This PR adds a code scanning workflow for javascript (ruby is not yet supported). Shall we experiment with running this workflow for a bit?

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'ruby', 'javascript' ]
+        language: [ 'javascript' ]
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'ruby', 'javascript' ]
```
srt32, line 20:
```suggestion
        language: [ 'javascript' ]
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,38 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
```
manuelpuyol, line 5: is this the correct syntax?

srt32: I think I pulled this from the sample config they gave me so I think so?!

manuelpuyol (approved):
not an expert on this, but looks good

PR #371, opened 2021-03-12
update primerCSS to use @16.0.0-rc.873c0d1

manuelpuyol:
Since PrimerCSS now has rc1 for v16, we should use it instead of `css-next`


srt32 (approved):
It'd be really nice if we could extract this value into a single config value somewhere...

colebemis (approved):

PR #373, opened 2021-03-12
Add action to sync Component status project board

srt32:
This action runs on pushes to main and updates the cards on https://github.com/primer/view_components/projects/3 so that they are on the correct column.

The "ideas" column is not under automation and people can add ideas there. The other status columns are used in automation and hold cards in their current status.

We use the <ComponentName --- > as an id on the card to keep track of them. We can add more info after this separator such as links to tracking issues. 

The script assumes a handful of things:

* the column names match the status names
* the card starts with the component name
* the project board number (3) is hardcoded

image

### DIFF
--- a/.github/workflows/statuses.yml
+++ b/.github/workflows/statuses.yml
@@ -16,9 +16,6 @@ jobs:
         uses: actions/setup-ruby@v1
         with:
           ruby-version: 2.7.x
-      - uses: actions/cache@v2
-        with:
-          path: vendor/bundle
-          key: gems-build-docs-${{ hashFiles('**/Gemfile.lock') }}
       - name: Rewrite code
+        run: gem install graphql-client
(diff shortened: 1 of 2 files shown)


.github/workflows/statuses.yml
```diff
@@ -0,0 +1,23 @@
+name: Docs Build
+
+on:
+  push:
+    branches:
+      - master
+      - st-statuses-action
```
srt32, line 7: This was for testing but since I'm on a fork, this doesn't work anyway, it seems.

srt32:
```suggestion
```

srt32:
@joelhawksley @manuelpuyol up to give this a go?

manuelpuyol (approved):

PR #375, opened 2021-03-15
Fix popover with HTML body

manuelpuyol:
Running `block.call` was only returning the last line of the HTML as content.
This was probably caused because we were evaluating the content before rendering the component.
To fix that, we have to run `view_context.capture`


BlakeWilliams (approved):
LGTM, good find!

PR #382, opened 2021-03-15
Fix syntax error in statuses action

srt32:
and pass the token directly.

### DIFF
--- a/.github/workflows/statuses.yml
+++ b/.github/workflows/statuses.yml
@@ -9,14 +9,13 @@ jobs:
   build-docs:
     name: Update statuses project
     runs-on: ubuntu-latest
-    env:
-      GITHUB_TOKEN: {{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v2
       - name: Setup Ruby
         uses: actions/setup-ruby@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/statuses.yml
```diff
@@ -4,19 +4,19 @@ on:
   push:
     branches:
       - main
+      - srt32-patch-2
```
srt32, line 7:
```suggestion
```

.github/workflows/statuses.yml
```diff
@@ -9,14 +9,13 @@ jobs:
...
       - name: Setup Ruby
         uses: actions/setup-ruby@v1
         with:
           ruby-version: 2.7.x
-      - name: Make updates
+      - name: Install deps
         run: gem install graphql-client
-        run: script/update-statuses-project.rb
+      - name: Make updates
+        run: GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} script/update-statuses-project.rb
```
srt32, line 21: I am sure there are other ways to pass this env var in but I could not figure them out.... Very open to feedback here!

manuelpuyol (approved):

PR #383, opened 2021-03-15
correctly point to iframe.html

manuelpuyol:
Build the JS to point to `view-components/stories/iframe.html` instead of relying in rails redirect


srt32 (approved):
LGTM if we can get ci happy.

Issue #377, opened 2021-03-15
Align font size with other Primer doc sites


yaili:
The base font size in the ViewComponents documentation site is smaller than on other Primer doc sites. This means that when going from one site to another (for example from [design guidelines](https://primer.style/design/) to [VC](https://primer.style/view-components/)) things look zoomed, and things move around a bit, especially in larger screens.

It would be great if we could keep the font size consistent.

PR #384, opened 2021-03-16
Reset body style to use font-size: 16px

manuelpuyol:
Closes https://github.com/primer/view_components/issues/377

Requiring PrimeCSS was changing the docs font-size to 14px. This is a quick fix to revert it back to 16px


srt32 (approved):
Using PrimerCSS for the docs site as well as for the examples is 🤯

PR #385, opened 2021-03-16
[WIP] Separate classes for each type of Button

manuelpuyol:


manuelpuyol:
@joelhawksley we discussed about multiple button components before, wdyt? I'm not sure how I feel about it...

joelhawksley:
> @joelhawksley we discussed about multiple button components before, wdyt? I'm not sure how I feel about it...

Just saw this! I'd love to discuss this as a group. Perhaps we could chat about it at the next weekly meeting? Unless you think it's urgent :)

srt32:
My first reaction is we need to update how the docs work ;)

I think we'd want to visually group all the "buttons" together in some way. Perhaps related we may also want to somehow hide or obscure the internal (subclasses) components.

Not blocking but I think also a time to rethink the docs hierarchy.

manuelpuyol:
Closing this for now since we landed the `button_type` -> `scheme` change

Issue #316, opened 2021-03-03
Redesign UnderlineNavComponent API


srt32:
Follow up to https://github.com/primer/view_components/pull/313#pullrequestreview-603117466

manuelpuyol:
Extra: we could find a way of merging some of the logic with `TabNavComponent` since they are so similar

PR #391, opened 2021-03-17
Update UnderlineNav API

manuelpuyol:
Closes https://github.com/primer/view_components/issues/316

Redesigns the component API to make better use of slots, making the API harder to use incorrectly.
Before, the `body` slot worked as a `content_area`, only allowing content to be added inside it.

Now, the user can use `tab` slots similarly to `TabNav`. Also supporting a TabContainer wrapper!

Before:

```erb
<%= render(Primer::UnderlineNavComponent.new) do |component| %>
  <% component.body do %>
    <%= render(Primer::LinkComponent.new(href: "#url")) { "Item 1" } %>
  <% end %>
  <% component.actions do %>
    <%= render(Primer::ButtonComponent.new) { "Button!" } %>
  <% end %>
<% end %>
```

After:

```erb
<%= render(Primer::UnderlineNavComponent.new) do |component| %>
  <% component.tab(href: "#", selected: true) do %>
    Item 1
  <% end %>
  <% component.tab(href: "#") do %>
    Item 2
  <% end %>
  <% component.actions do %>
    <%= render(Primer::ButtonComponent.new) { "Button!" } %>
  <% end %>
<% end %>
```


stories/primer/underline_nav_component/tab_component_stories.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class Primer::UnderlineNavComponent::TabComponentStories < ViewComponent::Storybook::Stories
+  layout "storybook_preview"
+
+  story(:tab) do
```
srt32, line 6: This is a sub story of the underline nav component or it's its own thing?

manuelpuyol: it is a substory! It will be inside the `UnderlineNav` folder

srt32 (approved):

demo/test/components/previews/primer/underline_nav_component_preview.rb
```diff
@@ -0,0 +1,20 @@
+module Primer
+  class UnderlineNavComponentPreview < ViewComponent::Preview
+    def default
+      render(Primer::UnderlineNavComponent.new(with_panel: true)) do |c|
+        c.tab(selected: true) do |t|
+          t.panel { "Panel 1" }
+          "Tab 1"
```
joelhawksley, line 7: How do you feel about mixing slots with a block like this? What if we instead did something like:

```suggestion
          t.panel { "Panel 1" }
          t.label { "Tab 1" } 
```

manuelpuyol: I wanted to avoid another sub-slot, since most of the use cases **aren't** using `tab-container`, so 90% of the time we won't even call `t.panel`

manuelpuyol: @joelhawksley I'm going to merge as-is and work in the improvements we discussed in the follow-up PR https://github.com/primer/view_components/pull/393

PR #392, opened 2021-03-17
remove unused `Primer::Slot` class

manuelpuyol:
This class was used for our SlotsV1 implementation. Since we migrated all components to SlotsV2, it is no longer needed


srt32 (approved):
🎉

PR #393, opened 2021-03-17
Extract logic from `UnderlineNav` and `TabNav` components

manuelpuyol:
`TabNav` and `UnderlineNav` are two very similar components and they have a similar goal too!
With that in mind, it made sense to make their API as close as possible.
During this refactor, I noticed that their `tab` slot was virtually the same, so I extracted `Primer::Navigation::TabComponent` that can be used by both, avoiding duplicating logic.

In this PR, I extracted `Primer::TabbedComponentHelper` that allows both components to share logic, like the tab validation and a wrapper that knows if `TabContainer` should be added or not.

## Breaking change

Part of this PR updates the API for `TabComponent`, adding slots for each part of the component (but still allowing custom content).
The new API has 4 slots:

1. `icon` (placed at the left)
2. `counter` (placed at the right)
3. `title` (title of the tab)
4. `panel` (content of the panel associated with the tab)

This will make building the tabs more strict and less prone to errors.
I also investigated current uses of tabs and noted that some cases still use custom elements and contents (such as emojis) or other structures (like adding more tags).
So, to support this, the component will also allow the use of `content` so you can add any HTML you want - given that no slot is used.


app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Primer
+  module Navigation
+    # This component is part of navigation components such as `Primer::TabNavComponent`
+    # and `Primer::UnderlineNavComponent` and shoult not be used as a standalone component.
```
srt32, line 6:
```suggestion
    # and `Primer::UnderlineNavComponent` and should not be used as a standalone component.
```

app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,45 @@
...
+      renders_one :panel, lambda { |**system_arguments|
+        system_arguments[:tag] ||= :div
+        system_arguments[:role] ||= :tabpanel
+        system_arguments[:hidden] = true unless @selected
+
+        Primer::BaseComponent.new(**system_arguments)
+      }
+
+      attr_reader :selected
+      def initialize(selected: false, with_panel: false, **system_arguments)
```
srt32, line 19:
```suggestion

      def initialize(selected: false, with_panel: false, **system_arguments)
```

app/components/primer/tab_nav_component.rb
```diff
@@ -3,34 +3,47 @@
...
     }
 
     # @example Default
     #   <%= render(Primer::TabNavComponent.new) do |c| %>
-    #     <% c.tab(selected: true, title: "Tab 1", href: "#") %>
-    #     <% c.tab(title: "Tab 2", href: "#") %>
-    #     <% c.tab(title: "Tab 3", href: "#") %>
+    #     <% c.tab(selected: true, href: "#") { "Tab 1" }%>
+    #     <% c.tab(href: "#") { "Tab 2" } %>
+    #     <% c.tab(href: "#") { "Tab 3" } %>
```
srt32, line 24: This is a much nicer API ;)

srt32 (approved):

Rakefile
```diff
@@ -273,43 +274,6 @@ namespace :docs do
             end
           end
         end
-
-        # Slots V1 docs
```
manuelpuyol, line 277: removed because we don't have any slotsv1 remaining 🎉

Rakefile
```diff
@@ -249,7 +250,7 @@ namespace :docs do
 
             if slot_documentation.base_docstring.present?
               f.puts
-              f.puts(slot_documentation.base_docstring)
+              f.puts(view_context.render(inline: slot_documentation.base_docstring))
```
manuelpuyol, line 253: updated so `link_to_component` works in the slot description

app/components/primer/navigation/tab_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render Primer::BaseComponent.new(**@system_arguments) do %>
+  <%= icon %>
+  <%= title %>
+  <%= counter %>
```
srt32, line 4: I would have expected to see `panel` in here somewhere. How does this work?  ;)

app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,105 @@
...
+          @icon_classes,
+          system_arguments[:classes]
+        )
+        Primer::OcticonComponent.new(**system_arguments)
+      }
+
+      # The Tab's title.
+      #
+      # @param kwargs [Hash] The same arguments as <%= link_to_component(Primer::TextComponent) %>.
+      renders_one :title, Primer::TextComponent
```
joelhawksley, line 35: What do you think about naming this something else? `title` is a valid argument to `content_tag` and thus might be nice to have as an option here.

This brings up a more general concern: I don't think we have a solid set of reserved words established for this library. Perhaps we could look into that soon?

app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,105 @@
...
+        if @system_arguments[:tag] == :a
+          @system_arguments[:"aria-current"] = :page
+        else
+          @system_arguments[:"aria-selected"] = true
+        end
+      end
+
+      def render_content?
+        !icon && !title && !counter
+      end
```
joelhawksley, line 102: 🤔 I wonder if this is the API we want to have here. I could see us wanting to allow the use of the `icon` and `counter` slots with block content. What do you think?

app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,105 @@
...
+      # @param kwargs [Hash] The same arguments as <%= link_to_component(Primer::OcticonComponent) %>.
+      renders_one :icon, lambda { |**system_arguments|
+        system_arguments[:classes] = class_names(
+          @icon_classes,
+          system_arguments[:classes]
+        )
+        Primer::OcticonComponent.new(**system_arguments)
+      }
+
+      # The Tab's title.
```
joelhawksley, line 32: Maybe we call this `label`?

app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,105 @@
+# frozen_string_literal: true
+
+module Primer
+  module Navigation
+    # This component is part of navigation components such as `Primer::TabNavComponent`
+    # and `Primer::UnderlineNavComponent` and should not be used as a standalone component.
```
joelhawksley, line 6:
```suggestion
    # This component is part of navigation components such as `Primer::TabNavComponent`
    # and `Primer::UnderlineNavComponent` and should not be used by itself.
```

app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,105 @@
...
+    class TabComponent < Primer::Component
+      include ViewComponent::SlotableV2
+
+      # Panel controlled by the Tab. It will only renderd if `with_panel` is `true`.
```
joelhawksley, line 10: So if we pass a panel, but `with_panel` is `false`, what happens?

app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,105 @@
...
+      #   <%= render(Primer::Navigation::TabComponent.new(selected: true)) do |c| %>
+      #     <% c.title { "Selected" } %>
+      #   <% end %>
+      #   <%= render(Primer::Navigation::TabComponent.new) do |c| %>
+      #     <% c.title { "Not selected" } %>
+      #   <% end %>
+      #
+      # @example With icons and counters
+      #   <%= render(Primer::Navigation::TabComponent.new) do |c| %>
+      #     <% c.icon(icon: :star) %>
```
joelhawksley, line 54: Gosh, I really wish `OcticonComponent` took positional arguments. This would be cleaner as:

```suggestion
      #     <% c.icon(:star) %>
```

app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,105 @@
...
+      renders_one :title, Primer::TextComponent
+
+      # Counter to be rendered in the Tab right.
+      #
+      # @param kwargs [Hash] The same arguments as <%= link_to_component(Primer::CounterComponent) %>.
+      renders_one :counter, Primer::CounterComponent
+
+      attr_reader :selected
+
+      # @example Default
```
joelhawksley, line 44: If this component isn't meant to be used by itself, do you think we should have examples for it?

app/components/primer/navigation/tab_component.rb
```diff
@@ -0,0 +1,105 @@
...
+      #     </div>
+      #   <% end %>
+      #
+      # @param selected [Boolean] Whether the Tab is selected or not.
+      # @param with_panel [Boolean] Whether the Tab has an associated panel.
+      # @param icon_classes [Boolean] Classes that must always be applied to icons.
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+      def initialize(selected: false, with_panel: false, icon_classes: "", **system_arguments)
+        @selected = selected
+        @icon_classes = icon_classes
```
joelhawksley, line 80: Can you explain the use case for `icon_classes`? In what situation do we need this argument?

app/lib/primer/tabbed_component_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+require "active_support/concern"
+
+module Primer
+  # Helper to share tab validation logic between components.
+  # The component will raise an error if there are 0 or 2+ selected tabs.
+  module TabbedComponentHelper
+    extend ActiveSupport::Concern
+
+    included do
+      include ViewComponent::SlotableV2
```
joelhawksley, line 12: If we update do `view_component` `v2.28.0`, this will no longer be needed 😄

app/lib/primer/tabbed_component_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      render Primer::TabContainerComponent.new do
+        yield
+      end
+    end
+
+    def validate_single_selected_tab
+      raise MultipleSelectedTabsError, "only one tab can be selected" if selected_tabs_count > 1
+      raise NoSelectedTabsError, "a tab must be selected" if selected_tabs_count != 1
+    end
```
joelhawksley, line 35: How do you feel about raising this error these days? I know we've gone back and forth about the risk of runtime exceptions in production.

app/components/primer/navigation/tab_component.html.erb
manuelpuyol: the panel position is implemented by the parent, because according to `https://github.com/github/tab-container-element` the panels are siblings to the `nav`
https://github.com/primer/view_components/blob/e0548ac0cc09a982c2858596b548e43ed5f9e459/app/components/primer/tab_nav_component.html.erb#L2-L14

app/components/primer/navigation/tab_component.rb
manuelpuyol: I think I have to update this description to be more clear
some context here https://github.com/primer/view_components/pull/393/files#r597718804

app/components/primer/navigation/tab_component.rb
manuelpuyol: I wanted to avoid `label` because it could be confused with `Primer::LabelComponent`

app/components/primer/navigation/tab_component.rb
manuelpuyol: that's a good idea!

app/components/primer/navigation/tab_component.rb
manuelpuyol: I think it is helpful, because someone implementing `TabNav` or `UnderlineNav` have to know how the tabs work.

app/components/primer/navigation/tab_component.rb
manuelpuyol: So, `UnderlineNav` requires the `UnderlineNav-octicon` class to be used with icons, so we have something like this:
https://github.com/primer/view_components/blob/e0548ac0cc09a982c2858596b548e43ed5f9e459/app/components/primer/underline_nav_component.rb#L22-L27

That class isn't hardcoded here because it would also affect `TabNav`, so I left it configurable to the caller
If we didn't have this, we'd require the user to call `c.icon(icon: :x, classes: "UnderlineNav-octicon")`, which would be repetitive and most people would forget about it

app/components/primer/navigation/tab_component.rb
manuelpuyol: The idea with `content` is a "you can do whatever you want"
so if someone wants to add some content **and** an icon, they'd need to pass the icon in their content, e.g:

```
c.tab do 
  Primer::Octicon..
  rest of the content
end
```

app/components/primer/navigation/tab_component.rb
manuelpuyol: https://github.com/primer/view_components/pull/393/commits/284421bf030150876541299b21a8bc4e7741b772

app/components/primer/navigation/tab_component.rb
manuelpuyol: we could potentially change
```rb
    def initialize(icon:, size: SIZE_DEFAULT, **system_arguments)
      @icon = icon
```

to

```rb
    def initialize(icon_name = nil, icon: nil, size: SIZE_DEFAULT, **system_arguments)
      @icon = icon_name || icon
```

which would avoid breaking changes?

app/lib/primer/tabbed_component_helper.rb
manuelpuyol: Yes!
But at the same time, the gem technically supports `view_component >= 2.0 && < 3`
Removing `include ViewComponent::SlotableV2` would drop support for anything `< 2.28`

What we could do is move this include to `Primer::Component` which would make things easier to migrate or even only include according to the `view_component` version

app/lib/primer/tabbed_component_helper.rb
manuelpuyol: Yeah, this is something we need to decide sooner than later.
Maybe what we need is a `validation` API that we make sure to always skip in production

app/components/primer/navigation/tab_component.rb
joelhawksley: `text`?

app/components/primer/navigation/tab_component.rb
joelhawksley: Gotcha. I think it would be nice to allow people to pass a block and still use the built-in OcticonComponent slot. So perhaps this could just be `!title`?

app/components/primer/navigation/tab_component.rb
joelhawksley: I like that. Could you file an issue?

app/lib/primer/tabbed_component_helper.rb
joelhawksley: > What we could do is move this include to Primer::Component which would make things easier to migrate or even only include according to the view_component version.

Let's do that. Maybe we could couple that with a tracking issue on this repo for things to do once we upgrade to VC v3?

app/lib/primer/tabbed_component_helper.rb
joelhawksley: Yeah, like `fetch_or_fallback` 😄

manuelpuyol: https://github.com/primer/view_components/issues/395

joelhawksley (approved):

PR #398, opened 2021-03-23
Adjust spinner to use CSS animations instead of SMIL.

theojulienne:
The current implementation of `SpinnerComponent` uses an SVG with SMIL animations to render a spinner that spins indefinitely. This unfortunately triggers a [bug in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1039243) that is very frustrating for users, where a page with SVGs with infinitely-repeating SMIL animations that is placed in the background will trigger a linear task to emit, then run, events for each frame that would have been run, as soon as the page is returned to the foreground. This execution is linear to the number of SVGs with repeating SMIL animations multiplied by the duration the tab is in the background. For GitHub.com webpages, this can take seconds, or even minutes, where the Chrome tab is frozen, or even locking up the entire OS if memory is limited.

This PR adjusts the SpinnerComponent to use a new `anim-rotate-360` CSS class that uses CSS animations to perform the same animation as the original SMIL `animateTransform` tag.

This PR is dependent on https://github.com/primer/css/pull/1251, which implements the CSS class itself.

### DIFF
--- a/app/components/primer/spinner_component.rb
+++ b/app/components/primer/spinner_component.rb
@@ -31,10 +31,7 @@ module Primer
       @system_arguments = system_arguments
       @system_arguments[:tag] = :svg
       @system_arguments[:style] ||= DEFAULT_STYLE
-      @system_arguments[:classes] = class_names(
-        @system_arguments[:classes],
-        "anim-rotate"
-      )
+      @system_arguments[:animation] = :rotate
       @system_arguments[:width] = SIZE_MAPPINGS[fetch_or_fallback(SIZE_OPTIONS, size, DEFAULT_SIZE)]
       @system_arguments[:height] = SIZE_MAPPINGS[fetch_or_fallback(SIZE_OPTIONS, size, DEFAULT_SIZE)]
       @system_arguments[:viewBox] = "0 0 16 16"
--- a/app/components/primer/spinner_component.rb
+++ b/app/components/primer/spinner_component.rb
@@ -31,6 +31,10 @@ module Primer
       @system_arguments = system_arguments
       @system_arguments[:tag] = :svg
       @system_arguments[:style] ||= DEFAULT_STYLE
+      @system_arguments[:classes] = class_names(
+        @system_arguments[:classes],
+        "anim-rotate"
+      )
       @system_arguments[:width] = SIZE_MAPPINGS[fetch_or_fallback(SIZE_OPTIONS, size, DEFAULT_SIZE)]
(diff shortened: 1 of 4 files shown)


simurai (changes requested):
Thanks for the PR and detailed explanation. 🙇 

Changes look good, but we might can consider the following tweaks:

1. ~~Use the `anim-rotate` class~~. Update: You're too fast. 😄  It already [exists on dotcom](https://github.com/github/github/blob/e874f21773dcffac8180deafa2e2489006bc4b05/app/assets/stylesheets/hacks/hx_animations.scss#L21-L31), so this PR could already ship ahead of https://github.com/primer/css/pull/1251.
2. Does the `transform-origin: 8px 8px;` work for [all sizes](https://primer.style/view-components/components/spinner)? An alternative might be to add the  `anim-rotate` class to the `<svg>` instead of `<path>`. Then `transform-origin` should not be needed. 
 
Another benefit of animating the `<svg>` would be that it can be done with the GPU and should be more performant (doesn't cause repaints). Animating the `<path>` is done with the CPU and needs repainting on each frame. Here a gif using the "Paint flashing" feature in the DevTools:

Animate `<path>` | Animate `<svg>`
--- | ---
spinner-1 | spinner-2
Repainting needed | No repainting needed

simurai (approved):
🙇  Thanks for the updates. Ok, let's get it into `16.0.0` https://github.com/primer/css/pull/1241.

simurai (changes requested):
nvm... wrong repo. The older review is still valid.

theojulienne:
> Another benefit of animating the <svg> would be that it can be done with the GPU and should be more performant (doesn't cause repaints). Animating the <path> is done with the CPU and needs repainting on each frame. Here a gif using the "Paint flashing" feature in the DevTools:

TIL! Thanks for explaining this, that's really interesting. I agree this makes a lot of sense :+1:

simurai (approved):
Ok, I think this looks good from Design Systems. 👍

app/components/primer/spinner_component.rb
```diff
@@ -31,6 +31,10 @@ def initialize(size: DEFAULT_SIZE, **system_arguments)
       @system_arguments = system_arguments
       @system_arguments[:tag] = :svg
       @system_arguments[:style] ||= DEFAULT_STYLE
+      @system_arguments[:classes] = class_names(
+        @system_arguments[:classes],
+        "anim-rotate"
+      )
```
manuelpuyol, line 37:
```suggestion
      @system_arguments[:animation] = :rotate
```

this will add `anim-rotate` for you

Issue #394, opened 2021-03-19
Update Primer::OcticonComponent to accept positional arguments


manuelpuyol:
we could potentially change
```rb
    def initialize(icon:, size: SIZE_DEFAULT, **system_arguments)
      @icon = icon
```

to

```rb
    def initialize(icon_name = nil, icon: nil, size: SIZE_DEFAULT, **system_arguments)
      @icon = icon_name || icon
```

which would avoid breaking changes?

_Originally posted by @manuelpuyol in https://github.com/primer/view_components/pull/393#discussion_r597728216_

PR #399, opened 2021-03-23
Allow passing the icon name as a positional argument to `OcticonComponent`

manuelpuyol:
Closes https://github.com/primer/view_components/issues/394

This will allow:
```rb
Primer::OcticonComponent.new(icon: :x)
Primer::OcticonComponent.new(:x)
primer_octicon(icon: :x)
primer_octicon(:x)
```

### DIFF
--- a/app/components/primer/octicon_component.rb
+++ b/app/components/primer/octicon_component.rb
@@ -18,19 +18,20 @@ module Primer
     SIZE_OPTIONS = SIZE_MAPPINGS.keys
 
     # @example Default
+    #   <%= render(Primer::OcticonComponent.new("check")) %>
     #   <%= render(Primer::OcticonComponent.new(icon: "check")) %>
     #
     # @example Medium
-    #   <%= render(Primer::OcticonComponent.new(icon: "people", size: :medium)) %>
+    #   <%= render(Primer::OcticonComponent.new("people", size: :medium)) %>
     #
     # @example Large
-    #   <%= render(Primer::OcticonComponent.new(icon: "x", size: :large)) %>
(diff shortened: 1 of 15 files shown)


srt32:
The driver here is a less verbosity when using the helper?

manuelpuyol:
> The driver here is less verbosity when using the helper?

yes!

app/components/primer/octicon_component.rb
```diff
@@ -18,19 +18,20 @@ class OcticonComponent < Primer::Component
...
-    #   <%= render(Primer::OcticonComponent.new(icon: "x", size: :large)) %>
+    #   <%= render(Primer::OcticonComponent.new("x", size: :large)) %>
     #
     # @param icon [String] Name of [Octicon](https://primer.style/octicons/) to use.
     # @param size [Symbol] <%= one_of(Primer::OcticonComponent::SIZE_MAPPINGS) %>
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
-    def initialize(icon:, size: SIZE_DEFAULT, **system_arguments)
-      @icon = icon
+    def initialize(icon_name = nil, icon: nil, size: SIZE_DEFAULT, **system_arguments)
+      @icon = icon_name || icon
```
srt32, line 34: Do we plan to keep this support forever or think we'll deprecate the old way at some point?

srt32 (approved):

app/components/primer/octicon_component.rb
manuelpuyol: we can deprecate it at some point

PR #406, opened 2021-03-25
Don't raise when TabbedComponent does not select a tab

manuelpuyol:


Gemfile.lock
```diff
@@ -116,7 +116,8 @@ GEM
     marcel (0.3.3)
       mimemagic (~> 0.3.2)
     method_source (1.0.0)
-    mimemagic (0.3.5)
+    mimemagic (0.3.8)
```
srt32, line 119: Is this change related?

srt32 (approved):
Could we add some explanation to the PR description?

joelhawksley (approved):

Gemfile.lock
manuelpuyol: this was needed for CI to work (and to run test locally 😓 )

Issue #389, opened 2021-03-17
Promote AvatarStack to beta


manuelpuyol:

srt32:
Dup of https://github.com/primer/view_components/issues/346

PR #408, opened 2021-03-25
Promote `AvatarStack` to beta

manuelpuyol:
Closes https://github.com/primer/view_components/issues/389


srt32:
I think we're using this component only once in dotcom so far. Should we use it more before we go to beta?

manuelpuyol:
> I think we're using this component only once in dotcom so far. Should we use it more before we go to beta?

we can wait, no problem

joelhawksley (approved):

Issue #283, opened 2021-02-25
Add new label to docs site for components that require JS behaviors


srt32:
Inspired by https://github.com/primer/view_components/pull/282/. We could do something like `Requires JS`

manuelpuyol:
After https://github.com/primer/view_components/pull/262 is merged, we'll have more control over the doc layout and we can implement this!

manuelpuyol:
@colebemis do you know if there is a way of customizing the layout without rewriting the `Layout` component from doctocat? #262 used to rewrite the component, but since the Storybook link was upstreamed we are using the default layout

PR #409, opened 2021-03-25
Add flash to indicate that a component requires JS

manuelpuyol:
Closes https://github.com/primer/view_components/issues/283

image


Rakefile
```diff
@@ -159,6 +159,14 @@ namespace :docs do
       Primer::UnderlineNavComponent
     ]
 
+    js_components = [
```
srt32, line 162: I wonder if we could make this declaration live on the component itself similar to status. I think that way it's harder to forget about this file ;) WDYT?

srt32 (approved):

Rakefile
manuelpuyol: I think that's a good followup! wanted to keep it simple for now

PR #411, opened 2021-03-26
Update demo mimemagic and improve cuprite setup

manuelpuyol:
`demo` was using the yanked version, which was breaking deploys
our cuprite setup was causing a lot of flaky errors, so I'm upping the timeout to 15s instead of 10 and extracting the setup into a support file

### DIFF
--- a/demo/Gemfile.lock
+++ b/demo/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
-    primer_view_components (0.0.32)
+    primer_view_components (0.0.33)
       octicons_helper (>= 9.0.0, < 13.0.0)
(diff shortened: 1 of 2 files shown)


demo/Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
-    primer_view_components (0.0.32)
+    primer_view_components (0.0.33)
```
srt32, line 4: Should we be bumping this as part of the release process?

srt32 (approved):

demo/Gemfile.lock
manuelpuyol: it failed because of `mimemagic` :/

PR #412, opened 2021-03-29
Output structured arg info

srt32:
Update the docs:build process to output a yaml blob of the arguments and types for each component. We can then programmatically (via http raw requests to dotcom probably) use this data to align our other component projects.

The file lives along the other static statuses.json file.

Since we run the build step on PR's, this file should stay up to date "for free" as we add / change components.


srt32:
cc/ @ashygee @colebemis this sort of export could be used to keep your parameters data fresh in your cross component project work

manuelpuyol (approved):

srt32:
the build docs build did not happen (I think because this is a fork?). We can merge and then confirm on the next PR it behaves nicely.

PR #413, opened 2021-03-29
Organize system arguments into sections

manuelpuyol:
Having everything in a single table makes it hard to find the argument you want. Here I'm separating the System Arguments into sections to make it easier to navigate:
image

### DIFF
--- a/docs/content/system-arguments.md
+++ b/docs/content/system-arguments.md
@@ -15,9 +15,10 @@ To apply different values across responsive breakpoints, pass an array with up t
 For example:
 
 ```erb
-<h1 class="mt-0 mt-lg-4 mt-xl-2">
+<%= render Primer::HeadingComponent.new(mt: [0, nil, nil, 4, 2]) do %>
   Hello world
-</h1>```
+<% end %>
+```
 
 Renders:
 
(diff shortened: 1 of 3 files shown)


srt32 (approved):
Very nice!

joelhawksley:
Thanks for taking this on @manuelpuyol! What do you think about aligning this with https://primer.style/components/system-props?

manuelpuyol:
> Thanks for taking this on @manuelpuyol! What do you think about aligning this with https://primer.style/components/system-props?

I've aligned most of the sections according to that!
I feel like the table layout is not as helpful as the sections though

docs/content/system-arguments.md
```diff
@@ -32,67 +31,112 @@ System arguments include most HTML attributes. For example:
...
+
+| Name | Type | Description |
+| :- | :- | :- |
+| `font_size` | String, Integer | One of `00`, `0`, `1`, `2`, `3`, `4`, `5`, or `6`. |
+| `font_weight` | Symbol | Font weight. One of `:light`, `:normal`, or `:bold`. |
+| `text_align` | Symbol | Text alignment. One of `:left`, `:right`, or `:center`. |
+| `underline` | Boolean | Whether text should be underlined. |
+| `word_break` | Symbol | Whether to break words on line breaks. Can only be `:break_all`. |
 
 ## Arguments
```
joelhawksley, line 136: Might we use a more specific title here? How about: `Tag options`?

joelhawksley (approved):

PR #420, opened 2021-03-30
Accessibility: Don't add tab roles when `UnderlineNav` or `TabNav` use link redirects.

manuelpuyol:
`TabNavComponent` and `UnderlineNavComponent` were using aria roles incorrectly.
There are two cases for those components.

## Navigation links

When we want to add navigation links that change the URL, we don't need to add `role=tablist` or `role=tab`. In this case, the main element is a `nav`.
Selected links will have `aria-current=page`.

### Links as a list

It is also recommended to wrap the links in a list. Now, when using `body_arguments: { tag: :ul }` will update the body to use `<ul>` and each link will be wrapped in a `<li>`

## Tab groups

When the navigation does not change the URL, only toggling the content, we need to follow a set of rules:

1. The element that serves as the container for the set of tabs has role tablist.
Each element that serves as a tab has role tab and is contained within the element with role tablist.
1. Each element that contains the content panel for a tab has role tabpanel.
If the tab list has a visible label, the element with role tablist has aria-labelledby set to a value that refers to the labelling element. Otherwise, the tablist element has a label provided by aria-label.
1. Each element with role tab has the property aria-controls referring to its associated tabpanel element.
1. The active tab element has the state aria-selected set to true and all other tab elements have it set to false.
1. Each element with role tabpanel has the property aria-labelledby referring to its associated tab element.
1. If a tab element has a popup menu, it has the property aria-haspopup set to either menu or true.
1. If the tablist element is vertically oriented, it has the property aria-orientation set to vertical. The default value of aria-orientation for a tablist element is horizontal.

So, when `with_panel` is `true`, we modify:

1. Change `nav` to `div`.
2. Change tabs from `a` to `button`.
3. Set `aria-label` in the element wrapping the `buttons` instead of the `nav` element.
4. Each tab and tabpanel receives a ID based on the component id.
5. Set `aria-controls` and `aria-labelledby` according to the generated id.
6. Selected tabs will have `aria-selected=true`.

## Reference

https://www.w3.org/TR/wai-aria-practices-1.2/#wai-aria-roles-states-and-properties-21
https://www.w3.org/TR/wai-aria-practices-1.2/examples/tabs/tabs-2/tabs.html


app/components/primer/underline_nav_component.rb
```diff
@@ -100,31 +117,38 @@ class UnderlineNavComponent < Primer::Component
...
+        @body_arguments[:classes],
+        "list-style-none" => list?
+      )
+      @body_arguments[:role] = :tablist if with_panel
     end
 
     private
 
+    def list?
+      @body_tag == :ul
```
joelhawksley, line 149: What do you think about using `fetch_or_fallback` to define @body_tag? That way the value will be more predictable here.

app/components/primer/underline_nav_component.rb
```diff
@@ -100,31 +117,38 @@ class UnderlineNavComponent < Primer::Component
     #
     # @param with_panel [Boolean] Whether the TabNav should navigate through pages or panels.
     # @param align [Symbol] <%= one_of(Primer::UnderlineNavComponent::ALIGN_OPTIONS) %> - Defaults to <%= Primer::UnderlineNavComponent::ALIGN_DEFAULT %>
+    # @param body_arguments [Hash] <%= link_to_system_arguments_docs %> for the body wrapper.
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
-    def initialize(with_panel: false, align: ALIGN_DEFAULT, body_classes: "", **system_arguments)
+    def initialize(with_panel: false, align: ALIGN_DEFAULT, body_arguments: { tag: BODY_TAG_DEFAULT }, **system_arguments)
```
joelhawksley, line 122: 🤔 interesting. I wonder if this API could prove tricky to use. For example, if someone passes in `body_arguments`, they need to be sure to pass in a `tag` key. What do you think about having `body_classes` and `body_tag`? Then again, this is starting to feel like a slot...

koddsson (approved):
This looks good to me. Me and @manuelpuyol pair-reviewed this and talked about dropping the `aria-control` for simplicity perceived "uselessness". `aria-control` is not used by `github/tab-container-element` so I feel pretty confident in dropping that attribute.

## References
https://heydonworks.com/article/aria-controls-is-poop/

PR #432, opened 2021-04-08
Ensure State component is not a link

### LINKED ISSUE #418: Ensure primer state component is not a link.
~~Let's add an example to https://primer.style/view-components/components/state that generates a link.~~

Let's not allow state components to be links.

khiga8:
Closes #418 

**What**
This makes it so a State component cannot be rendered as a link.

**Why**
The purpose of `State` is to render the status of an item. It seems like `State` is being in some places in dotcom for the rounded aesthetic, but there are other components that are probably a better fit for rendering a link.

On our end, we can enforce correct future usage of the `StateComponent` by not allowing it to be a link. 

**Notes**
[Related PR][1]

[1]: https://github.com/github/github/pull/176322

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* **Breaking change**: Don't allow `StateComponent` to be a link.
+
+    *Kate Higa*
+
 ## 0.0.37
 
 * Update NPM package to include subdirectory JS files.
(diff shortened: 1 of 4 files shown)


koddsson (approved):
This looks good to me ✨

manuelpuyol:
@khiga8 mind adding a CHANGELOG entry?

manuelpuyol (approved):

Issue #443, opened 2021-04-13
Add more `Button` schemes


manuelpuyol:
We are missing:
- [ ] [Block](https://primer.style/css/components/buttons#block-button)
- [ ] [Link](https://primer.style/css/components/buttons#link-button)
- [ ] [HiddenText](https://primer.style/css/components/buttons#hidden-text-button)
- [ ] [Icon only](https://primer.style/css/components/buttons#icon-only-button)
- [ ] [Close](https://primer.style/css/components/buttons#close-button)
- [ ] [With counts](https://primer.style/css/components/buttons#button-with-counts)

PR #444, opened 2021-04-13
Add `link`, `block` and `invisible` schemes to Button

manuelpuyol:
Part of https://github.com/primer/view_components/issues/443

Adds some missing schemes to button.

image


app/components/primer/button_component.rb
```diff
@@ -6,11 +6,15 @@ class ButtonComponent < Primer::Component
...
 
     DEFAULT_SCHEME = :default
+    LINK_SCHEME = :link
     SCHEME_MAPPINGS = {
       DEFAULT_SCHEME => "",
       :primary => "btn-primary",
       :danger => "btn-danger",
-      :outline => "btn-outline"
+      :outline => "btn-outline",
+      :block => "btn-block",
```
joelhawksley, line 15: Is it possible to have a button that is both `primary` and `block`? What about `outline` and `primary`?

manuelpuyol: oh that's true, `block` is just a modifier, going to move it to a boolean
`outline` though is a scheme and shouldn't be used with another schemes

joelhawksley (approved):

PR #445, opened 2021-04-13
Create `CloseButton` component

manuelpuyol:
Part of https://github.com/primer/view_components/issues/443.

This PR creates `CloseButton` component base on https://primer.style/css/components/buttons#close-button
Since this button is very specific, I decided to make it a separate component from `Button` instead of a `scheme`.

image


app/components/primer/close_button.rb
```diff
@@ -0,0 +1,30 @@
...
+      @system_arguments[:tag] = :button
+      @system_arguments[:type] = fetch_or_fallback(TYPE_OPTIONS, type, DEFAULT_TYPE)
+      @system_arguments[:classes] = class_names(
+        "close-button",
+        system_arguments[:classes]
+      )
+    end
+
+    def call
+      render(Primer::BaseComponent.new(**@system_arguments)) do
```
joelhawksley, line 25: This would be a nice place to use `with_content` 😄

docs/content/components/closebutton.md
```diff
@@ -0,0 +1,29 @@
...
+status: Alpha
+source: https://github.com/primer/view_components/tree/main/app/components/primer/close_button.rb
+storybook: https://primer.style/view-components/stories/?path=/story/primer-close-button-component
+---
+
+import Example from '../../src/@primer/gatsby-theme-doctocat/components/example'
+
+<!-- Warning: AUTO-GENERATED file, do not edit. Add code comments to your Ruby instead <3 -->
+
+Use close button to render a `×` without default button styles.
```
joelhawksley:
```suggestion
Use CloseButton to render a `×` without the default button styles.
```

app/components/primer/close_button.rb
manuelpuyol: soon!

joelhawksley (approved):

manuelpuyol:
@joelhawksley I'm thinking about the way it is rendered..

right now it is `render(Primer::BaseComponent.new(**@system_arguments))` but I wonder if it would be a good idea to have something like `render(Primer::Buttonmponent.new(unstyled: true, classes: "close-button", type: @type)` where `unstyled` makes `ButtonComponent` not add the `btn` class.

The reasoning here would be that **all** `<button>` tags should be rendered via `ButtonComponent`, but I'm not sure if that's true

joelhawksley:
@manuelpuyol I could see that, as we might want to have some underlying behavior shared by all buttons 🤔

manuelpuyol:
> @manuelpuyol I could see that, as we might want to have some underlying behavior shared by all buttons 🤔

Yeah that's what I thought too. I'm going to leave this as-is for now and then work on a PR with those changes then (to avoid conflicts with https://github.com/primer/view_components/pull/444 😬 )

PR #447, opened 2021-04-13
Extract `Primer::Button::Base`

manuelpuyol:
Initial discussion: https://github.com/primer/view_components/pull/445#issuecomment-818828571

# Summary

We always want to use a component when rendering a Button since we can share some behaviors (like not adding `type='button'` when using a `summary` tag). Our current button's problem is that it always adds the `btn` style, which is needed in most cases, but not all.

We have some buttons like `Button::Close`, `Button::Link`, `Button::Marketing` and `HiddenTextExpander` that don't need all of our `ButtonComponent` styles, but they do need the shared behaviors.

# Solution

I decided to extract a `Button::Base`, which is basically an unstyled button. It holds the logic for possible `types` and `tags` and turning buttons into `blocks`. In the future, we could add some accessibility features too!
The new component should be used any time you need a button without the need for `btn`, `schemes`, and `variants`.

`Button::Base` also paves a path for a better structure (and also approaches us to the React implementation). So I also decided to move:

`CloseButton` -> `Button::Close`
`ButtonMarketing` -> `Button::Marketing`
`ButtonComponent.new(scheme: :link)` -> `Button::Link`


manuelpuyol:
Close in favor of https://github.com/primer/view_components/pull/473

Issue #449, opened 2021-04-14
Add a component generator


manuelpuyol:
# Problem

It is very hard for someone to onboard on this project. When creating a component we have to add it to multiple places that aren't obvious.

# Proposal

Based on this checklist:

```
- [ ] Create rb file
- [ ] Create template file
- [ ] Add npm packages
- [ ] Create storybook
- [ ] Add tests
- [ ] Add docs
- [ ] Add component to Rakefile
- [ ] Add component to nav.yml
- [ ] Add component to `component_test`
```

We should create a generator that does all of the above for you. We can use flags to skip/add steps.

PR #450, opened 2021-04-14
Add component generator

manuelpuyol:
Closes https://github.com/primer/view_components/issues/449

This PR creates a component generator using [Thor](https://github.com/erikhuda/thor). I'm not using rails generators here since we don't have access to rails in the gem.
The generator will:

1. Create a `rb` file
2. Create a `html.erb` file
3. Create a failing test
4. Create a story file
5. Add the component to the `Rakefile` list
6. Add the component to our `component_test`
7. Add the component to `nav.yml`

If the `js` flag is used, it will also:

1. Create a `ts` file importing the package
2. Import the component `ts` in `primer.ts`
3. Add the component to `js_components` list in `Rakefile`
4. Install the package

---

Things to note:
1. The generator can't create components inside modules, only directly in `Primer::`
2. You will need to fix the `nav.yml` order manually after running the generator


CONTRIBUTING.md
```diff
@@ -8,6 +8,20 @@ Contributions to this project are [released](https://help.github.com/articles/gi
...
 
+## Creating a component
+
+You can use a generator to create a component and add it to all the necessary files.
+
+```sh
+bundle exec thor component_generator my_component_name
+```
+
+If your component requires a JS package, you can pass the `js` flag to the generator.
```
joelhawksley, line 19:
```suggestion
To declare a dependency on an `npm` package, pass `js` to the generator:
```

joelhawksley (approved):

PR #452, opened 2021-04-15
Create DetailsMenu component.

khiga8:
**What**
This PR introduces a new `DetailsMenu` component which renders a custom `<details-menu>`element ported from [this repo][1]. This is **NOT** a standalone component and should only be used as a menu within a `<details>` element. 

[1]: https://github.com/github/details-menu-element

**Why**
`<details-menu>` provides enhanced keyboard navigation and javascript options. We use it frequently throughout dotcom in conjunction with classes `select-menu-modal`, `SelectMenu`, `dropdown-menu` . Since this is a widely-used element and is a building block for other components (including `Dropdown` which we already have), it makes sense to encapsulate this with a view component.

**Notes**
- I found [these slides][3] helpful to gain more context on `details-menu-element`.

[3]: https://docs.google.com/presentation/d/1hvnPpsJo44BTPfJx28CV95vqk_dt6na1awUbk0kmZYM/edit#slide=id.g3e31444916_1_26

Co-authored-by: Manuel Puyol <manuelpuyol@github.com>


khiga8:
Still a work in progress but opened a draft to maybe get a lookover and thoughts before I proceed. Let me know if these new changes seem reasonable or if you have other insights/ideas!

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@
 
     *Manuel Puyol*
 
+* Create `DetailsMenu` component.
+    *Kate Higa*
```
manuelpuyol, line 22:
```suggestion
* Create `DetailsMenu` component.

    *Kate Higa*
```

app/components/primer/details_menu_component.rb
```diff
@@ -0,0 +1,25 @@
...
+
+module Primer
+  # This component is used by [DetailsComponent][1] and [Dropdown::MenuComponent][1],
+  # and should not be used as a standalone component.
+  #
+  # This can be used to provide `<details>` with a menu that has enhanced keyboard navigations as well as
+  # configurable [javascript][2] behaviors. For further guidance on usage, please see [details-menu-element][1].
+  #
+  # [1]: https://primer.style/view-components/components/details
+  # [2]: https://github.com/github/details-menu-element
```
manuelpuyol, line 11: I think there is some confusion here with the links?

test/components/details_menu_component_test.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class PrimerDetailsMenuComponentTest < Minitest::Test
```
manuelpuyol, line 5: may we also add a test to check if the content is rendered?
I'd also like to add a test checking that the HTML tag can't be modified

manuelpuyol (commented):
I wonder why `app/assets/javascripts/primer_view_components.js.map` was changed but `app/assets/javascripts/primer_view_components.js` wasn't

khiga8:
hey @manuelpuyol, thanks for reviewing!

I realized something and would love your thoughts on this... Taking a step back, I noticed that our [current](https://primer.style/view-components/components/dropdown#default) Dropdown component renders with an extra div:

```
<details classname='dropdown'>
  <summary>
  <div>
    <details-menu classname='dropdown-menu'>
  </div>
</details>
```

which we probably don’t want because this extra div interferes with expected javascript behavior in [details-menu-element](https://github.com/github/details-menu-element) like we found the other day. This extra div comes from our [lambda](https://github.com/primer/view_components/blob/main/app/components/primer/details_component.rb#L31) body slot in the Details component. Looks like this allows users to flexibly set the tag of the body and do something like:
```.erb
<%= render(Primer::DetailsComponent.new(**@system_arguments)) do |c| %>
  <% c.summary %>
  <% c.body(tag: "details-dialog") do %>
    ...
  <% end %>
<% end %>
```
The default tag is `div`. 
However, with `Dropdown` component which is currently rendering `Details` component, we have our own custom `Dropdown::Menu` we want to pass through instead of rendering what's in lambda. From my understanding, we can't conditionally do lambda slot or pass through slot.

I’m thinking we have these options:
* Update `Dropdown` to not use `Details` and have own implementation. 
* Simplify `Details` to only support pass through body slots and remove lambda rendering (breaking change in dotcom)

I think I'm leaning towards one of these... but wanted to hear other ideas or thoughts you might have.  

Aside from our existing `Dropdown` component, it seems very likely we'll have more components that are variations of `Details` with potentially its own custom menu component so that would be something to keep in mind.

Wanted to write this out, but also happy to chat in person!

manuelpuyol:
> hey @manuelpuyol, thanks for reviewing!
> 
> I realized something and would love your thoughts on this... Taking a step back, I noticed that our [current](https://primer.style/view-components/components/dropdown#default) Dropdown component renders with an extra div:
> 
> ```
> <details classname='dropdown'>
>   <summary>
>   <div>
>     <details-menu classname='dropdown-menu'>
>   </div>
> </details>
> ```
> 
> which we probably don’t want because this extra div interferes with expected javascript behavior in [details-menu-element](https://github.com/github/details-menu-element) like we found the other day. This extra div comes from our [lambda](https://github.com/primer/view_components/blob/main/app/components/primer/details_component.rb#L31) body slot in the Details component. Looks like this allows users to flexibly set the tag of the body and do something like:
> 
> ```
> <%= render(Primer::DetailsComponent.new(**@system_arguments)) do |c| %>
>   <% c.summary %>
>   <% c.body(tag: "details-dialog") do %>
>     ...
>   <% end %>
> <% end %>
> ```
> 
> The default tag is `div`.
> However, with `Dropdown` component which is currently rendering `Details` component, we have our own custom `Dropdown::Menu` we want to pass through instead of rendering what's in lambda. From my understanding, we can't conditionally do lambda slot or pass through slot.
> 
> I’m thinking we have these options:
> 
> * Update `Dropdown` to not use `Details` and have own implementation.
> * Simplify `Details` to only support pass through body slots and remove lambda rendering (breaking change in dotcom)
> 
> I think I'm leaning towards one of these... but wanted to hear other ideas or thoughts you might have.
> 
> Aside from our existing `Dropdown` component, it seems very likely we'll have more components that are variations of `Details` with potentially its own custom menu component so that would be something to keep in mind.
> 
> Wanted to write this out, but also happy to chat in person!

huh, that's really interesting! We had a similar problem when we tried to test multiple components for buttons instead of a single one.

At that time we though that we _could_ do something like

```rb
renders_one :body, lambda { |klass: Primer::BaseComponent, **system_arguments|
  klass.new(**system_arguments)
}
```

then, for DetailsMenu it could be something like

```erb
render Primer::Details.new do |c|
  c.body(klass: Primer::DetailsMenu) do
    some content
  end
end
```

and for Dropdown:

```erb
render Primer::Details.new do |c|
  c.body(klass: Primer::Dropdown::MenuComponent) do
    some content
  end
end
```

we never implemented that solution, but it may be the useful here?
cc @joelhawksley

joelhawksley:
@khiga8 from what I can tell, I think we should add alternative `body` slots for each of the possible body implementations, instead of allowing any `Klass` to be passed in. That way, we can ensure control over how the component is used.

khiga8:
Closing for now to ponder on this....

Issue #439, opened 2021-04-09
Create `ClipboardCopyComponent`


manuelpuyol:
https://github.com/github/clipboard-copy-element

PR #457, opened 2021-04-19
Add ClipboardCopy component

koddsson:
This PR introduces the ClipboardCopy behavioral component. The functionality for this component is provided by the [`<clipboard-copy>` Web Component](https://github.com/github/clipboard-copy-element).

Backed by JavaScript, behavioral components bring interactivity to Primer ViewComponents. Behavioral components are in contrast to the "presentational" components that enforce layout and styling with CSS. Components can provide both interactivity and styling, but `ClipboardCopy` is strictly behavioral.

This is my first pull request to this project, so I'm open to all kinds of feedback. I'm mostly unsure of the inability to pass another ViewComponent into the target slot. You can see from the tests and the Storybook instance that I pass a `:a` attribute to the slot, but it would be much better if you could pass `Primer::LinkComponent` into the slot. Not sure if something like this is even feasible.

Closes https://github.com/primer/view_components/issues/439


stories/primer/clipboard_copy_stories.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class Primer::ClipboardCopyStories < ViewComponent::Storybook::Stories
+  layout "storybook_preview"
+
+  story(:clipboard_copy_simple) do
+    controls do
+      text(:id, "id")
+      text(:value, "Text to copy")
+      text(:classes, "btn btn-sm")
```
joelhawksley, line 10:
```suggestion
      text(:value, "Text to copy")
```

For the sake of these examples, I don't think it's necessary to provide controls for `id` or `classes`.

app/components/primer/clipboard_copy.rb
```diff
@@ -0,0 +1,57 @@
...
+    #   <% end %>
+    #
+    # @example With a link target
+    #   <%= render(Primer::ClipboardCopy.new(id: "foo")) do |component|
+    #     <%= component.target(tag: :a, href: "/path/to/copy") %>
+    #
+    #     Click to copy!
+    #   <% end %>
+    #
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
```
joelhawksley, line 46: Might we add a `@param` for `value`, as we have an example above showing its use?

app/components/primer/clipboard_copy.rb
```diff
@@ -0,0 +1,57 @@
...
+      Primer::BaseComponent.new(**system_arguments)
+    }
+
+    # @example Default
+    #   <%= render(Primer::ClipboardCopy.new(value: "Text to copy")) do %>
+    #     "Click to copy!"
+    #   <% end %>
+    #
+    # @example With a target
+    #   <%= render(Primer::ClipboardCopy.new(id: "foo")) do |component|
```
joelhawksley, line 24: Do we need to have the consumer provide an `id` at all, or could we generate one randomly for them? (Or maybe both?

app/components/primer/clipboard_copy.rb
```diff
@@ -0,0 +1,57 @@
...
+
+module Primer
+  # Use ClipboardCopy to populate copy values from the DOM
+  class ClipboardCopy < Primer::Component
+    status :alpha
+
+    # Optional target element that holds the data the user will copy.
+    #
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    renders_one :target, lambda { |**system_arguments|
```
joelhawksley, line 11: Looking around dotcom, I think it might make sense to also have an `icon` slot that wraps an OcticonComponent, as we often use an octicon with this component. 

This can of course wait for another PR if you'd prefer ❤️

joelhawksley: Taking this one step further, we almost _always_ use this with the `clippy` icon, but inconsistently swap it for the `check` icon when clicked. Perhaps we could encode that mode as an option on the component?

joelhawksley (commented):
Looking good! I think we can ship this more or less as-is and iterate as we implement it in dotcom.

app/components/primer/clipboard_copy.rb
koddsson: I thought about randomly generating an ID. I didn't want to complicate too much, but it's certainly possible for us to do that. We want to support the user providing an ID to link any DOM node they want to the clipboard copy element.

I'll add some code to generate IDs automatically when not provided by the user.

app/components/primer/clipboard_copy.rb
koddsson: I think including the icon here makes sense. I think defaulting to the icon makes sense, and we can let users pass `icon: nil` to render the component without the icon.

koddsson:
@joelhawksley, I decided to vastly simplify the component since, as you say, we primarily use it in the simple icon configuration. Simplifying it will make it easier to ship, and we avoid some problems with accessibility. We can start integrating the component and add functionality to it as it's needed.

test/components/component_test.rb
```diff
@@ -21,6 +21,7 @@ class PrimerComponentTest < Minitest::Test
     [Primer::ButtonComponent, {}],
     [Primer::ButtonGroupComponent, {}, proc { |component| component.button { "Button" } }],
     [Primer::ButtonMarketingComponent, {}],
+    [Primer::ClipboardCopy, { label: "Sting that will be read to screenreaders", value: "String that will be copied" }],
```
joelhawksley, line 24:
```suggestion
    [Primer::ClipboardCopy, { label: "String that will be read to screenreaders", value: "String that will be copied" }],
```

app/components/primer/clipboard_copy.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Primer
```
joelhawksley, line 3: Might you be able to add this component to the docs per https://github.com/primer/view_components/blob/main/CONTRIBUTING.md#submitting-a-pull-request?

joelhawksley (approved):

koddsson:
The test failures feel a bit flakey. The tests consistently run green on my machine, not sure what the difference is between my machine and CI 🤔 

image

PR #459, opened 2021-04-19
Read `TESTS` environmental variable for tests to run

koddsson:
This PR adds the capability for developers to pass in a `TESTS` environmental variable when running the `test` rake task. The rake task will populate the tests to run after the glob that is passed into `TESTS` and subsequentially only run the tests that match that glob. If no `TESTS` environmental variable is set, the rake task will default to `test/**/*_test.rb` as a glob.

This change will make it easier for developers to run only specific tests and not bother running tests unrelated to their debugging or development.


joelhawksley (approved):
Nice! Mind adding some docs on how to use this?

PR #464, opened 2021-04-19
Add aria-label to `CloseButton`

### LINKED ISSUE #460: CloseButton component is missing accessible label
[ARIA14: Using aria-label to provide an invisible label where a visible label cannot be used](https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html#:~:text=The%20'close'%20element%20is%20implemented,accessible%20name%20to%20the%20button.)

The `CloseButton` component is missing an accessible label so the screen reader will simply announce "button". If we are not providing an aria-label ourselves, we should update the documentation to recommend use of label.

khiga8:
Closes #460 

**What**
This PR adds a missing `aria-label` to the `CloseButton` component along with relevant documentation

**Notes**
- It seems beneficial to have a section in the documentation specifically around `Accessibility` (and maybe eventually even sections like `When to use`, `When not to use`) to guide developers to correct component usage. I think it's worth adding a custom `@accessibility` tag. I explored modifying `Rake` but couldn't quite figure out how to add a custom tag to output associated comments. If reviewer is yard savvy and has ideas on how this is done, please let me know! :-)

### DIFF
--- a/demo/.storybook/main.js
+++ b/demo/.storybook/main.js
@@ -1,8 +1,8 @@
 module.exports = {
   stories: ['../../**/*.stories.json'],
   addons: [
-    '@storybook/addon-controls',
-    '@storybook/addon-a11y'
+    '@storybook/addon-a11y',
+    '@storybook/addon-controls'
   ],
   webpackFinal: async (config, { configType }) => {
     if(configType == 'PRODUCTION')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
     *Kate Higa*
 
+* Add default `aria-label` of "Close" to `CloseButton` component.
+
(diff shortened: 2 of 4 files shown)


demo/.storybook/main.js
```diff
@@ -1,8 +1,8 @@
 module.exports = {
   stories: ['../../**/*.stories.json'],
   addons: [
-    '@storybook/addon-controls',
-    '@storybook/addon-a11y'
+    '@storybook/addon-a11y',
+    '@storybook/addon-controls'
```
khiga8, line 5: Not directly related to this PR, but thought it minor enough to slip in!
Switching the order of these per warning that appears on dev boot.

joelhawksley (approved):
👏🏻

PR #468, opened 2021-04-22
Introduce custom `@accessibility` tag for docs

### LINKED ISSUE #467: Idea: Add additional sections in docs around usage guidelines
Proposal: Modify `Rake` to allow additional docs sections aside from `Examples`, `Slots`. Update `templates/component.tt` with additional sections.

Potential sections: 
- Description of component and when to use. 
- Usage/Best practices (do this, don't do that, offer alternative components)
- Accessibility-related information to be aware of when using the component.

Reasons: 
1. Consumers of `primer/view_components` will likely rely heavily on the docs to make informed UI decisions. We should empower consumers to make informed decisions as easily as possible by surfacing relevant information through docs. Though we can have systemic code checks in place to prevent some bad decisions from being made, there are decisions that consumers may have to make for their specific context.
1. Having to document usage information also allows us to develop primer view components with relevant code checks in place. (e.g. putting in restrictions around allowed tags)

Side note:
I don't quite understand how yard works and if it's possible to add things like custom tags for new sections so got stuck in approaching this. If you are familiar, let's pair!

khiga8:
Closes #467 

**What**
This PR introduces a new `@accessibility` tag to allow an `Accessibility` section to be rendered in yard. 

**Why**
This idea came about as one step towards improving documentation and surfacing best practices to consumers of `primer/view_components`. Though this introduces one new `@accessibility` tag, we may find the need to introduce other tags like `@best_practices`, etc. I think this `Accessibility` section might belong under a broader `Usage` section... but I think we can re-evaluate needs as we extend documentation.


koddsson (approved):
Looks good to me ✨

PR #469, opened 2021-04-22
Add ClipboardCopy to site sidebar

koddsson:
We can merge this after we tag a new version. Since `ClipboardCopy` requires JS, we need to reference the newly built JS package that includes `<clipboard-copy>`.


koddsson:
> We can merge this after we tag a new version. Since `ClipboardCopy` requires JS, we need to reference the newly built JS package that includes `<clipboard-copy>`.

@manuelpuyol We need to bump the Primer ViewComponent version since https://primer.style/view-components/components/clipboardcopy references the 0.37 npm package, which doesn't include the JS behaviors for `<clipboard-copy>`.

Issue #470, opened 2021-04-22
Don't allow mixed `variants` on ButtonGroup


manuelpuyol:
With the current `ButtonGroupComponent`, we can have:

```rb
render(Primer::ButtonGroupComponent.new) do |c|
  c.button(variant: :small) { "Small" }
  c.button { "Default" }
  c.button(variant: :large) { "Large" }
end
```

Mixed variants shouldn't be allowed, so we could change the `ButtonGroup` API to receive `variant` and pass it to the button slots

PR #471, opened 2021-04-22
Set Button variant on ButtonGroup

manuelpuyol:
Closes https://github.com/primer/view_components/issues/470

This PR updates the `ButtonGroupComponent` API to set the variant of **all** the buttons in the group, not allowing different variants in the same group


app/components/primer/button_group_component.rb
```diff
@@ -6,19 +6,43 @@ class ButtonGroupComponent < Primer::Component
     # Required list of buttons to be rendered.
     #
     # @param kwargs [Hash] The same arguments as <%= link_to_component(Primer::ButtonComponent) %>.
```
joelhawksley, line 8: The same, except for `variant`?

static/arguments.yml
```diff
@@ -219,6 +219,10 @@
 - component: ButtonGroup
   source: https://github.com/primer/view_components/tree/main/app/components/primer/button_group_component.rb
   parameters:
+  - name: variant
+    type: Symbol
+    default: "`Primer::ButtonComponent::DEFAULT_VARIANT`"
```
joelhawksley, line 224: I _think_ we evaluate constants to their values in other docs, perhaps we could do the same here? That way this would read `:medium`

joelhawksley (approved):
👏🏻

PR #472, opened 2021-04-22
don't set default tag for Heading component and improve docs

### LINKED ISSUE #466: Idea: don't set default tag for `Heading` component, restrict allowed tags, add more examples
Headings elements carry semantic meaning in conveying page hierarchy but it is often used mistakenly to convey stylistic meaning. Incorrect usage of headings can be confusing for screen reader users and detrimental to accessibility.

Proposed changes:
1.  Don't have any default tag and require the consumer to always pass it in.
- Currently the tag defaults to `:h1`. However, the heading level is largely dependent on the page/context it is to be used so it may not make sense to have a default. By not offering a fallback, we encourage the consumer to be intentional about heading levels.

2. Restrict the possible tags to `<h1>` ~ `<h6>` tags for this component
- Currently there are no restrictions. 

3. Add examples that dissociate heading levels from aesthetics.
- Often times, people using heading levels for stylistic purposes (e.g. they want a large text and so they use `<h1>`), but this should be discouraged. We should have examples in the documentation that disassociate heading levels from aesthetics (maybe examples using `font_size:`). We can also redirect the user to use the `Text` component with [relevant typography arguments](https://primer.style/view-components/system-arguments#typography) as an alternative.

khiga8:
Closes #466 

**What**

1.  Don't have any default tag and require the consumer to always pass it in. (We will still provide a fallback tag of `:h2` for prod)
2.  Restrict the possible tags to `<h1>` ~ `<h6>` tags for this component
3. Add examples that dissociate heading levels from aesthetics.

**Why**

1. Heading tag level is largely dependent on context so it may not make sense to have a default. 
2. There is currently no limitation on tag that can be set. We should limit ability of consumer to make incorrect decisions.
3. People often use heading levels for stylistic purposes (e.g. they want a large text and so they use `<h1>`), but this should be discouraged. We should have examples in the documentation that disassociate heading levels from aesthetics (maybe examples using `font_size:`).

### DIFF
--- a/app/components/primer/heading_component.rb
+++ b/app/components/primer/heading_component.rb
@@ -1,19 +1,47 @@
 # frozen_string_literal: true
 
 module Primer
-  # Use the Heading component to wrap a component that will create a heading element
+  # Heading can be used to communicate page organization and hierarchy.
+  #
+  # - Set tag to one of `:h1`, `:h2`, `:h3`, `:h4`, `:h5`, `:h6` based on what is
+  #   appropriate for the page context.
+  # - Use Heading as the title of a section or sub section.
+  # - Do not use Heading for styling alone. To style text without conveying heading semantics,
+  #   consider using <%= link_to_component(Primer::TextComponent) %> with relevant <%= link_to_typography_docs %>.
+  # - Do not jump heading levels. For instance, do not follow a `<h1>` with an `<h3>`. Heading levels should
+  #   increase by one in ascending order.
+  #
+  # @accessibility
+  #   Headings convey semantic meaning. Assistive technology users rely on headings to quickly navigate and scan information on a page.
+  #   Inappropriate use of headings can lead to a confusing experience.
+  #   [Learn more about best heading practices (WAI Headings)](https://www.w3.org/WAI/tutorials/page-structure/headings/)
   class HeadingComponent < Primer::Component
     status :beta
 
+    TAG_FALLBACK = :h2
+    TAG_OPTIONS = [:h1, TAG_FALLBACK, :h3, :h4, :h5, :h6].freeze
+
     # @example Default
-    #   <%= render(Primer::HeadingComponent.new) { "H1 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h1)) { "H1 Text" } %>
     #   <%= render(Primer::HeadingComponent.new(tag: :h2)) { "H2 Text" } %>
     #   <%= render(Primer::HeadingComponent.new(tag: :h3)) { "H3 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h4)) { "H4 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h5)) { "H5 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h6)) { "H6 Text" } %>
     #
+    # @example With `font_size`
+    #   <%= render(Primer::HeadingComponent.new(tag: :h1, font_size: 6)) { "h1 tag, font_size 6" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h2, font_size: 3)) { "h2 tag, font_size 3" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h3, font_size: 2)) { "h3 tag, font_size 2" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h4, font_size: 0)) { "h4 tag, font_size 0" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h5, font_size: 1)) { "h5 tag, font_size 1" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h6, font_size: 4)) { "h6 tag, font_size 4" } %>
+    #
+    # @param tag [String]  <%= one_of(Primer::HeadingComponent::TAG_OPTIONS) %>
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
-    def initialize(**system_arguments)
+    def initialize(tag:, **system_arguments)
       @system_arguments = system_arguments
-      @system_arguments[:tag] ||= :h1
+      @system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, TAG_FALLBACK)
     end
 
     def call
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* **Breaking change**: Do not provide default for `Heading` and improve documentation.
+
+    *Kate Higa*
+
 * Extract `BaseButton` component.
 
     *Manuel Puyol*
@@ -11,8 +15,6 @@
     *Kate Higa*
 
 * **Breaking change**: Don't allow `StateComponent` to be a link.
-
-    *Kate Higa*
 
 ## 0.0.37
 
--- a/Rakefile
+++ b/Rakefile
@@ -91,6 +91,10 @@ namespace :docs do
     "[System arguments](/system-arguments)"
   end
 
+  def link_to_typography_docs
+    "[Typography](/system-arguments#typography)"
+  end
+
   def link_to_component(component)
     short_name = component.name.gsub(/Primer|::|Component/, "")
     "[#{short_name}](/components/#{short_name.downcase})"
--- a/docs/content/components/heading.md
+++ b/docs/content/components/heading.md
@@ -9,22 +9,53 @@ import Example from '../../src/@primer/gatsby-theme-doctocat/components/example'
 
 <!-- Warning: AUTO-GENERATED file, do not edit. Add code comments to your Ruby instead <3 -->
 
-Use the Heading component to wrap a component that will create a heading element
+Heading can be used to communicate page organization and hierarchy.
+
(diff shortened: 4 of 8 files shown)


app/components/primer/heading_component.rb
```diff
@@ -1,23 +1,46 @@
...
+  # - Use one `<h1>`` per page. 
+  # ### Do not
+  # - Skip heading levels. For instance, do not jump from `<h1>` to `<h3>`.
+  # - Use Heading for sole stylistic purposes. For styling text, consider using [Text] with relevant
+  # [typography arguments].
+  #
+  # [Text]: https://primer.style/view-components/components/text
+  # [typography arguments]: https://primer.style/view-components/system-arguments#typography
+  #
+  # ## Accessibility
```
joelhawksley, line 18: Can we use the new `accessibility` tag here?

app/components/primer/heading_component.rb
```diff
@@ -1,23 +1,46 @@
 # frozen_string_literal: true
 
 module Primer
-  # Use the Heading component to wrap a component that will create a heading element
+  # Use Heading to communicate page organization and hierarchy.
+  #
+  # ## Usage
```
joelhawksley, line 6: I'd be 👍🏻 to adding a `usage` tag too 😄

app/components/primer/heading_component.rb
```diff
@@ -1,23 +1,46 @@
...
+  #
+  # - Use as the title of a section or sub section. 
+  # - Use one `<h1>`` per page. 
+  # ### Do not
+  # - Skip heading levels. For instance, do not jump from `<h1>` to `<h3>`.
+  # - Use Heading for sole stylistic purposes. For styling text, consider using [Text] with relevant
+  # [typography arguments].
+  #
+  # [Text]: https://primer.style/view-components/components/text
+  # [typography arguments]: https://primer.style/view-components/system-arguments#typography
```
joelhawksley, line 16: Since we run YARD comments through ERB, we could add helper methods to the Rakefile that return these URLs and call them here, in case we want to change the URLs across the library in the future 😄

app/components/primer/heading_component.rb
```diff
@@ -1,23 +1,46 @@
...
 
+    TAG_OPTIONS = [:h1, :h2, :h3, :h4, :h5, :h6].freeze
+
     # @example Default
-    #   <%= render(Primer::HeadingComponent.new) { "H1 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h1)) { "H1 Text" } %>
     #   <%= render(Primer::HeadingComponent.new(tag: :h2)) { "H2 Text" } %>
     #   <%= render(Primer::HeadingComponent.new(tag: :h3)) { "H3 Text" } %>
     #
+    # @param tag [String]  <%= one_of(Primer::HeadingComponent::TAG_OPTIONS) %>
```
joelhawksley, line 31:
```suggestion
    # @param tag [String] <%= one_of(Primer::HeadingComponent::TAG_OPTIONS) %>
```

app/components/primer/heading_component.rb
```diff
@@ -1,23 +1,46 @@
...
-      @system_arguments[:tag] ||= :h1
+      @system_arguments[:tag] = tag
     end
 
     def call
       render(Primer::BaseComponent.new(**@system_arguments)) { content }
     end
+
+    def before_render
+      raise ArgumentError, "Tag must be set to one of #{TAG_OPTIONS.join(',')}" unless TAG_OPTIONS.include?(@system_arguments[:tag])
```
joelhawksley, line 43: I'd love to see us abstract this into something similar to `fetch_or_fallback`, what do you think?

joelhawksley (commented):
😍 I'm loving the accessibility and usage guidelines!

app/components/primer/heading_component.rb
khiga8: I noticed that method and some unless Rails.env.production?` checks and wanted to get some insight. Why do we need to provide a graceful fallback if the env is production in some cases?

joelhawksley: 🤔 good question. I'm not sure how we'd handle this case here, but the reason we fallback in production is to prevent component misuse from leading to 500 errors. Do you think we should 500 for passing an incorrect tag option?

khiga8: I see. How does a component misuse get into production? I think I would have expected an error to be raised at dev stage that must be fixed. Or.. maybe this happens when an API changes later on and existing usage isn't corrected?

khiga8: Some of our components have restricted tags defined by `TAG_OPTIONS` and some don't. I think we've discussed that eventually we'll want to define allowed tags for all components and it seems like something that's [on the radar ](https://github.com/github/primer/issues/115)of PRC team as well!

Maybe we can have the tag validation check called in `initialize` of `BaseComponent` like:

```.rb
    def validate_tags(tag)
      return unless TAG_OPTIONS.present? # Remove once allowed tags are clearly defined for all components. 
      
      raise ArgumentError, "tag must be set to one of #{TAG_OPTIONS.join(', ')}" unless TAG_OPTIONS.include?(tag)
    end
```

But this depends on whether it's important to fail gracefully for incorrect tags and I don't really have context.
Is there a reason to provide this fallback for tags in production, or is it safe to fail loudly? cc: @manuelpuyol

joelhawksley: The most common route is untested code. Someone makes a change, it's not covered by a test, and they don't run it locally or in review lab. It's not ideal, but we've run into it before.

manuelpuyol: > How does a component misuse get into production? 

This does happen on big migrations (especially when we add a breaking change to a component). Heading may not be widely used, but we have some components, such as `Subhead` that are used hundreds of times.
When migrating all of those component calls to use a new API, you may miss 1 or 2 in untested places, which could lead to 500s

I think it is safer to not raise in production simply because we are quickly changing our components and would lead us to less headaches 😅

khiga8: I see, thanks for providing some context! 

I'll remove the default but keep a fallback to handle invalid tags.

app/components/primer/heading_component.rb
```diff
@@ -1,19 +1,47 @@
...
+  #   appropriate for the page context.
+  # - Use Heading as the title of a section or sub section.
+  # - Do not use Heading for styling alone. To style text without conveying heading semantics,
+  #   consider using <%= link_to_component(Primer::TextComponent) %> with relevant <%= link_to_typography_docs %>.
+  # - Do not jump heading levels. For instance, do not follow a `<h1>` with an `<h3>`. Heading levels should
+  #   increase by one in ascending order.
+  #
+  # @accessibility
+  #   Headings convey semantic meaning. Assistive technology users rely on headings to quickly navigate and scan information on a page.
+  #   Inappropriate use of heading can lead to a confusing experience so be sure to follow best practices.
```
joelhawksley, line 16:
```suggestion
  #   Inappropriate use of headings can lead to a confusing experience.
```

app/components/primer/heading_component.rb
```diff
@@ -1,19 +1,47 @@
...
+    #   <%= render(Primer::HeadingComponent.new(tag: :h5)) { "H5 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h6)) { "H6 Text" } %>
+    #
+    # @example With `font_size`
+    #   <%= render(Primer::HeadingComponent.new(tag: :h1, font_size: 6)) { "H1 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h2, font_size: 3)) { "H2 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h3, font_size: 2)) { "H3 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h4, font_size: 0)) { "H4 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h5, font_size: 1)) { "H5 Text" } %>
+    #   <%= render(Primer::HeadingComponent.new(tag: :h6, font_size: 4)) { "H6 Text" } %>
```
joelhawksley, line 38: I wonder if these examples might read more clearly as `h1 tag, font_size 6`.

app/components/primer/heading_component.rb
```diff
@@ -1,19 +1,47 @@
...
+  # @accessibility
+  #   Headings convey semantic meaning. Assistive technology users rely on headings to quickly navigate and scan information on a page.
+  #   Inappropriate use of heading can lead to a confusing experience so be sure to follow best practices.
+  #   [Learn more about best heading practices (WAI Headings)](https://www.w3.org/WAI/tutorials/page-structure/headings/)
   class HeadingComponent < Primer::Component
     status :beta
 
+    TAG_OPTIONS = [:h1, :h2, :h3, :h4, :h5, :h6].freeze
+    TAG_FALLBACK = :h2
+
```
joelhawksley, line 23: If you want to be fancy:

```suggestion
    TAG_OPTIONS = [:h1, TAG_FALLBACK, :h3, :h4, :h5, :h6].freeze
    TAG_FALLBACK = :h2

```

khiga8: i love ✨fancy

app/components/primer/heading_component.rb
khiga8: Good point!

app/components/primer/heading_component.rb
khiga8: I'll come back to whether we need a `@usage` tag/section after extending some other component documentations! :)

joelhawksley (approved):

PR #473, opened 2021-04-22
Extract button base

manuelpuyol:
Initial discussion: https://github.com/primer/view_components/pull/445#issuecomment-818828571

# Summary

We always want to use a component when rendering a Button since we can share some behaviors (like not adding `type='button'` when using a `summary` tag). Our current button's problem is that it always adds the `btn` style, which is needed in most cases, but not all.

We have some buttons like `CloseButton`, `ButtonMarketing` and `HiddenTextExpander` that don't need all of our `ButtonComponent` styles, but they do need the shared behaviors.

# Solution

I decided to extract a `BaseButton`, which is basically an unstyled button. It holds the logic for possible `types` and `tags` and turning buttons into `blocks`. In the future, we could add some accessibility features too!
The new component should be used any time you need a button without the need for `btn`, `schemes`, and `variants`.


app/components/primer/base_button.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Primer
+  # Use BaseButton to render an unstyles `<button>` tag that can be customized.
```
joelhawksley, line 4:
```suggestion
  # Use BaseButton to render an unstyled `<button>` tag that can be customized.
```

app/components/primer/base_button.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
```
joelhawksley, line 1: Do you think we should be exposing this component as a public API? IMO we should keep it private, at least to start.

manuelpuyol: I'm fine keeping it private, that's why I didn't add it to the docs sidebar 😄

joelhawksley: Ah, cool. 👍🏻

joelhawksley (approved):

PR #475, opened 2021-04-22
Only run Docs Build workflow on PR and skip if actor is dependabot

manuelpuyol:
Running the docs workflow on push was causing some trouble in local development since it may add unwanted commits when you push without opening a PR.
I'm also skipping dependabot PRs from this workflow since they shouldn't change any docs.

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,13 +1,10 @@
 name: Docs Build
 
-on:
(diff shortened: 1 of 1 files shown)


joelhawksley (approved):

manuelpuyol:
well, it looks like on `pull_request` we don't have access to the branch to push :/

PR #476, opened 2021-04-23
Add `icon` and `counter` slots to `ButtonComponent`

manuelpuyol:
Based on [buttons with icons](https://primer.style/css/components/buttons#button-with-icons) and [buttons with counter](https://primer.style/css/components/buttons#button-with-counts).

Note: This PR **does not** add support for icon only buttons nor `social-count`.


app/components/primer/button_component.rb
```diff
@@ -25,6 +25,24 @@ class ButtonComponent < Primer::Component
     }.freeze
     VARIANT_OPTIONS = VARIANT_MAPPINGS.keys
 
+    DEFAULT_ICON_ALIGN = :left
```
joelhawksley, line 28:
```suggestion
    DEFAULT_ICON_ALIGNMENT = :left
```

app/components/primer/button_component.rb
```diff
@@ -25,6 +25,24 @@ class ButtonComponent < Primer::Component
...
+    ICON_ALIGN_OPTIONS = [DEFAULT_ICON_ALIGN, :right].freeze
```
joelhawksley, line 29:
```suggestion
    ICON_ALIGNMENT_OPTIONS = [DEFAULT_ICON_ALIGNMENT, :right].freeze
```

app/components/primer/button_component.rb
```diff
@@ -25,6 +25,24 @@ class ButtonComponent < Primer::Component
...
+
+    # Icon to be rendered in the button.
+    #
+    # @param align [Symbol] <%= one_of(Primer::ButtonComponent::ICON_ALIGN_OPTIONS) %>
+    # @param system_arguments [Hash] Same arguments as <%= link_to_component(Primer::OcticonComponent) %>.
+    renders_one :icon, lambda { |align: DEFAULT_ICON_ALIGN, **system_arguments|
+      @icon_align = fetch_or_fallback(ICON_ALIGN_OPTIONS, align, DEFAULT_ICON_ALIGN)
```
joelhawksley, line 36:
```suggestion
    renders_one :icon, lambda { |align: DEFAULT_ICON_ALIGNMENT, **system_arguments|
      @icon_align = fetch_or_fallback(ICON_ALIGNMENT_OPTIONS, align, DEFAULT_ICON_ALIGNMENT)
```

app/components/primer/button_component.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= render Primer::BaseButton.new(**@system_arguments) do %>
+  <% if icon.present? && @icon_align == :left %>
+    <%= icon %>
+  <% end %>
+
+  <%= content %>
+  <%= counter %>
```
joelhawksley, line 7: @vdepizzol is it valid to have a `counter` _and_ and icon to the right of a the content of a button? IMO we should ensure this isn't the case.

vdepizzol: @joelhawksley I'm trying to visualize which cases where a right-aligned icon makes sense — rendering a `triangle-down` to indicate the button opens a menu or an overlay being the only example I can think of. And that feels like a boolean option that could appear even if the button already has a left-positioned icon. Is that something that could be supported with the icon alignment considerations from this PR?

Example:
> Screen Shot 2021-04-23 at 15 31 43

For a counter + right icon scenario, having it with a dropdown makes sense to me. Like in the example below:

> Screen Shot 2021-04-23 at 15 35 33

@manuelpuyol are there previously discussed use cases of a right-aligned icon that I'm missing? ✨ 

Thanks! 🙇 🙇

manuelpuyol: I just rolled with a comment made when I asked about it:

> We have icons on the right sometimes

I can't imagine any other icon other than the caret being right-aligned, though. @colebemis, do you know of any other use cases? I'm down to always aligning left and having a `caret` flag to render the caret

joelhawksley: @manuelpuyol I'd vote to start with the more restrictive case (left alignment, caret flag). We can always make the API more permissive later 😄

app/components/primer/button_component.html.erb
```diff
@@ -1,12 +1,9 @@
...
-  <% end %>
-
+  <%= icon %>
   <%= content %>
   <%= counter %>
 
-  <% if icon.present? && @icon_align == :right %>
-    <%= icon %>
+  <% if @caret %>
+    <%= primer_octicon(icon: "triangle-down") %>
```
joelhawksley, line 7:
```suggestion
    <%= primer_octicon("triangle-down") %>
```

test/components/button_component_test.rb
```diff
@@ -127,4 +127,15 @@ def test_renders_icon_and_counter
       assert_text("Button")
     end
   end
+
+  def test_renders_caret
+    render_inline(Primer::ButtonComponent.new(caret: true)) do |_c|
+      "Button"
+    end
```
joelhawksley, line 134:
```suggestion
    render_inline(Primer::ButtonComponent.new(caret: true).with_content("Button"))
```

joelhawksley (approved):

Gemfile.lock
```diff
@@ -166,7 +166,7 @@ GEM
     tzinfo (2.0.4)
       concurrent-ruby (~> 1.0)
     unicode-display_width (1.7.0)
-    view_component (2.30.0)
+    view_component (2.31.0)
```
manuelpuyol, line 169: for some reason my other PR updated to 2.30 instead of 2.31

PR #477, opened 2021-04-23
Extract LinkButton component

manuelpuyol:
`LinkButton` is simpler than what `ButtonComponent` has to offer, since it shouldn't have any schemes or variants. Because of that, I'm extracting it to its own component which calls `BaseButton`

image


app/components/primer/link_button.rb
```diff
@@ -0,0 +1,25 @@
...
+module Primer
+  # Use LinkButton to create a button that looks like a link rather than using an `<a>` to trigger JS.
+  class LinkButton < Primer::Component
+    # @example Default
+    #   <%= render(Primer::LinkButton.new) { "Link" } %>
+    #
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    def initialize(**system_arguments)
+      @system_arguments = system_arguments
+      @system_arguments[:tag] = :button
```
joelhawksley, line 12: Do we want to document that `tag` cannot be overridden? (And `type`)

khiga8:
I brought this up in a slack in `design-systems` as well, but it's unclear to me why/when we'd want to encourage a non-link element like a `Button` as a typical html link. Semantically it seems a little confusing and I'm having trouble grepping for where this pattern occurs in dotcom. Curious to hear if you had any thoughts/ideas!

PR #478, opened 2021-04-23
Add --inline option to component generator

manuelpuyol:
The `--inline` option will skip creating a template file and will add a `call` method to the component


joelhawksley (approved):

component_generator.thor
```diff
@@ -83,4 +84,8 @@ class ComponentGenerator < Thor::Group
   def js_package_name
     options[:js]
   end
+
+  def inline?
+    options[:inline]
```
joelhawksley, line 89: Might we want to document this option somewhere?

manuelpuyol: added a description!

PR #479, opened 2021-04-23
Create `IconButton` component

manuelpuyol:
Based on https://primer.style/css/components/buttons#icon-only-button


app/components/primer/icon_button.rb
```diff
@@ -0,0 +1,48 @@
...
+    #   <%= render(Primer::IconButton.new(icon: :search, label: "Search")) %>
+    #   <%= render(Primer::IconButton.new(icon: :trash, label: "Delete", scheme: :danger)) %>
+    #
+    # @param scheme [Symbol] <%= one_of(Primer::IconButton::SCHEME_OPTIONS) %>
+    # @param icon [String] Name of [Octicon](https://primer.style/octicons/) to use.
+    # @param label [String] String that will be read to screenreaders when the component is focused
+    # @param tag [Symbol] <%= one_of(Primer::BaseButton::TAG_OPTIONS) %>
+    # @param type [Symbol] <%= one_of(Primer::BaseButton::TYPE_OPTIONS) %>
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    def initialize(scheme: DEFAULT_SCHEME, icon:, label:, **system_arguments)
```
khiga8, line 30: For API consistency, I wonder if we should be surfacing the aria argument that's listed in [system_arguments](https://primer.style/view-components/system-arguments#html-attributes) rather than adding a new keyword (though I do see the benefit of mandating it like this!). 

I think that if it's important to set an `aria-label` for the component, we can mention it in the docs and examples, but otherwise it seems like it wouldn't get set via `**system_arguments`.

joelhawksley: I agree with @khiga8 comment 👍🏻. Having a `label` argument that doesn't set the `label` attribute is in conflict with the expectations set by `system_arguments`, which wrap `content_tag`.

manuelpuyol: I'd like to discuss a way of **requiring** the aria-label then. In this case, it is pretty hard to come up with a default, since we can use any icon which could mean anything.
what do you think of `raise ArgumentError if @system_arguments[:"aria-label].nil?` at the end of the `initialize`?

joelhawksley: I think that we do need to require it here, yes. I'd just want to make sure we don't end up with code path that could raise exceptions in production.

manuelpuyol: we can still follow the "don't raise in prod" strategy

app/components/primer/icon_button.rb
```diff
@@ -0,0 +1,48 @@
...
+    #
+    #   <%= render(Primer::IconButton.new(icon: :search, "aria-label": "Search")) %>
+    #
+    # @example Schemes
+    #
+    #   <%= render(Primer::IconButton.new(icon: :search, "aria-label": "Search")) %>
+    #   <%= render(Primer::IconButton.new(icon: :trash, "aria-label": "Delete", scheme: :danger)) %>
+    #
+    # @param scheme [Symbol] <%= one_of(Primer::IconButton::SCHEME_OPTIONS) %>
+    # @param icon [String] Name of [Octicon](https://primer.style/octicons/) to use.
```
joelhawksley, line 25: Could we write a helper method for this link?

test/components/icon_button_test.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    assert_selector("button[aria-label='Label'].btn-octicon") do
+      assert_selector(".octicon.octicon-star")
+    end
+  end
+
+  def test_does_not_render_content
+    render_inline(Primer::IconButton.new(icon: :star, "aria-label": "Label")) { "content" }
+
+    refute_text("content")
```
joelhawksley, line 19: Do you think we should provide stronger feedback (such as an error) if content is passed to components we don't want to accept it?

joelhawksley (approved):

test/components/icon_button_test.rb
manuelpuyol: we can come up with something in a later PR!

PR #486, opened 2021-04-26
Deprecate FlexItem in favor of Box

manuelpuyol:
`FlexItem` is simply a wrapper that accepts `flex_auto`, which is supported by system_arguments via `flex: :auto`.
Since it is a component without real meaning, I'm deprecating it in favor of the more generic `BoxComponent`


app/components/primer/flex_item_component.rb
```diff
@@ -3,7 +3,11 @@
 module Primer
   # Use FlexItemComponent to specify the ability of a flex item to alter its
   # dimensions to fill available space
+  #
+  # **Deprecation note:** Use <%= link_to_component(Primer::BoxComponent) %> instead.
```
joelhawksley, line 7: What do you think about providing an example before and after for the suggested migration?

joelhawksley (commented):
Looks good. I assume we'll also deprecate `Flex`?

manuelpuyol:
> Looks good. I assume we'll also deprecate `Flex`?

Yes, that'd be a follow up PR

app/components/primer/flex_item_component.rb
manuelpuyol: Mind if I add that in a follow-up PR? I'm planning on doing something similar to the `@accessibility` custom tag so we can have a section on deprecation

docs/content/components/flexitem.md
```diff
@@ -12,7 +12,21 @@ import Example from '../../src/@primer/gatsby-theme-doctocat/components/example'
...
 
-**Deprecation note:** Use [Box](/components/box) instead.
+## Deprecation
+
+Use [Box](/components/box) instead.
+
+**Before**:
+
+```erb
+ <%= render Primer::FlexItemComponent.new(flex_auto: :auto) %>
```
joelhawksley, line 22: Do we want leading spaces on these lines?

manuelpuyol: good catch, fixed it

joelhawksley (commented):
Should we have this component raise any deprecation warnings when it is rendered in an application?

manuelpuyol:
> Should we have this component raise any deprecation warnings when it is rendered in an application?

good idea! do you think that should happen on `initialize` or `before_render`?

joelhawksley:
> > Should we have this component raise any deprecation warnings when it is rendered in an application?
> 
> good idea! do you think that should happen on `initialize` or `before_render`?

I think `before_render`, but either is fine 👍🏻

manuelpuyol:
> > > Should we have this component raise any deprecation warnings when it is rendered in an application?
> > 
> > 
> > good idea! do you think that should happen on `initialize` or `before_render`?
> 
> I think `before_render`, but either is fine 👍🏻

added it to `initialize` 😬

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
 
     *Jon Rohan*
 
+* Depreacate `FlexItem` component in favor of `BoxComponent`.
```
joelhawksley, line 25:
```suggestion
* Deprecate `FlexItem` in favor of `BoxComponent`.
```

docs/content/components/flexitem.md
```diff
@@ -12,6 +12,22 @@ import Example from '../../src/@primer/gatsby-theme-doctocat/components/example'
...
+**Before**:
+
+```erb
+<%= render Primer::FlexItemComponent.new(flex_auto: :auto) %>
+```
+
+**After**:
+
+```erb
+<%= render Primer::BoxComponent.new(flex_auto: :auto) %>
```
joelhawksley, line 28: Is this correct? Or would we need `display: :flex` here as well?

CHANGELOG.md
joelhawksley: Also, do you have a plan for when this will be removed? If so, perhaps we could document that here or in the deprecation note?

docs/content/components/flexitem.md
manuelpuyol: we don't! `FlexItem` doesn't actually set `display: :flex`
The only thing it did was dealing with `flex_auto:`

joelhawksley (approved):

PR #488, opened 2021-04-26
Extract flex related CSS generation from classify

manuelpuyol:
This PR extracts the `Primer::Classify::Flex` class, which generates flex related CSS to Classify.
It provides explicit value validation and also adds the `flex_wrap` classes, which is needed to deprecate `FlexComponent`


app/lib/primer/classify/flex.rb
```diff
@@ -0,0 +1,121 @@
...
+      JUSTIFY_CONTENT_VALUES = [:flex_start, :flex_end, :center, :space_between, :space_around].freeze
+
+      ALIGN_ITEMS_KEY = :align_items
+      ALIGN_ITEMS_VALUES = [:flex_start, :flex_end, :center, :baseline, :stretch].freeze
+
+      KEYS = [FLEX_KEY, WRAP_KEY, SHRINK_KEY, GROW_KEY, ALIGN_SELF_KEY, DIRECTION_KEY, JUSTIFY_CONTENT_KEY, ALIGN_ITEMS_KEY].freeze
+      RESPONSIVE_KEYS = [DIRECTION_KEY, JUSTIFY_CONTENT_KEY, ALIGN_ITEMS_KEY].freeze
+
+      class << self
+        def css(key, value, breakpoint)
```
joelhawksley, line 39: IMO `classes` would be more idiomatic given our current naming patterns:

```suggestion
        def classes(key, value, breakpoint)
```

app/lib/primer/classify/flex.rb
```diff
@@ -0,0 +1,121 @@
...
+          "flex#{breakpoint}-items-#{formatted_value}"
+        end
+
+        def generate(value:, allowed_values:, key:, prefix:)
+          validate(value, allowed_values, key)
+
+          "#{prefix}-#{value}"
+        end
+
+        def validate(val, allowed_values, key)
```
joelhawksley, line 113: Could we use `fetch_or_fallback` in this case, or adapt it to handle it?

joelhawksley: If nothing else, this code feels like it could belong on `Classify` as it's more general purpose.

test/lib/classify_test.rb
```diff
@@ -609,21 +597,17 @@ def test_width_and_height
   def test_flex_grow
     assert_generated_class("flex-grow-0", { flex_grow: 0 })
 
-    err = assert_raises ArgumentError do
+    err =assert_raises Primer::FetchOrFallbackHelper::InvalidValueError do
```
joelhawksley:
```suggestion
    assert_raises Primer::FetchOrFallbackHelper::InvalidValueError do
```

joelhawksley (approved):

PR #494, opened 2021-04-27
Strip whitespace from classify class output

jonrohan:
I noticed that the class attributes contained a trailing whitespace character. Which results in weird test cases like the one in classify_test where you have to include the whitespace to pass the test. ie

```rb
def test_does_not_raise_error_when_passing_in_a_primer_css_class_otherwise
  assert_generated_class("bg-blue text-center float-left ml-1 ", { classes: "bg-blue text-center float-left ml-1" })
end
```

In this PR I'm stripping the trailing whitespace and adding a test.


joelhawksley (approved):
Mind adding a CHANGELOG?

PR #496, opened 2021-04-27
Remove requirement of octicons_helper and start using octicons gem directly

jonrohan:
This refactors out the use of the [octicons_helper](https://rubygems.org/gems/octicons_helper) gem and uses the [octicons](https://rubygems.org/gems/octicons) gem directly.

The reasoning behind this change. I'm currently working on making the octicons render on dotcom with a use tag https://github.com/github/design-systems/issues/1294 instead of inlining the path all the time. While exploring this work, it was clear that I would need to introduce a second helper to allow the inlining of the symbols with an api similar to `octicon_symbols([:icon, :keys, :foo, :bar])`. But what didn't sit well with me is that we have view_components now. When I first wrote the octicons helper there was no such technology. Now we have an excellent system, and to me it seemed like no use for the helper. 
 
Long term goals:

* `Primer::Octicon or primer_octicon` becomes the new octicon helper directly accessing the octicons gem.
* `Primer::OcticonSymbols` is a new component that accepts an array of keys that will render the necessary symbol defs for all the icons on a page.

The idea being that at the end of the buffer rendering, we will know what icons were rendered and inline the sheet.

```rb
<%= render Primer::OcticonSymbols.new(icons: [:alert, :x, :octocat]) %>
```

Drawbacks:

Currently there's caching [built into the octicons_helper](https://github.com/primer/octicons/blob/338afdf9e54ee593ff394bf1108e332338bb7919/lib/octicons_helper/lib/octicons_helper/helper.rb#L20) that caches any octicon that has the same arguments.
   * ~**_Option 1_**: We could move this caching to the main octicon_gem, this is where most of the options go anyways so I could see this working similar to how it currently works.~
   * ~**_Option 2_**: Maybe this should be something that functionally works in PVC? Wouldn't all the components benefit if we cache them based on arguments per page? So that when a component is rendered twice in the same page with the same arguments it pulls from cache the second time.~
   * **_Option 3_** ✅ Build caching into the OcticonComponent


app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,25 @@ class OcticonComponent < Primer::Component
...
+      @icon = Octicons::Octicon.new(icon_name || icon, octicon_options)
+
+      @system_arguments[:classes] = [@icon.options[:class], @system_arguments[:classes]].compact.join(" ")
+      @system_arguments.merge!(@icon.options.reject { |k| [:class].include?(k) })
     end
 
     def call
-      octicon(@icon, { **@system_arguments })
+      # rubocop:disable Rails/OutputSafety
+      render(Primer::BaseComponent.new(**@system_arguments)) { @icon.path.html_safe }
```
jonrohan, line 49: I'm not sure if there's a preferred way to handle this `html_safe` here.

joelhawksley: I think this is fine 👍🏻

joelhawksley (commented):
Looking good. I'd expect us to port the caching logic to this component, or into `octicons` itself before landing this PR 🙏🏻

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,25 @@ class OcticonComponent < Primer::Component
...
       # Note height and width are both classify options and valid html attributes.
-      octicon_helper_options = @system_arguments.slice(:height, :width)
-      @system_arguments = add_test_selector(@system_arguments).except(*Primer::Classify::VALID_KEYS, :classes).merge(octicon_helper_options)
+      octicon_options = {
+        height: SIZE_MAPPINGS[fetch_or_fallback(SIZE_OPTIONS, size, SIZE_DEFAULT)]
+      }.merge(@system_arguments.slice(:height, :width, :class))
+
+      @icon = Octicons::Octicon.new(icon_name || icon, octicon_options)
+
+      @system_arguments[:classes] = [@icon.options[:class], @system_arguments[:classes]].compact.join(" ")
```
manuelpuyol, line 43:
```suggestion
      @system_arguments[:classes] = class_names(
        @icon.options[:class],
        @system_arguments[:classes]
      )
```

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,25 @@ class OcticonComponent < Primer::Component
...
-      @system_arguments[:height] ||= SIZE_MAPPINGS[size]
+      @system_arguments[:tag] = :svg
 
       # Filter out classify options to prevent them from becoming invalid html attributes.
       # Note height and width are both classify options and valid html attributes.
-      octicon_helper_options = @system_arguments.slice(:height, :width)
-      @system_arguments = add_test_selector(@system_arguments).except(*Primer::Classify::VALID_KEYS, :classes).merge(octicon_helper_options)
+      octicon_options = {
+        height: SIZE_MAPPINGS[fetch_or_fallback(SIZE_OPTIONS, size, SIZE_DEFAULT)]
+      }.merge(@system_arguments.slice(:height, :width, :class))
```
manuelpuyol, line 39: this feels a bit weird to me, we are setting a specific `height` according to `size` but letting the `height` system_argument override it so I think we either want

```rb
@system_arguments[:height] = SIZE_MAPPINGS[fetch_or_fallback(SIZE_OPTIONS, size, SIZE_DEFAULT)]
octicon_options = @system_arguments.slice(:height, :width, :class)
```

or

```rb
octicon_options = {
  height: SIZE_MAPPINGS[fetch_or_fallback(SIZE_OPTIONS, size, SIZE_DEFAULT)]
}.merge(@system_arguments.slice(:width, :class))
```

also, I don't think we ever use the `class` key, it should probably be `classes` when adding custom-classes

jonrohan: We have a specific, custom height and width test, which I don't think we should have allowed, but removing that functionality made me worry there would be migrations on dotcom I would need to verify. So I wrote it in such a way that height passed in will override size.

app/components/primer/octicon_component.rb
```diff
@@ -29,28 +29,56 @@ class OcticonComponent < Primer::Component
...
+
+        Primer::OcticonComponent::Cache.set(cache_key, [@icon, @system_arguments])
+      end
     end
 
     def call
-      # rubocop:disable Rails/OutputSafety
-      render(Primer::BaseComponent.new(**@system_arguments)) { @icon.path.html_safe }
-      # rubocop:enable Rails/OutputSafety
+      render(Primer::BaseComponent.new(**@system_arguments)) { @icon.path.html_safe } # rubocop:disable Rails/OutputSafety
```
BlakeWilliams, line 60: Not saying we have to, or even should, but I could see us caching the markup and returning it directly instead of going through `render`.

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,56 @@ class OcticonComponent < Primer::Component
...
+      end
     end
 
     def call
-      octicon(@icon, { **@system_arguments })
+      render(Primer::BaseComponent.new(**@system_arguments)) { @icon.path.html_safe } # rubocop:disable Rails/OutputSafety
+    end
+
+    # :nodoc:
+    class Cache
```
manuelpuyol, line 64: I wonder if we could pre-populate some icons here, like we do with `Classify::Cache`

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,56 @@ class OcticonComponent < Primer::Component
...
+      # rubocop:disable Style/MutableConstant
+      LOOKUP = {}
+      # rubocop:enable Style/MutableConstant
```
manuelpuyol, line 67:
```suggestion
      LOOKUP = {} # rubocop:disable Style/MutableConstant
```

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,56 @@ class OcticonComponent < Primer::Component
...
+
+      class <<self
```
manuelpuyol, line 69:
```suggestion
      class << self
```

app/components/primer/octicon_component.rb
jonrohan: I don't know if we can make the assumption of what icons we'll be accessing. This also caches based on all system arguments passed to the component.

app/components/primer/octicon_component.rb
```diff
@@ -29,28 +29,56 @@ class OcticonComponent < Primer::Component
...
+        def read(key)
+          LOOKUP[key]
+        end
+
+        def set(key, value)
+          LOOKUP[key] = value
```
BlakeWilliams, line 73: As-is I think this will let the cache grow infinitely in size. I don't believe this is a problem for our uses, but it would be nice to somehow limit how much memory the cache can use.

I think we have a few options, like:

* Caching only known common values
* Implementing some kind of cache eviction strategy when the `LOOKUP` hash reaches a certain size

The first is easier, but would likely result in way more cache misses than hits, especially as the application changes. The second is more work but will probably result in a higher cache hit/miss ratio.

test/benchmarks/bench_octicons.rb
```diff
@@ -0,0 +1,42 @@
...
+    assert_allocations 43 do
+      Primer::OcticonComponent.new(**@options)
+    end
+  end
+
+  def bench_allocations_with_cache
+    Primer::OcticonComponent::Cache.clear!
+
+    Primer::OcticonComponent.new(**@options)
+    assert_allocations 9 do
```
BlakeWilliams, line 24: This is awesome 😍

BlakeWilliams (commented):
This is very cool!

app/components/primer/octicon_component.rb
jonrohan: yeah I would love for this to be more wide spread, whenever a component is used more than once on the page with the same input

app/components/primer/octicon_component.rb
manuelpuyol: I think it would be good to have simple cases, we have some icons that are used a lot:
e.g.: `x` icon without any arguments

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,87 @@ class OcticonComponent < Primer::Component
...
+          KEY_ORDER.push(key)
+
+          # Clear when the cache is too large.
+          LOOKUP.delete(KEY_ORDER.shift) if KEY_ORDER.size > LIMIT
```
BlakeWilliams, line 80: `Hash` supports `shift`, I think that would allow us to remove the `KEY_ORDER` constant?

```suggestion
          LOOKUP.shift if LOOKUP.size > LIMIT
```

jonrohan: oh awesome!

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,85 @@ class OcticonComponent < Primer::Component
...
+      class << self
+        def read(key)
+          LOOKUP[key]
+        end
+
+        def set(key, value)
+          LOOKUP[key] = value
+
+          # Remove first item when the cache is too large.
+          LOOKUP.shift if LOOKUP.size > LIMIT
```
BlakeWilliams, line 78: It might be worth stubbing the `LIMIT` and adding a test to ensure we properly evict values as the hash increases in size?

BlakeWilliams (approved):
I had one thought about testing the cache eviction, but this looks good!

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,85 @@ class OcticonComponent < Primer::Component
     # @param size [Symbol] <%= one_of(Primer::OcticonComponent::SIZE_MAPPINGS) %>
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     def initialize(icon_name = nil, icon: nil, size: SIZE_DEFAULT, **system_arguments)
-      @icon = icon_name || icon
+      icon_key = icon_name || icon
+      cache_key = [icon_key, size, system_arguments.slice(:height, :width)].to_s
```
BlakeWilliams, line 33:
```suggestion
      cache_key = [icon_key, size, system_arguments.slice(:height, :width)].join("_")
```

Minor, but I think that will create a cache key that doesn't look like an array.

joelhawksley (approved):

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,94 @@ class OcticonComponent < Primer::Component
...
+          Primer::OcticonComponent.new(icon: :mail)
+          Primer::OcticonComponent.new(icon: :pencil)
+          Primer::OcticonComponent.new(icon: :plus)
+          Primer::OcticonComponent.new(icon: :question)
+          Primer::OcticonComponent.new(icon: :repo)
+          Primer::OcticonComponent.new(icon: :search)
+          Primer::OcticonComponent.new(icon: :"shield-lock")
+          Primer::OcticonComponent.new(icon: :star)
+          Primer::OcticonComponent.new(icon: :trash)
+          Primer::OcticonComponent.new(icon: :x)
```
manuelpuyol, line 115: I think we could extract a
```rb
PRELOADED_ICONS = [:alert, :check, :"chevron-down", :clippy, :clock, :"dot-fill", :info, :"kebab-horizontal", :link, :lock, :mail, :pencil, :plus, :question, :repo, :search, :"shield-lock", :star, :trash, :x].freeze

def preload!
  PRELOADED_ICONS.each { |icon| Primer::OcticonComponent.new(icon: icon) }
end
```

app/components/primer/octicon_component.rb
```diff
@@ -27,20 +29,94 @@ class OcticonComponent < Primer::Component
...
+      @system_arguments.merge!(@icon.options.except(:class, :'aria-hidden'))
     end
 
     def call
-      octicon(@icon, { **@system_arguments })
+      render(Primer::BaseComponent.new(**@system_arguments)) { @icon.path.html_safe } # rubocop:disable Rails/OutputSafety
     end
+
+    # :nodoc:
+    class Cache
```
manuelpuyol, line 70: could we move this to a separate file?

manuelpuyol (approved):
only two non-blocking comments 👍

app/components/primer/octicon_component.rb
jonrohan: Yeah I agree we should, but I'd like to have some discussion on where. I can follow up after this PR is out.

PR #497, opened 2021-04-28
Remove border_color: :overlay

simurai:
The `border.overlay` will be removed from https://github.com/primer/primitives/pull/82. In preparation for that change, this PR removes `border_color: :overlay`. It can be replaced with `border_color: :primary` and should not cause a visual difference since both use the same color.

**Note**: I think `border_color: :overlay` is not yet used on dotcom, but we probably still wanna release this change before releasing https://github.com/primer/css/pull/1370.

This is part of https://github.com/github/design-systems/issues/1393.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@
 
     *Jon Rohan*
 
+* **Breaking change:** Remove `:overlay` option from `border_color`.
+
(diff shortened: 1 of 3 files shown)


joelhawksley (approved):
Mind adding a CHANGELOG?

Issue #491, opened 2021-04-27
Consolidate how we set default tag and allowed tags for accurate documentation


khiga8:
In [System arguments](https://primer.style/view-components/system-arguments), we have: 
`All Primer ViewComponents accept a standard set of options called system arguments`. This includes a `tag` option but it isn't necessarily true that a consumer can update tags. In reality, some of our components don't allow tags to be change since we set internally:
```
system_arguments[:tag] = :input
```

Other times, we have a default and allow tag to be changed based on user input:
```
system_arguments[:tag] ||= :input
```
OR
```
 def initialize(
      tag: DEFAULT_TAG,
```

And other times, we have a default tag which can be changed to limited tag options.
```
@system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, DEFAULT_TAG)
```

The documentation for components don't always surface this information making it unclear what the default tag is, what tags are allowed, etc. 

Our `Rake` populates the default tag section by looking the component initializer and checking if there's a tag keyword: 
```
 def initialize(
      tag: DEFAULT_TAG,
```
However, we don't always set default tag like this so that section in the doc ends up being`N/A` for most components. This is a problem because it is ambiguous without looking at the component source code what tags are allowed, what the default tag is, etc. 

TODOS:

- [x] Remove tag from system arguments docs since it's misleading
- [ ] Create tag allow list (currently known as TAG_OPTIONS) for the components and slots that do allow restricted set of tags. These allowed tags should be surfaced in the docs in Arguments table. Looks like we do this for a few components already like Button.
- [ ] For components that can render any tag, declare a tag parameter with a default. 
- [ ] For components that have fixed tags that can't be changed, no need to document in Arguments though we may mention what the tag is if it's important.
- [ ] Maybe.... update `Contributing` docs?

joelhawksley:
@khiga8 off the top of my head, I _think_ we should be explicitly allow listing tags for every component.

khiga8:
```
# remove tag from system arguments docs 
# create TAG_ALLOWLIST for each component and slot we want to limit 
# when tag has fixed tag, don't need to surface unless we want to 
# for components that can render any tag, declare a tag parameter with default 
```

khiga8:
TO REVIEW - components / slots that allow flexible tag

- [ ] avatar_stack_component 
- [ ] avatar_stack_component `body_arguments` 
- [ ] blankslate_component
- [ ] button_group
- [ ] details_component `:body` slot
[...]

PR #502, opened 2021-04-28
Allow `@param` tag to accept default value

khiga8:
Part of #491 

**What**
- Updates the yard tag`@param` by redefining the tag with `:with_types_name_and_default`. Currently,`@param` can only hold: 
```
@param name [Types] description
```
This change allows `@param` to also optionally hold a default value within parentheses so we can use this info when populating docs: 
```
@param name (Optional default value) [Types] description
```
- Update logic in `Rake` for populating the `Arguments` table to use default information on tag if its available. The current logic infers what the default is set by looking at the method argument. This may not work for cases like `Button` which inherits a default from `BaseButton`. This also makes it hard to surface defaults that are set within initialize that may not be part of method arguments.
- Updates a few component to see how documentation turns out 

**Why**
The documentation is missing some pieces of info that are useful for consumer to know. We have defaults that aren't exposed in the method arguments that we may want to surface manually. A good example is `AutoComplete`. 

**Notes**
- Instead of overriding `@param`, I considered introducing [@option](https://rubydoc.info/gems/yard/file/docs/Tags.md#option) which can already hold a default value. It isn't clear to me what should be `@param` and what should be `@option`.  Since we already use `@param` to refer to any argument, it feels more intuitive to stick to it.

 [Defining a custom tag docs](https://www.rubydoc.info/gems/yard/YARD/Tags/Library)
[Methods available for tag parsing docs](https://www.rubydoc.info/gems/yard/YARD/Tags/DefaultFactory#parse_tag_with_types_name_and_default-instance_method)

### DIFF
--- a/app/components/primer/auto_complete.rb
+++ b/app/components/primer/auto_complete.rb
@@ -5,22 +5,19 @@ module Primer
   class AutoComplete < Primer::Component
     status :beta
 
-    DEFAULT_TAG = :ul 
-    TAG_OPTIONS = [DEFAULT_TAG].freeze
+    DEFAULT_TAG = :"auto-complete"
 
     DEFAULT_INPUT_TYPE = :text
     INPUT_TYPE_OPTIONS = [DEFAULT_INPUT_TYPE, :search].freeze
     DEFAULT_INPUT_TAG = :input
-    INPUT_TAG_OPTIONS = [DEFAULT_INPUT_TAG].freeze
 
     DEFAULT_RESULTS_TAG = :ul
-    RESULTS_TAG_OPTIONS = [DEFAULT_RESULTS_TAG].freeze
 
     # Required input used to search for results
     #
-    # @param type [Symbol] <%= one_of(Primer::AutoComplete::INPUT_TYPE_OPTIONS) %>
+    # @param type [Symbol] (<%= pretty_value(Primer::AutoComplete::DEFAULT_INPUT_TYPE) %>) <%= one_of(Primer::AutoComplete::INPUT_TYPE_OPTIONS) %>
+    # @param tag [Symbol] (<%= pretty_value(Primer::AutoComplete::DEFAULT_INPUT_TAG) %>) <%= locked_tag_text %>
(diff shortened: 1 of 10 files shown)


app/components/primer/auto_complete.rb
```diff
@@ -5,15 +5,21 @@ module Primer
...
     INPUT_TYPE_OPTIONS = [DEFAULT_INPUT_TYPE, :search].freeze
+    DEFAULT_INPUT_TAG = :input
+
+    DEFAULT_RESULTS_TAG = :ul
 
     # Required input used to search for results
     #
-    # @param type [Symbol] <%= one_of(Primer::AutoComplete::INPUT_TYPE_OPTIONS) %>
+    # @param type [Symbol] (<%= pretty_value(Primer::AutoComplete::DEFAULT_INPUT_TYPE) %>) <%= one_of(Primer::AutoComplete::INPUT_TYPE_OPTIONS) %>
+    # @param tag [Symbol] (<%= pretty_value(Primer::AutoComplete::DEFAULT_INPUT_TAG) %>) <%= locked_tag_text %>
```
khiga8, line 19: Feels a little cluttered having to call `pretty_value` but need in order to retain symbol as string.

app/components/primer/auto_complete.rb
```diff
@@ -5,15 +5,21 @@ module Primer
   class AutoComplete < Primer::Component
     status :beta
 
+    DEFAULT_TAG = :"auto-complete"
```
khiga8, line 8: Based on [this conversation](https://github.com/primer/view_components/issues/491#issuecomment-827845486), wonder if we should explicitly have:

```
TAG_OPTIONS = [DEFAULT_TAG].freeze
```
and for the slot tags: 
```
INPUT_TAG_OPTIONS = [DEFAULT_INPUT_TAG].freeze
RESULTS_TAG_OPTIONS = [DEFAULT_RESULTS_TAG].freeze
```

It felt a bit overkill since there's only one allowed, fixed tag but maybe having an explicit allowed list of tags is good for consistency ¯\_(ツ)_/¯  open to thoughts

PR #506, opened 2021-04-28
Deprecate tag from system_arguments docs

khiga8:
**What**
Remove `tag` from our System arguments docs

**Why**
The top of our[ System argument docs](https://primer.style/view-components/system-arguments) page states: `All Primer ViewComponents accept a standard set of options called system arguments`. Came up in discussion with @joelhawksley that this is not true for `tag:` for most of our components.  Many of our components have fixed tags that can't be updated even if a user were to pass in a `tag:` argument. 

Let's remove it from the docs as it is misleading.

**Note**
Part of #491

### DIFF
--- a/app/components/primer/base_component.rb
+++ b/app/components/primer/base_component.rb
@@ -138,7 +138,6 @@ module Primer
     # | Name | Type | Description |
     # | :- | :- | :- |
     # | classes | String | CSS class name value to be concatenated with generated Primer CSS classes. |
-    # | tag | Symbol | HTML tag name to be passed to `content_tag`. |
     # | test_selector | String | Adds `data-test-selector='given value'` in non-Production environments for testing purposes. |
     def initialize(tag:, classes: nil, **system_arguments)
       @tag = tag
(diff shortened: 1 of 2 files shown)


manuelpuyol:
should we also update the docs on components that do allow tag changes?

khiga8:
> should we also update the docs on components that do allow tag changes?

@manuelpuyol 
Yes! I'll follow up in an upcoming PR!

In discussion with @joelhawksley, we're thinking of the following: 

- Remove tag from system arguments docs since it's misleading **this PR**
- Create tag allow list (currently known as `TAG_OPTIONS`) for the components and slots that do allow restricted set of tags. These allowed tags should be surfaced in the docs in `Arguments` table. Looks like we do this for a few components already like `Button`. 
- For components that can render any tag, declare a tag parameter with a default. Also surface this information in `Arguments`. I don't think there will be too many components that accept any tag??
- For components that have fixed tags that can't be changed, no need to document in `Arguments` though we may mention what the tag is if it's important.

manuelpuyol (approved):

PR #510, opened 2021-04-29
Restrict allowed tags for `AvatarStack`

khiga8:
**What**
- Restricts allowed tags for `AvatarStack` to `:div` and `:span`. I grepped dotcom for `AvatarStackComponent` initialized with `tag:` and did not find any. I did see a usage of non-primer `AvatarStack` on a `<span>`. Based off this, `:div` and `:span` as allowed tags and keeping the current `:div` seems reasonable.
- Updates doc with allowed tags. 

**Why**
Currently `AvatarStack` accepts any tags which we probably don't want. 

**Notes**
Part of #491


koddsson (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main
 
+* **Breaking change:** Explicitly limit tag for `AvatarStack`
```
joelhawksley, line 5:
```suggestion
* **Breaking change:** Explicitly limit tag for `AvatarStack` to `:div` and `:span`.
```

joelhawksley (approved):

PR #513, opened 2021-04-29
Restrict tag to`:div` for `Blankslate`

khiga8:
Part of #491 

**What**
Don't allow `Blankslate` tag to be changed. Currently we have no restrictions but we only ever instantiate this component as a `div`. We already don't mention the `tag` attribute in the docs here so no need to change anything here. 

**Notes**
 I verified that there's no instance of `Blankslate` with `tag` argument by grepping the following regex in dotcom:
```
Primer::BlankslateComponent.new(((?!%>).)|\n)*(tag:)+(((?!%>).)|\n)*%>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* **Breaking change:** Restrict `Blankslate` tag to `div`.
+
+    *Kate Higa*
+
 * **Breaking change:** Explicitly limit tag for `AvatarStack` to `:div` and `:span`
 
     *Kate Higa*
--- a/app/components/primer/blankslate_component.rb
+++ b/app/components/primer/blankslate_component.rb
@@ -111,7 +111,7 @@ module Primer
       **system_arguments
     )
       @system_arguments = system_arguments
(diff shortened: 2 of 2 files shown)


koddsson (approved):
This change looks good to me. ✨ 

Can we document that it is invalid to pass the `tag` system argument to this component?

joelhawksley (approved):

khiga8:
> This change looks good to me. ✨
> 
> Can we document that it is invalid to pass the `tag` system argument to this component?

hey @koddsson!  For a bit of context, we recently deprecated `tag:` from the System argument docs ( https://github.com/primer/view_components/pull/506) as the following statement wasn't true with regards to `tag`:

```
All Primer ViewComponents accept a standard set of options called system arguments, 
```

With this update, we're trying to eliminate the notion that primer view components accept a `tag` unless it's explicitly documented in the `Arguments` section of the component page.

We're moving towards the following [plan](https://github.com/primer/view_components/pull/506#issuecomment-828731247), so for now we're thinking that we don't need to mention `tag` for components that have locked tags like this.

PR #520, opened 2021-04-30
Make `&hellip;` html safe when rendering HiddenTextExpander

manuelpuyol:
Without the `html_safe` call, the page rendered the actual `&hellip;` text instead of `...`


manuelpuyol:
> LGTM!
> 
> any idea why this looks fine on the prod docs page?

probably something in the HTML -> markdown conversion

PR #523, opened 2021-04-30
Split rake file into separate files

### LINKED ISSUE #522: Break up `Rakefile` tasks into separate files.
`Rakefile` is getting a bit gigantic so let's break it up by task

Reference: https://github.com/primer/view_components/issues/467#issuecomment-825068041

khiga8:
Closes #522 

**What**
This PR extracts out some of the tasks in our `Rakefile` file into `lib/tasks`.
This is organized by task namespace like so:

- `lib/tasks/docs.rake`
- `lib/tasks/coverage.rake`
- `lib/tasks/statuses.rake`

This PR also pulls out the yard documentation related methods currently inside the docs build task into its own module `YARD::DocsHelper`.

**Why**
The `Rakefile` is getting very gigantic and splitting it up will make it easier to navigate .

**Notes**
Open to suggestions on where things belong/naming in case I'm not following conventions.


joelhawksley (approved):
Looks good! Do you think we should add tests for any of this code (in a later PR, of course)

khiga8:
> Looks good! Do you think we should add tests for any of this code (in a later PR, of course)

@joelhawksley I think so! Ideally we extract out these tasks into classes that can be tested so we can be confident that it's doing what we're expecting it to. :)  I'll throw in a task to the project board

PR #524, opened 2021-04-30
add test to ensure classify doesn't reference nonexistent CSS classes

joelhawksley:
In our engineering chat yesterday, we realized that it was possible, if not likely that Primer ViewComponents could reference Primer CSS classes that don't exist. 

This PR introduces a basic test for the values we cache in Primer::Classify. It already caught several gaps, one of which we've addressed.


test/lib/classify/cache_test.rb
```diff
@@ -10,4 +10,39 @@ def test_clear_clears_the_cache
...
+    css_data =
+      JSON.parse(
+        File.read(
+          File.join(
+            __FILE__.split("test")[0], "/node_modules/@primer/css/dist/stats/primer.json"
+          )
+        )
+      )
+
+    # TODO remove these
```
joelhawksley, line 34:
```suggestion
    # TODO: remove these
```

jonrohan (approved):

PR #526, opened 2021-04-30
Creating OcticonSymbolComponent

jonrohan:
This PR creates the OcticonSymbolComponent that doesn't have arguments directly, but has `icons` slots that each take arguments for `:symbol, :size`

I'm looking for feedback on API and naming of variables. Check out the docs for current naming https://view-components-6divzhql3-primer.vercel.app/view-components/components/octiconsymbol


manuelpuyol:
What happens if I use `render(Primer::OcticonComponent.new(icon: :check, use_symbol: true))` but I don't have a `OcticonSymbolComponent` calling `icon: :check`?

jonrohan:
> What happens if I use `render(Primer::OcticonComponent.new(icon: :check, use_symbol: true))` but I don't have a `OcticonSymbolComponent` calling `icon: :check`?

You wouldn't see an icon, it would be missing.

manuelpuyol:
`OcticonSymbolComponent` is rendering an element in the DOM, is that expected? Shoulfn't the root `svg` be `display: none`?

image

manuelpuyol:
> > What happens if I use `render(Primer::OcticonComponent.new(icon: :check, use_symbol: true))` but I don't have a `OcticonSymbolComponent` calling `icon: :check`?
> 
> You wouldn't see an icon, it would be missing.

Could it check if the symbol exists? Or raise some kind of error in development? I can see a few people copying the code from somewhere else and using on their views then having no clue why their icon is not being rendered

jonrohan:
> Could it check if the symbol exists? Or raise some kind of error in development? I can see a few people copying the code from somewhere else and using on their views then having no clue why their icon is not being rendered

I'm not sure it could, it's relying on browser technology to lookup the symbol. Our intention is to put the dictionary at the bottom of the page, so we'll not know if it's been used or not. @joelhawksley any ideas here?

app/components/primer/octicon_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= render(Primer::BaseComponent.new(**@system_arguments)) do %>
+  <% if @use_symbol %>
+    <use href="#octicon-<%= [@icon.symbol, @icon.height].join("-") %>"></use>
```
jonrohan, line 3: I wonder if there's performance here I can do by not looking up the `@icon` from cache or the gem, since I should have these 2 bits of data.

app/components/primer/octicon_component.rb
```diff
@@ -31,14 +31,16 @@ class OcticonComponent < Primer::Component
     #
     # @param icon [String] Name of <%= link_to_octicons %> to use.
     # @param size [Symbol] <%= one_of(Primer::OcticonComponent::SIZE_MAPPINGS) %>
+    # @param use_symbol [Boolean] Set to true when you setup a symbol dictionary using <%= link_to_component(Primer::OcticonSymbolComponent) %> on the page.
```
jonrohan, line 34: Is `use_symbol` the best name? Could be `use:` or ??

app/components/primer/octicon_symbol_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" hidden>
```
jonrohan, line 1: I didn't see any a11y concerns locally with adding `hidden` here, but would be curious to know if there's something I'm missing.

app/components/primer/octicon_symbol_component.rb
```diff
@@ -0,0 +1,68 @@
...
+    #     <%= c.icon(symbol: :check) %>
+    #     <%= c.icon(symbol: :check, size: :medium) %>
+    #   <% end %>
+    def initialize
+      super
+    end
+
+    def before_render
+      # Make sure we don't have any duplicate icons
+      icons.uniq! { |icon| "#{icon.icon.symbol}#{icon.icon.height}" }
```
jonrohan, line 28: There's a possibility that someone might try and get a size of an icon that doesn't exist. In that case the library will default to the closest size. Which could end up with the same symbol in the defs. This won't break anything with rendering, but seemed like a waste of resources?

docs/src/@primer/gatsby-theme-doctocat/nav.yml
```diff
@@ -70,6 +70,8 @@
       url: /components/markdown
     - title: Octicon
       url: /components/octicon
+    - title: OcticonSymbol
+      url: /components/octiconsymbol
```
jonrohan, line 74: Is `OcticonSymbolComponent` the best name and is `/components/octiconsymbol` a url that makes sense?

test/components/component_test.rb
```diff
@@ -66,7 +66,7 @@ class PrimerComponentTest < Minitest::Test
   ].freeze
 
   def test_registered_components
-    ignored_components = ["Primer::Component"]
+    ignored_components = ["Primer::Component", "Primer::OcticonsSymbolComponent"]
```
jonrohan, line 69: Ignored because the symbol component doesn't directly take any arguments

app/components/primer/octicon_component.rb
manuelpuyol: I think that the name is fine

app/components/primer/octicon_symbol_component.html.erb
manuelpuyol: @khiga8 👀

app/components/primer/octicon_symbol_component.rb
```diff
@@ -0,0 +1,68 @@
...
+          tag: :symbol,
+          id: "octicon-#{@icon.symbol}-#{@icon.height}",
+          viewBox: @icon.options[:viewBox],
+          width: @icon.width,
+          height: @icon.height
+        }
+      end
+
+      def call
+        render(Primer::BaseComponent.new(**@system_arguments)) { @icon.path.html_safe } # rubocop:disable Rails/OutputSafety
```
joelhawksley, line 64: If you want to save some performance, I think we could just use `content_tag` here.

app/components/primer/octicon_component.html.erb
jonrohan: I think I'll need to keep it the way it is, since the icon from octicons is still needed for some of the system_arguments. Maybe in the future, the gem just becomes a lookup gem.

app/components/primer/octicon_symbol_component.rb
```diff
@@ -61,7 +60,7 @@ def initialize(symbol:, size: Primer::OcticonComponent::SIZE_DEFAULT)
       end
 
       def call
-        render(Primer::BaseComponent.new(**@system_arguments)) { @icon.path.html_safe } # rubocop:disable Rails/OutputSafety
+        content_tag(:symbol, @icon.path.html_safe, @system_arguments) # rubocop:disable Rails/OutputSafety
```
joelhawksley, line 63: As we're not sending these arguments into Classify, might we call them `tag_arguments`?

app/components/primer/octicon_symbol_component.rb
```diff
@@ -50,17 +50,10 @@ def initialize(symbol:, size: Primer::OcticonComponent::SIZE_DEFAULT)
...
-          id: "octicon-#{@icon.symbol}-#{@icon.height}",
-          viewBox: @icon.options[:viewBox],
-          width: @icon.width,
-          height: @icon.height
-        }
       end
 
       def call
-        content_tag(:symbol, @icon.path.html_safe, @system_arguments) # rubocop:disable Rails/OutputSafety
+        content_tag(:symbol, @icon.path.html_safe, id: "octicon-#{@icon.symbol}-#{@icon.height}", viewBox: @icon.options[:viewBox], width: @icon.width, height: @icon.height) # rubocop:disable Rails/OutputSafety
```
joelhawksley, line 56: Is this a semantic usage of `id`? I've always thought they should globally unique on the page.

app/components/primer/octicon_component.rb
joelhawksley: What is the long-term plan for this argument? Are we hoping to move _all_ implementations of Octicon to use this approach? I feel like it could be confusing to pass `true` here and not receive any error message if OcticonSymbol isn't then rendered on the page.

app/components/primer/octicon_symbol_component.rb
```diff
@@ -0,0 +1,60 @@
...
+    # @param symbol [String] Name of <%= link_to_octicons %> to use.
+    # @param size [Symbol] <%= one_of(Primer::OcticonComponent::SIZE_MAPPINGS) %>
+    renders_many :icons, "Symbol"
+
+    # @example Symbol dictionary
+    #   <%= render(Primer::OcticonComponent.new(icon: :check, use_symbol: true, color: :icon_success)) %>
+    #   <%= render(Primer::OcticonComponent.new(icon: :check, use_symbol: true, color: :text_danger)) %>
+    #   <%= render(Primer::OcticonComponent.new(icon: :check, use_symbol: true, size: :medium)) %>
+    #   <%= render(Primer::OcticonSymbolComponent.new) do |c| %>
+    #     <%= c.icon(symbol: :check) %>
```
joelhawksley, line 19: My gut says that we might be better off passing the icons in as a hash, as we'll be instantiating an object for each icon in this case.

app/components/primer/octicon_component.rb
joelhawksley, line 34:
```suggestion
    # @param use_symbol [Boolean] EXPERIMENTAL (May change or be removed) - Set to true when using with <%= link_to_component(Primer::OcticonSymbolComponent) %>.
```

app/components/primer/octicon_component.html.erb
joelhawksley, line 3: It's a total nitpick, but ids are usually underscored.

app/components/primer/octicon_symbol_component.html.erb
khiga8: whoops sorry I missed this! 

It looks like there's pretty decent support for `hidden` now based on [this][0]. 
But... the hidden attribute can easily be overridden if there's CSS setting the `display`. For more reliable support I might consider  using `style="display: none;"`

[0]: https://www.tpgi.com/the-state-of-hidden-content-support-in-2016/

Sources: 
https://www.tpgi.com/the-state-of-hidden-content-support-in-2016/
https://css-tricks.com/the-hidden-attribute-is-visibly-weak/
https://meowni.ca/hidden.is.a.lie.html

khiga8: Side note: It looks like there's [no need] for `<defs>` given we're rendering `<symbol>` !

[no need]: https://css-tricks.com/svg-symbol-good-choice-icons/#why-symbol-is-better-for-icons

jonrohan: We do have a selector in primer for `[hidden]` to set it display none I think Monica suggests in her post. https://github.com/primer/css/blob/63764f7edd59c5f4949c91e5373471aa8b0d4865/src/base/normalize.scss#L84

joelhawksley (approved):

app/components/primer/octicon_symbols_component.rb
```diff
@@ -13,7 +13,26 @@ class OcticonSymbolsComponent < Primer::Component
     #
     # @param icons [Array<Hash>] List of icons to render, in the format { symbol: :icon_name }
```
manuelpuyol, line 14: should we mention that you can also pass `size` to the hash?

app/components/primer/octicon_symbols_component.rb
```diff
@@ -13,7 +13,26 @@ class OcticonSymbolsComponent < Primer::Component
...
+        if (cache_icon = Primer::Octicon::Cache.read(cache_key))
+          icon_instance = cache_icon
+        else
+          icon_instance = Octicons::Octicon.new(symbol, height: size)
+
+          Primer::Octicon::Cache.set(cache_key, icon_instance)
+        end
+
+        # Don't put the same icon twice
+        @icons[[symbol, icon_instance.height]] = icon_instance if @icons[[symbol, icon_instance.height]].nil?
```
manuelpuyol, line 34: is `icon_instance.height == size`?

if so, we could add a 
`next if @icons[[symbol, size]].present?` before reading the cache

jonrohan: Not always. If someone asks for a medium size and the octicon gem doesn't have a 24px size, it will return the 16px octicon. In that case size would equal medium and height would equal 16

PR #527, opened 2021-05-03
Preload cache using _after_compile

joelhawksley:
I _believe_ this use of `_after_compile` should be equivalent to what we had before, and perhaps could serve as a nice example to point to when folks ask what the lifecycle hook is useful for!


manuelpuyol (approved):
LGTM!
When were we preloading before? 🤔

joelhawksley:
> When were we preloading before? 🤔

When the class was first loaded by Ruby 😄

PR #530, opened 2021-05-03
Set tag for Text and ButtonGroup

khiga8:
**What**
- Add `tag` to `Text` component docs for explicitness. 
- Restrict `ButtonGroup` tag to `:div`

**Notes** 
Part of #491

### DIFF
--- a/app/components/primer/text_component.rb
+++ b/app/components/primer/text_component.rb
@@ -5,14 +5,17 @@ module Primer
   class TextComponent < Primer::Component
     status :beta
 
+    DEFAULT_TAG = :span
+
     # @example Default
     #   <%= render(Primer::TextComponent.new(tag: :p, font_weight: :bold)) { "Bold Text" } %>
     #   <%= render(Primer::TextComponent.new(tag: :p, color: :text_danger)) { "Danger Text" } %>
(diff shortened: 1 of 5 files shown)


app/components/primer/text_component.rb
```diff
@@ -5,14 +5,17 @@ module Primer
   class TextComponent < Primer::Component
     status :beta
 
+    DEFAULT_TAG = :span
+
     # @example Default
     #   <%= render(Primer::TextComponent.new(tag: :p, font_weight: :bold)) { "Bold Text" } %>
     #   <%= render(Primer::TextComponent.new(tag: :p, color: :text_danger)) { "Danger Text" } %>
     #
+    # @param tag [Symbol]
```
khiga8, line 14: This follows current implementation of having no tag restrictions but I'm open to thoughts on if we should be restricting. 

I went with not restricting because this seems like a pretty generic component whose purpose is to give html elements access to primer typography like fontsize and text alignment. These typography classes seems applicable to a wide array of html tags based on [primer typography][1] examples

[1]: https://primer.style/css/utilities/typography

joelhawksley (approved):

PR #531, opened 2021-05-03
Validate aria-label by checking aria-label and aria: { label }

manuelpuyol:
`IconButton` was only checking `system_arguments[:"aria-label"]` instead of also checking `system_arguments[:aria][:label]`, which caused an incorrect raise


khiga8 (approved):
nice catch!

Issue #284, opened 2021-02-25
Run accessibility audits automatically


srt32:
Let's find a way to use https://github.com/dequelabs/axe-core

Some options:
* https://www.npmjs.com/package/@storybook/addon-a11y
* https://github.com/microsoft/accessibility-insights-action

khiga8:
Previous PR only addressed partially so reopening

joelhawksley:
5/17 update: @khiga8 has another PR in progress for this issue.

PR #543, opened 2021-05-04
Integrate `axe-core` so that CI build fails

khiga8:
Partially addresses #284
 
**What**
- Bumps `cuprite`  gem up to get `axe-core-api` working. 
- Introduces `axe-core-api` gem
- Introduces `assert_accessible` method which wraps `axe-core-api` methods to run accessibility assertions.
- Adds `assert_accessible` to `def with_preview(preview_name)` to ensure accessibility checks are run in every integration test on page load.
- Overrides Capybara methods that trigger UI change so axe check gets ran automatically
- Update `Autocomplete` component in preview with missing label.

**Todo in separate PR**

- More comprehensive `Autocomplete` component doc update with every example having label. 
- Integration tests for every component to run checks.


test/application_system_test_case.rb
```diff
@@ -4,16 +4,52 @@
...
 
+  AXE_RULES_TO_SKIP = [:region].freeze
+  AXE_WITHIN_SELECTOR = "body"
+
   def with_preview(preview_name)
     component_uri = self.class.name.gsub("Test", "").gsub("Integration", "").underscore
 
     visit("/rails/view_components/primer/#{component_uri}/#{preview_name}")
+
+    assert_accessible(page)
```
joelhawksley, line 25: Does this mean that all of our existing previews pass Axe?

khiga8: Of the 5 component previews, `Autocomplete` failed. The component renders an input without any labels. I made some updates here to address it but I'm planning to review the component more thoroughly in a follow up PR!

test/application_system_test_case.rb
```diff
@@ -4,16 +4,52 @@
...
+require "axe/expectation"
+
 require "test_helpers/cuprite_setup"
 
 Ferrum::Browser.new(process_timeout: 60, timeout: 60)
 
 class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   driven_by :cuprite, using: :chrome, screen_size: [1400, 1400]
 
+  AXE_RULES_TO_SKIP = [:region].freeze
```
koddsson, line 17: Can we add a comment for this? Reading the code, it's not apparent to me why we are skipping regions.

test/application_system_test_case.rb
```diff
@@ -4,16 +4,52 @@
...
+        key, value = option
+        a.send(key, *value)
+      end
+    end
+
+    Axe::AccessibleExpectation.new.assert page, is_axe_clean
+  end
+
+  # Capybara Overrides
+  # From axe-core docs- "insert calls at each point in your tests where a new piece of UI becomes visible or exposed"
```
koddsson, line 43: Are `fill_in` and `click_button` the only ones that change the UI? Do we need these for other APIs as well?

app/components/primer/auto_complete.rb
```diff
@@ -2,6 +2,8 @@
 
 module Primer
   # Use `AutoComplete` to populate input values from server search results.
+  # @accessibility
+  #   Always provide a label for the `Autocomplete` component.
```
koddsson, line 6: Is this something that should be built into the component, so users can't make the component inaccessible?

koddsson (commented):
Can we expand this beyond just the systems tests as a follow-up?

manuelpuyol:
> Can we expand this beyond just the systems tests as a follow-up?

I don't think this will work for unit tests since we don't have a browser available there

test/application_system_test_case.rb
manuelpuyol: those are the only ones we use right now

koddsson:
> > Can we expand this beyond just the systems tests as a follow-up?
> 
> I don't think this will work for unit tests since we don't have a browser available there

I think it's a more extensive discussion then, but should we run all the tests in the browser to do accessibility checks on them?

manuelpuyol:
> > > Can we expand this beyond just the systems tests as a follow-up?
> > 
> > 
> > I don't think this will work for unit tests since we don't have a browser available there
> 
> I think it's a more extensive discussion then, but should we run all the tests in the browser to do accessibility checks on them?

yeah I think we do need previews for all components at some point, we don't do it now because it would be a lot slower than unit tests. We are also running into some flakiness in system tests, we need to look into that before expanding it

app/components/primer/auto_complete.rb
khiga8: Yep! I plan to follow up in a separate PR for this component.

test/application_system_test_case.rb
koddsson: Can we implement the other APIs? For example, it's not clear that if I'm writing a system test and start using a new API that my component can become inaccessible without the tests catching it.

joelhawksley (approved):

test/application_system_test_case.rb
khiga8: Yes! I'll do this in a followup :)

Issue #544, opened 2021-05-04
Change size argument for `octicons` viewcomponent


ashygee:
I was taking a look at @jonrohan's [issue](https://github.com/primer/view_components/pull/526) to update the octicon component to use the SVG `use` tag and noticed that for the [size argument](https://primer.style/view-components/components/octicon#arguments) we're using t-shirt sizing and also are providing sizes that are outside of the [octicons guidelines](https://primer.style/octicons/guidelines/usage#sizing).

> Octicons are available in two sizes: 16px and 24px. They should only be used at their dedicated sizes to maintain consistency in stroke width and **legibility.**

I want to propose that we change the allowed values to use the numerical pixel values of either 16 or 24 which we currently allow the icons to use.

PR #547, opened 2021-05-05
Update OcticonComponent sizes to only 16 and 24

jonrohan:
Closes https://github.com/primer/view_components/issues/544

@ashygee pointed out we're not using standard sizes in the octicon component. The docs say to use 16 or 24 for the icon sizes. This PR drops the `:large` option and changes `:medium` to 24.

I looked up everywhere we're using `:large` in github/github and listed below, these would need to be updated.
* https://github.com/github/github/blob/e7753131d6400f2285b171c035ea990ef80fa639/app/views/device_authorization/success.html.erb#L2
* https://github.com/github/github/blob/e7753131d6400f2285b171c035ea990ef80fa639/app/components/explore/event/related_item_component.html.erb#L24
* https://github.com/github/github/blob/94a4322b78231cb3a8a35ae8bffdd7d5bc871f7b/app/components/device_authorization/failure_component.html.erb#L2


test/components/octicon_component_test.rb
```diff
@@ -56,13 +56,13 @@ def test_renders_default_size_small
...
 
   def test_renders_the_provided_size
-    render_inline(Primer::OcticonComponent.new(:star, size: :large))
+    render_inline(Primer::OcticonComponent.new(:star, size: :medium))
 
-    assert_selector("[height='64']")
+    assert_selector("[height='24']")
   end
 
   def test_renders_with_overridden_height_and_width_despite_given_a_size
```
joelhawksley, line 64: I'd be interested to hear if we still want to allow this override.

joelhawksley (approved):

test/components/blankslate_component_test.rb
```diff
@@ -54,17 +54,17 @@ def test_renders_a_blankslate_component_with_an_icon
...
 
-    assert_selector(".blankslate-icon[height=32]")
+    assert_selector(".blankslate-icon[height=24]")
   end
 
   def test_renders_a_blankslate_component_with_an_icon_with_a_custom_size
     render_inline(Primer::BlankslateComponent.new(
                     icon: "octoface",
-                    icon_size: :large,
+                    icon_size: :medium,
```
joelhawksley, line 63: Dod you check to see if any blankslate components use the `large` option?

jonrohan: I did, there isn't 
image

test/components/octicon_component_test.rb
jonrohan: I don't think we should, but this was added after the fact in https://github.com/primer/view_components/pull/100 

Searching github/github there's not too many places that don't use standard sizes. But this does bring up the fact that  a lot of places are using `height` and not `size` eg. `<%= render Primer::OcticonComponent.new(icon: "squirrel", height: 24) %>`

* https://github.com/github/github/blob/197c0ccba0f24c060163f423c39d205758bb720c/app/components/sponsors/dependencies_banner_component.html.erb#L5
* https://github.com/github/github/blob/197c0ccba0f24c060163f423c39d205758bb720c/app/components/sponsors/profile/corporate_sponsor_beta_discount_component.html.erb#L7
* https://github.com/github/github/blob/197c0ccba0f24c060163f423c39d205758bb720c/app/views/businesses/billing_settings/budgets/_budget_filters.html.erb#L11
* https://github.com/github/github/blob/197c0ccba0f24c060163f423c39d205758bb720c/app/views/layouts/bootstrap_instance.html.erb#L12

jonrohan:
Putting this back in draft for a moment, I realized that we're using `github-mark` in the header and setting `height: 32` on it. Which might mean we'll either need to keep allowing size overrides with height or not change medium to 24.

test/components/octicon_component_test.rb
jonrohan: After more searching, I think we should work up a longer term plan to deprecate the support of custom sizes. There's a lot of places in dotcom that use `height:` instead of `size:` and there's a few instances where a valid icon size 24 would break because they're using height: 32

PR #548, opened 2021-05-05
Only run CI builds once per PR

jonrohan:
I was noticing that the CI builds run twice per Pull Request.

image

This PR updates the action script to run on pushes on main and any commits to a pr.


.github/workflows/ci.yml
```diff
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
manuelpuyol, line 7: I'm 99% sure we had to have both because of forks

jonrohan: We can use `pull_request_target` which is what I do on primer/css that will run on forks

jonrohan: Example here https://github.com/primer/css/pull/1357

manuelpuyol (dismissed):

jonrohan:
> wait

uh oh

PR #555, opened 2021-05-07
[exploration] action for storybook a11y report failure

khiga8:
A non-working exploration just to see what a GitHub action using [storybook-a11y-report](https://github.com/kalbeekatz/storybook-a11y-report)  look like. Would love to pair because I am unfamiliar with configuring an action 🌝


khiga8:
Exploring another idea

PR #556, opened 2021-05-07
ImageCrop: Don't set `rounded` attribute to `"false"`

koddsson:
`rounded` is a "boolean atrribute" which means that it can either be present or not. It doesn't take any values such as "true" or "false".

This PR makes sure to only set `rounded` if the `rounded` keyword is `true` and omit it from the element if not.


manuelpuyol (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main
 
+* Don't set boolean attribute to false
```
joelhawksley, line 5: Might we come up with something more descriptive here?

PR #558, opened 2021-05-07
Don't allow Octicon height or width below 16

jonrohan:
This change checks the `height` and `width` system arguments passed into `primer_octicon` that are less than 16px and coerces them to be 16.

Currently this effects 4 icons used in production. Will follow up with them to move them to appropriate sizes.


app/components/primer/octicon_component.rb
```diff
@@ -33,6 +33,13 @@ class OcticonComponent < Primer::Component
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     def initialize(icon_name = nil, icon: nil, size: SIZE_DEFAULT, use_symbol: false, **system_arguments)
       icon_key = icon_name || icon
+
+      # Don't allow sizes under 16px
+      if system_arguments[:height].present? && system_arguments[:height] < 16 || system_arguments[:width].present? && system_arguments[:width] < 16
```
manuelpuyol, line 38: we should call this out in the docs

manuelpuyol: also, extracting `16` to a constant like `MIN_SIZE` could be helpful for the future :)

app/components/primer/octicon_component.rb
```diff
@@ -33,6 +33,13 @@ class OcticonComponent < Primer::Component
...
+        system_arguments.delete(:height)
+        system_arguments.delete(:width)
```
manuelpuyol, line 40: I wonder if we should accept those arguments at all.
maybe we could only accept `size` and that'd be either one of the accepted symbols or an int >= 16

jonrohan: I think we should deprecated the use of these, but there's a lot of cleanup work to do on dotcom to prepare to only also `size:` there's a lot of violations that would break if we forced their hand. The most visible of which would be the github logo in the header.

app/components/primer/octicon_component.rb
jonrohan: It's currently a secret that the component accepts height and width. but it shouldn't. Unfortunately this will require a lot of refactoring on dotcom to adjust to the proper sizes. This was an attempt to start cutting down what's possible. 

It would be awesome to make this more of an allowlist type thing where we list the sizes currently being used in dotcom and start chipping away at what's possible for custom sizes.

manuelpuyol (approved):

PR #563, opened 2021-05-11
Accept `aria-current="true"` in tabbed components.

manuelpuyol:
We were enforcing `aria-current` to be `page` on selected tabs (if they are links). But that can cause some confusion with nested tabs. 

E.g., https://github.com/explore

image

In this case, only the `Explore` tab should have `aria-current=page` since it represents the current page the user is on. If the `Code` tabs (which are links and are selected) add `aria-current=page`, the screen reader would tell the user that they are on the `Code` page, which is incorrect. If they use `aria-current=true` it will only say "current item", avoiding confusion


khiga8 (approved):
lgtm!

PR #565, opened 2021-05-12
Fail PR if there is no CHANGELOG.md change.

### LINKED ISSUE #560: Add check to make sure that each change is accompanied with a CHANGELOG entry.
We should ensure that each of the changes that land on the main branch has a corresponding CHANGELOG entry. This will make sure that consumers of the library have a clear idea of what changes have been made to the library and how they can check that those changes don't negatively interact with their system.

We considered in slack to use something like [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) and generate the CHANGELOG automatically from the commits to the repository. Still, We decided that we didn't want to add the additional cognitive load on developers.

The exit criteria for this issue are that a developer should not be able to merge a PR without a CHANGELOG entry to the main branch. This can be implemented as a GitHub action that checks to see if any of the changed files in the PR are the CHANGELOG and fail the build if it is not.

koddsson:
Shows an error and fails CI if the CHANGELOG.md file is not changed in the PR.

Fixes #560


koddsson:
Here's how the error looks in the diff view:

image

manuelpuyol (approved):

PR #566, opened 2021-05-12
Only run linters and tests on pull requests

koddsson:
`main` is a protected branch that developers cannot push to directly. Therefore we don't need to run the test and linters on pushes. This change means that a pull request will run the tests and linters just once for each test or lint job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   markdown:


manuelpuyol:
huh, didn't we change that in https://github.com/primer/view_components/pull/548/files?

manuelpuyol:
cc @jonrohan

jonrohan (approved):
I suppose that's fine. One thing that I worry about is somehow through a merge main ci breaks. We could liming push builds to only the main branch

PR #571, opened 2021-05-12
Explicitly limit tag for `Label` component 

khiga8:
**What**
Explicitly limited tags for `Label` component and updated docs.

**Why**
Part of #491 

**Notes**
I chose these tags based on current usage in dotcom and am open to other recommendations.

### DIFF
--- a/app/components/primer/label_component.rb
+++ b/app/components/primer/label_component.rb
@@ -5,6 +5,9 @@ module Primer
   class LabelComponent < Primer::Component
     status :beta
 
+    DEFAULT_TAG = :span
+    TAG_OPTIONS = [DEFAULT_TAG, :summary, :a, :div].freeze
+
     SCHEME_MAPPINGS = {
       primary: "Label--primary",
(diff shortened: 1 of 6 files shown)


app/components/primer/label_component.rb
```diff
@@ -5,6 +5,9 @@ module Primer
   class LabelComponent < Primer::Component
     status :beta
 
+    DEFAULT_TAG = :span
+    TAG_OPTIONS = [DEFAULT_TAG, :summary, :a, :div].freeze
```
manuelpuyol, line 9: huh, why `summary` and `a`. Didn't we try to move off clickable labels?

khiga8: We moved off of clickable `State`, and replaced the one example in dotcom with a `Label` which seemed to be acceptable

https://github.com/primer/view_components/pull/432
https://github.com/github/github/pull/176322

manuelpuyol (approved):

PR #572, opened 2021-05-13
Add GitHub actions checks on documentation examples

khiga8:
**What**
This PR introduces a GitHub action which runs axe checks for every component `@example` we have in our documentation by using [@axe-core/cli](https://www.npmjs.com/package/@axe-core/cli). This will exit if any failures, and will also generate a report which can be accessed as an artifact. Failed violations can also be viewed in actions output. 

[0]: https://github.com/primer/view_components/runs/2571680932?check_suite_focus=true
[1]: https://github.com/github/primer/issues/148#issuecomment-833756481

**Why**
#543 introduced accessibility checks into our integration tests. This is nice because we don't have to wait for actions to run these checks and can run locally. This also allows for axe to run after UI changes are evoked in behavioral (javascript) components. 

The problem is, most of our components don't have integration tests/previews. [See discussion](https://github.com/primer/view_components/pull/543#issuecomment-833578402). On the otherhand, most of our components have examples which are shown to consumers in docs. This action will help catch issues in those examples and may help us integrate fixes into the component.


.github/workflows/docs-build.yml
```diff
@@ -36,3 +36,51 @@ jobs:
...
+        run: |
+          npm install -g @axe-core/cli
+          files=($(grep -r -l "Examples" docs/content/components))
+          declare -a needs_to_be_fixed=(
+            autocomplete 
+            autocompleteitem
+            dropdown
+            dropdownmenu
+            markdown
+            navigationtab
```
khiga8, line 58: Excluding the check for component pages that are failing violations at this starting point. This will fail build for all new new components with bad examples.

.github/workflows/docs-build.yml
```diff
@@ -36,3 +36,52 @@ jobs:
...
+            component="${filename%.*}"
+            if [[ " ${needs_to_be_fixed[@]} " =~ " ${component} " ]]
+            then
+              continue
+            else
+              doc_urls+=("localhost:8000/components/${component}")
+            fi
+          done
+          printf -v joined '%s, ' "${doc_urls[@]}"
+          axe $joined -i .frame-example --save docs-example-accessibility.json --disable color-contrast --exit
```
khiga8, line 82: Followup TODO:

Pull this out into a script. 
Need to look into how to run script in github actions.

.github/workflows/docs-build.yml
koddsson: Instead of building in escape hatches for some components, can we fix those issues in separate PRs and then return to this PR and merge it?

.github/workflows/docs-build.yml
koddsson: If we pull this into its own script, we can even re-write this in something like ruby or node since bash debugging and developing can become a bit unwieldy 😬

.github/workflows/docs-build.yml
khiga8: I agree 100% that we should make sure to address these. My concern is that it may take some time to look into each of these and meanwhile more components may get merged in. I was thinking we should get this action out the door, and take time to look into each of these carefully

khiga8:
**ON HOLD**

Exploring alternative after pair session with Joel

khiga8:
Hey @koddsson, thanks for the thoughtful review ✨

After pair sesh with @joelhawksley, we're thinking of pursuing an alternate path which involves:

1) Auto-generating view component previews from the examples in documentation. (call `bundle exec rake docs:build` )
2) An integration test that loops through these previews and calls `assert_accessible`. 

This path allows us to make use of the gems and systems put in place in #543 PR and means no bash gnarlyness.
Have a look #587 !

PR #574, opened 2021-05-13
"Encapsulate" JS behaviors

manuelpuyol:
This PR adds a `data-view-component` attribute to all JS components so we can target elements with that tag to add behaviors instead of adding them globally.


app/components/primer/image_crop.rb
```diff
@@ -31,6 +31,8 @@ def initialize(src:, rounded: true, **system_arguments)
       @system_arguments[:tag] = "image-crop"
       @system_arguments[:src] = src
       @system_arguments[:rounded] = true if rounded
+
+      add_js_data_attributes(@system_arguments)
```
koddsson, line 35: Not a huge fan of adding a requirement to developers that they have to remember to call this function in each behavioral component. Can we somehow do this automatically for behavioral components?

app/components/primer/clipboard_copy_component.ts
```diff
@@ -1,5 +1,7 @@
 import '@github/clipboard-copy-element'
 
+const CLIPBOARD_COPY_TIMER_DURATION = 2000
```
koddsson, line 3: We can just inline this variable since it's only used once.

app/components/primer/clipboard_copy_component.ts
```diff
@@ -18,9 +20,20 @@ function toggleCopyButton(button: HTMLElement) {
   toggleSVG(checkIcon)
 }
 
+const clipboardCopyElementTimers = new WeakMap<HTMLElement, number>()
+
 document.addEventListener('clipboard-copy', function ({target}) {
   if (!(target instanceof HTMLElement)) return
-  toggleCopyButton(target)
+  if (!target.hasAttribute('data-view-component')) return
```
koddsson, line 27: This is great 👍🏻

koddsson (approved):

app/components/primer/image_crop.rb
manuelpuyol: I'm also not a huge fan, but we don't have a clear distinction of which components are behavioral, maybe we could make **all** components add the `data-view-component` attribute. We could make this change to `BaseComponent` which would reflect on all of our components.

WDYT?

app/components/primer/clipboard_copy_component.ts
manuelpuyol: I only extracted it for clarity, since someone that is not very familiar with JS wouldn't have to guess what `2000` is in that function. But I can inline it if you really prefer :)

joelhawksley (approved):

PR #581, opened 2021-05-17
Restrict tags for `HiddenTextExpander` and `Truncate`

khiga8:
**What**
This PR restricts tags for `HiddenTextExpander` to `:span`, and also restricts tags for `Truncate` to `[:div, :span, :p]`. I chose these tags based on current usage in dotcom and what seemed reasonable. 

This PR also makes explicit the tag argument for `Markdown` so consumers know they can set the `tag` argument. I was unsure what the allowed tags for this component should be based on current usage of `article, `td` in dotcom. `test_renders_different_tags` also seemed to imply that this component can be relatively flexible so I did not add restrictions. Open to thoughts if you have any.
 
**Why**
Part of #491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@
 * Do not raise error if Primer CSS class name is passed to component if `PRIMER_WARNINGS_DISABLED` is set.
 
     *Joel Hawksley*
+### Breaking changes
+
+* Restrict allowed tags for `Truncate` and `HiddenTextExpander` components.
 
 ### Accessibility
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+### Breaking changes
+
+* Restrict allowed tags for `Truncate`, `Markdown`, and `HiddenTextExpander`.
+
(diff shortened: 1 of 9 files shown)


joelhawksley:
@manuelpuyol you wrote `test_renders_different_tags`. Did you intend to imply that any tag could be passed for `Markdown`?

manuelpuyol:
> @manuelpuyol you wrote `test_renders_different_tags`. Did you intend to imply that any tag could be passed for `Markdown`?

Looking at dotcom usage for `markdown-body` we also use it with `td` and `article`. The test was intended to guarantee that we could use those tags
I think that restricting that component to `:div, :td, :article` is fine!

manuelpuyol (approved):

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
 
     *Joel Hawksley*
 
+### Breaking changes
+
+* Restrict allowed tags for `Truncate`, `Markdown`, and `HiddenTextExpander`.
+
```
manuelpuyol, line 28: mind updating this changelog to be after `# main`?

PR #582, opened 2021-05-17
Do not raise error if `PRIMER_WARNINGS_DISABLED` is set.

joelhawksley:
We've seen some friction from this error message, so I've added the option to disable it via an ENV flag.

cc @phillmv

### DIFF
--- a/app/lib/primer/classify.rb
+++ b/app/lib/primer/classify.rb
@@ -123,13 +123,13 @@ module Primer
       def validated_class_names(classes)
         return if classes.blank?
 
-        if ENV["RAILS_ENV"] == "development"
+        if ENV["RAILS_ENV"] == "development" && !ENV["PRIMER_WARNINGS_DISABLED"]
           invalid_class_names =
             classes.split(" ").each_with_object([]) do |class_name, memo|
               memo << class_name if INVALID_CLASS_NAME_PREFIXES.any? { |prefix| class_name.start_with?(prefix) }
             end
 
-          raise ArgumentError, "Use System Arguments (https://primer.style/view-components/system-arguments) instead of Primer CSS class #{'name'.pluralize(invalid_class_names.length)} #{invalid_class_names.to_sentence}. This warning will not be raised in production." if invalid_class_names.any?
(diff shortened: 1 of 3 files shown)


app/lib/primer/classify.rb
```diff
@@ -123,13 +123,13 @@ def call(classes: "", style: nil, **args)
       def validated_class_names(classes)
         return if classes.blank?
 
-        if ENV["RAILS_ENV"] == "development"
+        if ENV["RAILS_ENV"] == "development" && !ENV["PRIMER_WARNINGS_DISABLED"]
```
manuelpuyol, line 126: could we use an engine config instead? Something similar to `silence_deprecations`

test/lib/classify_test.rb
```diff
@@ -626,6 +626,16 @@ def test_raises_error_when_passing_in_a_primer_css_class_name_in_development
     ENV["RAILS_ENV"] = "test"
   end
 
+  def test_raises_does_not_raise_error_when_passing_in_a_primer_css_class_name_in_development_and_flag_is_set
+    ENV["RAILS_ENV"] = "development"
+    ENV["PRIMER_WARNINGS_DISABLED"] = "1"
+
+    assert_generated_class("bg-blue text-center float-left ml-1", { classes: "bg-blue text-center float-left ml-1" })
+  ensure
+    ENV["PRIMER_WARNINGS_DISABLED"] = "0"
```
manuelpuyol, line 635: we'd need to unset this var, or set it to nil to actually reset
`!ENV["PRIMER_WARNINGS_DISABLED"] == !"0" == false`

app/lib/primer/classify.rb
joelhawksley: 🤔 we could, but I was hoping to make this an easy command line flag, IE: `PRIMER_WARNINGS_DISABLED=1 script/server`

phillmv: FWIW the use case here is:

"I have to fix a bug in dotcom, but I'm getting a deprecation error when looking at the page. I don't own the view, or the fifty places it's linked, so I guess I'll just go and comment out the error raise in `/vendor`" 😉.

PR #588, opened 2021-05-19
Set custom controller for previews

khiga8:
**What**
In order for our preview templates to have access to methods in `Primer::ViewHelper`, we must configure a custom preview controller which includes the helper.

[Configure preview controller docs](https://viewcomponent.org/guide/previews.html#configuring-preview-controller) 

**Why**
An upcoming PR will introduce preview templates that use view helpers. Without this change, an error will be thrown when trying to view the preview.

**Notes**
thanks @manuelpuyol, @joelhawksley for helping me with this!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,17 @@
 
 ## main
 
+### Misc
+
+* Configure previews controller to allow view helper usage in preview template.
+
+    *Kate Higa*
+
 ### Breaking changes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,10 +2,16 @@
 
 ## main
 
+* Configure previews controller to allow view helper usage in preview template.
```
manuelpuyol, line 5: could we add a `Misc` section with this?

manuelpuyol (approved):

PR #593, opened 2021-05-21
Enforce `AutoComplete` has associated label

### LINKED ISSUE #550: Enforce `Autocomplete` component has label
This component has been flagged by our axe checks as missing a label.

[Form <input> elements must have labels](https://dequeuniversity.com/rules/axe/4.1/label)

This component renders an `input`. In all our doc examples we don't show this component rendered with a label and we don't mention that it should have an associated label. [The auto-complete-element repo example][repo] shows a visible label. 

Along with improving docs, we should enforce that this component is always rendered with an associated label -- ideally an `<label>` element.

[repo]: https://github.github.com/auto-complete-element/examples/

khiga8:
Closes #550 

**What**
This enforces there is always an accessible label provided for `AutoComplete` component.  This provides support for either a visible or non-visible label.

This PR introduces an optional `label` slot for when a user wants a visible label. A user can also associate a visible element as a label using `aria-labelledby`.

If a user does not want the label to be visible, they should set an `aria-label` instead. For ease, we allow the user pass in a `:"aria-label"` to `AutoComplete` component rather than each nested slots. If set at parent level, we can forward this value to`<input>` and `<ul>`. 

**Why**
All form elements must have an accessible label so we should raise if label is missing.

[0]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

**Notes**
[w3 editable combobox aria 1.2](https://www.w3.org/TR/wai-aria-practices-1.2/examples/combobox/combobox-autocomplete-list.html) Though 1.2 is a draft, the combobox listbox implementation in 1.1 is flawed and so the one in 1.2 has been encouraged.
[auto-complete-element example](https://github.github.com/auto-complete-element/examples/)

thanks @BlakeWilliams for the help!


app/components/primer/auto_complete.rb
```diff
@@ -83,6 +129,8 @@ class AutoComplete < Primer::Component
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     def initialize(src:, id:, **system_arguments)
       @id = id
+      @aria_label = system_arguments[:"aria-label"] || system_arguments.dig(:aria, :label)
+      system_arguments.delete(:"aria-label") && system_arguments[:aria]&.delete(:label)
```
khiga8, line 133: We need to delete `aria-label` so it doesn't get passed into `<auto-complete>` element. This would violate [aria-allowed-attr][0] because aria attributes are only expected on certain elements and roles.

However, this rule is [excluded in Microsoft Accessibility Insights][1] which makes me wonder if we should also exclude it. 

Keeping this rule as I wait [on this][2]

[0]: https://dequeuniversity.com/rules/axe/4.1/aria-allowed-attr
[1]: https://dequeuniversity.com/rules/axe/4.1/aria-allowed-attr
[2]: https://github.com/github/primer/issues/148#issuecomment-845974804

khiga8, line 133: `<auto complete aria-label="blah">` will fail [aria-allowed-attr][0] so we need to delete it. 

Note: `<auto-complete>` is a custom web component element so I'm not entirely sure whether this is a false positive. However, does not seem *critical* to have this label so going to keep this change rather than disabling the rule entirely. 

[0]: https://dequeuniversity.com/rules/axe/4.1/aria-allowed-attr

khiga8:
NOTE: 

This includes breaking change which will raise production errors  and also involves some API changes such as a new slot. I'm not sure what the typical way to roll out something like this is.

app/components/primer/auto_complete.rb
```diff
@@ -1,88 +1,116 @@
...
       )
 
+      aria_label = system_arguments[:"aria-label"] || system_arguments.dig(:aria, :label) || @aria_label
+      system_arguments[:"aria-label"] = aria_label if aria_label.present?
+
       Primer::BaseComponent.new(**system_arguments)
     }
 
     # @example Default
-    #   <%= render(Primer::AutoComplete.new(src: "/users/search", id: "user-popup", position: :relative)) do |c| %>
```
khiga8, line 42: I changed these to fruit examples so we can interact with them in our auto-generated previews. Wonder if there's a way to have this work in the docs....

app/components/primer/auto_complete.rb
```diff
@@ -1,88 +1,116 @@
...
 
     DEFAULT_INPUT_TYPE = :text
     INPUT_TYPE_OPTIONS = [DEFAULT_INPUT_TYPE, :search].freeze
 
+    # Optionally render a visible label. See <%= link_to_accessibility %>
+    #
+    # @param for [Symbol] id of input
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    renders_one :label, lambda { |**system_arguments|
+      raise ArgumentError, "missing keyword: for" if !system_arguments.key?(:for) && !Rails.env.production?
```
khiga8, line 24: `for` is a reserved keyword so we cannot do the following:
```
|for:, **system_arguments|
   @input_id_for_label = for # error
```

However, `for:` can still be accessed like `system_arguments[:for]` so we can manually raise a keyword argument error instead.

app/components/primer/auto_complete.rb
```diff
@@ -1,88 +1,116 @@
...
+    }
+
     # Required input used to search for results
     #
     # @param type [Symbol] <%= one_of(Primer::AutoComplete::INPUT_TYPE_OPTIONS) %>
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     renders_one :input, lambda { |type: DEFAULT_INPUT_TYPE, classes: "form-control", **system_arguments|
       system_arguments[:tag] = :input
+      system_arguments[:role] = :combobox
+
```
khiga8, line 38: Priority order:

- `<label>` associated with this input id
- `aria-label` set to this slot
- `aria-label` passed to parent

app/components/primer/auto_complete.rb
```diff
@@ -1,88 +1,116 @@
...
+      system_arguments[:role] = :listbox
+      system_arguments[:id] = @list_id
       system_arguments[:classes] = class_names(
         "autocomplete-results",
         system_arguments[:classes]
       )
 
+      aria_label = system_arguments[:"aria-label"] || system_arguments.dig(:aria, :label) || @aria_label
+      system_arguments[:"aria-label"] = aria_label if aria_label.present?
+
```
khiga8, line 70: The order of precedence for labeling of `<ul role="listbox">` will be:

- Setting `aria-label` passed to this slot
- Setting `aria-label` passed to parent

For now, we won't raise if this listbox is missing a label (API is getting complex and does not seem *as critical*), but happy to revisit this.

CHANGELOG.md
```diff
@@ -40,6 +46,10 @@
 
 ### Breaking changes
 
+* Require a label or `aria-label` to be provided for `AutoComplete` component.
```
manuelpuyol, line 49: mind moving this up?

app/components/primer/auto_complete.rb
```diff
@@ -1,88 +1,115 @@
...
     # Required input used to search for results
     #
     # @param type [Symbol] <%= one_of(Primer::AutoComplete::INPUT_TYPE_OPTIONS) %>
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     renders_one :input, lambda { |type: DEFAULT_INPUT_TYPE, classes: "form-control", **system_arguments|
       system_arguments[:tag] = :input
+
+      aria_label = system_arguments[:"aria-label"] || system_arguments.dig(:aria, :label) || @aria_label
+      if @input_id_for_label.present? && @input_id_for_label == system_arguments[:id]
```
manuelpuyol, line 40: this empty if seems weird

app/components/primer/auto_complete.rb
```diff
@@ -1,88 +1,115 @@
...
+      elsif aria_label.present?
+        system_arguments[:"aria-label"] = aria_label
```
manuelpuyol, line 42: should we remove `system_arguments[:aria][:label]` if present to avoid any conflicts?

app/components/primer/auto_complete.rb
```diff
@@ -1,88 +1,115 @@
...
       system_arguments[:tag] = :ul
-      system_arguments[:id] = @id
+      system_arguments[:id] = @list_id
       system_arguments[:classes] = class_names(
         "autocomplete-results",
         system_arguments[:classes]
       )
 
+      aria_label = system_arguments[:"aria-label"] || system_arguments.dig(:aria, :label) || @aria_label
+      system_arguments[:"aria-label"] = aria_label if aria_label.present?
```
manuelpuyol, line 68: same comment about `[:aria][:label]`

app/components/primer/auto_complete.rb
```diff
@@ -1,88 +1,115 @@
...
     INPUT_TYPE_OPTIONS = [DEFAULT_INPUT_TYPE, :search].freeze
 
+    # Optionally render a visible label. See <%= link_to_accessibility %>
+    #
+    # @param for [Symbol] id of input
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    renders_one :label, lambda { |**system_arguments|
+      raise ArgumentError, "missing keyword: for" if !system_arguments.key?(:for) && !Rails.env.production?
+
+      @input_id_for_label = system_arguments[:for]
```
manuelpuyol, line 27: I think this may be relying on the order in which slots are called, I'd prefer if we could set an `input_id:` argument in the parent that will pass it directly to `label` and `input`

khiga8:
>  This provides support for either a visible or non-visible label.

On second thought, we probably very rarely would want the latter, and would *almost always* want a visible label. It might be acceptable to not have a visible label if there's absolutely enough context to help a sighted user (like there's a button adjacent to an input with a `Search` text).

In many places though, we're using `aria-label` on the input field which helps screen reader users, and we make a placeholder text function as a label for sighted users. This actually isn't great from an accessibility perspective because placeholders can disappear when a user types input and arne't actual labels.

Technically, setting `aria-label` on the input will pass automated checks but as much as possible, we should be encouraging explicit labels. I'm not sure if this would require some design input or what on all those places where we are missing a visible text label.. 🤔

khiga8:
This is finally ready for another review, sorry for the delay! @manuelpuyol 

This has been updated with `input` converted to a component. 
This also requires `input_id` to be passed into parent, instead of having it passed in individually to slots like before.

app/components/primer/auto_complete.rb
```diff
@@ -1,98 +1,150 @@
...
+        system_arguments[:"aria-label"] = aria_label
+        system_arguments[:aria]&.delete(:label)
+      end
+
+      Input.new(id: @input_id, **system_arguments)
     }
 
+    # This component is part of `Primer::AutoCompleteComponent` and should not be
+    # used as a standalone component.
+    class Input < Primer::Component
```
manuelpuyol, line 41: I'd prefer if this class was on the bottom, it gets hard to read having a class in the middle of another class

app/components/primer/auto_complete.rb
```diff
@@ -1,98 +1,150 @@
...
+      DEFAULT_TYPE = :text
+      TYPE_OPTIONS = [DEFAULT_TYPE, :search].freeze
+
+      # @param type [Symbol]
```
manuelpuyol, line 45: missing the possible types

app/components/primer/auto_complete.rb
```diff
@@ -1,98 +1,150 @@
...
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+      def initialize(type: DEFAULT_TYPE, classes: "form-control", **system_arguments)
```
manuelpuyol, line 47: is there a case where we **don't** want `form-control`?

app/components/primer/auto_complete.rb
```diff
@@ -1,98 +1,150 @@
...
+        @system_arguments = system_arguments
+        @system_arguments[:tag] = :input
+
+        @aria_label = system_arguments[:"aria-label"]
+        @aria_labelledby = system_arguments[:"aria-labelledby"] || system_arguments.dig(:aria, :labelledby)
```
manuelpuyol, line 52: should we add instructions on how/when to use this?

app/components/primer/auto_complete.rb
khiga8: Good call-- doesn't look like it!

I'll use the `class_names` method :)

app/components/primer/auto_complete.rb
```diff
@@ -1,100 +1,156 @@
...
-      system_arguments[:tag] = :input
-      system_arguments[:type] = fetch_or_fallback(INPUT_TYPE_OPTIONS, type, DEFAULT_INPUT_TYPE)
-      system_arguments[:classes] = classes
-      Primer::BaseComponent.new(**system_arguments)
+    renders_one :input, lambda { |**system_arguments|
+      aria_label = system_arguments[:"aria-label"] || system_arguments.dig(:aria, :label) || @aria_label
+      if aria_label.present?
+        system_arguments[:"aria-label"] = aria_label
+        system_arguments[:aria]&.delete(:label)
+      end
```
manuelpuyol, line 35: not for now, but I can see us having a helper:

```rb
def set_aria_label(system_arguments, aria_label = nil)
  aria_label = system_arguments[:"aria-label"] || system_arguments.dig(:aria, :label) || @aria_label
  system_arguments[:"aria-label"] = aria_label if aria_label.present?
  system_arguments[:aria]&.delete(:label)
end
```

manuelpuyol (approved):

PR #595, opened 2021-05-21
Add `force_system_arguments` options instead of relying on RAILS_ENV

manuelpuyol:
Adds a configuration that allows users to choose if not using System Arguments should raise an error or not. It will allow users to select which environments will raise an error.

### DIFF
--- a/app/lib/primer/classify.rb
+++ b/app/lib/primer/classify.rb
@@ -123,7 +123,7 @@ module Primer
       def validated_class_names(classes)
         return if classes.blank?
 
-        if ENV["RAILS_ENV"] == "development" && !ENV["PRIMER_WARNINGS_DISABLED"]
+        if force_system_arguments? && !ENV["PRIMER_WARNINGS_DISABLED"]
           invalid_class_names =
             classes.split(" ").each_with_object([]) do |class_name, memo|
               memo << class_name if INVALID_CLASS_NAME_PREFIXES.any? { |prefix| class_name.start_with?(prefix) }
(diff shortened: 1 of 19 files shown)


app/lib/primer/classify.rb
```diff
@@ -236,6 +236,10 @@ def extract_value(memo, key, val, breakpoint)
           memo[:classes] << "#{key.to_s.dasherize}#{breakpoint}-#{val.to_s.dasherize}"
         end
       end
+
+      def force_system_arguments?
+        Rails.application.config.primer_view_components.force_system_arguments
```
joelhawksley, line 241: Can we explicitly set this value to false by default? We wouldn't want this to accidentally be turned on in production.

lib/primer/view_components/engine.rb
```diff
@@ -13,7 +13,9 @@ class Engine < ::Rails::Engine
       ]
 
       config.primer_view_components = ActiveSupport::OrderedOptions.new
+
       config.primer_view_components.force_functional_colors = true
+      config.primer_view_components.force_system_arguments = true
```
BlakeWilliams, line 18: Would it make sense to keep the default value based on the environment?

Maybe `config.primer_view_components.force_system_arguments = !Rails.env.production?`

BlakeWilliams (approved):
LGTM!

lib/primer/view_components/engine.rb
manuelpuyol: I'll leave it false by default and let users decide

PR #501, opened 2021-04-28
Fix the conditional inclusion of ViewComponent::SlotableV2

laserlemon:
For some `ViewComponent::VERSION::STRING` values, the comparison is unexpectedly `true`:

```ruby
"2.31.0".to_f >= 2.28 # true
"2.29.0".to_f >= 2.28 # true
"2.28.0".to_f >= 2.28 # true
"2.27.0".to_f >= 2.28 # false
"2.19.0".to_f >= 2.28 # false
```

But what's this? 🙈 

```ruby
"2.3.0".to_f >= 2.28 # true
```

Because we're currently locked to ViewComponent major version 2 by the `primer_view_components.gemspec`, we can use `ViewComponent::VERSION::MINOR` for the comparison (which exists for all ViewComponent 2.X versions).

PR #598, opened 2021-05-21
fix conditional inclusion of ViewComponent::SlotableV2

manuelpuyol:
Continues https://github.com/primer/view_components/pull/501

### DIFF
--- a/app/components/primer/component.rb
+++ b/app/components/primer/component.rb
@@ -5,7 +5,7 @@ require "view_component/version"
 module Primer
   # @private
   class Component < ViewComponent::Base
-    include ViewComponent::SlotableV2 unless ViewComponent::VERSION::STRING.to_f >= 2.28
+    include ViewComponent::SlotableV2 unless ViewComponent::Base < ViewComponent::SlotableV2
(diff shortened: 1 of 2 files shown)


app/components/primer/component.rb
```diff
@@ -5,7 +5,7 @@
 module Primer
   # @private
   class Component < ViewComponent::Base
-    include ViewComponent::SlotableV2 unless ViewComponent::VERSION::STRING.to_f >= 2.28
+    include ViewComponent::SlotableV2 unless ViewComponent::Base < ViewComponent::SlotableV2
```
khiga8, line 8: **I think** this is including the deprecated slotable v2 API if the consumer's view component version still supports it but I'm not sure. Mind elaborating why we need to do this? Do any of our components use the deprecated `ViewComponent::SlotableV2`?

manuelpuyol: the idea is to only include it if `ViewComponent::Base` doesn't
we just don't want to include it twice to avoid bloating

khiga8 (approved):
LGTM!

PR #600, opened 2021-05-25
Layout component

manuelpuyol:
- Creates a new `Layout` component based on https://primer.style/css/components/layout
- Deprecates the old `LayoutComponent`


app/components/primer/layout.rb
```diff
@@ -0,0 +1,347 @@
+# frozen_string_literal: true
+
+module Primer
+  # Add a general description of component here
+  # Add additional usage considerations or best practices that may aid the user to use the component correctly.
+  # @accessibility Add any accessibility considerations
+  class Layout < Primer::Component
+    CONTAINER_DEFAULT = :full
+    CONTAINER_OPTIONS = [:full, :xl, :lg, :md].freeze
```
joelhawksley, line 9:
```suggestion
    CONTAINER_DEFAULT = :full
    CONTAINER_OPTIONS = [CONTAINER_DEFAULT, :xl, :lg, :md].freeze
```

app/components/primer/layout.rb
```diff
@@ -0,0 +1,347 @@
...
+      system_arguments[:tag] = :div
+      system_arguments[:classes] = class_names(
+        "Layout-sidebar",
+        system_arguments[:classes]
+      )
+
+      Primer::BaseComponent.new(**system_arguments)
+    }
+
+    # @example Default
```
joelhawksley, line 80: Might we add descriptions for these use cases?

app/components/primer/layout.rb
```diff
@@ -0,0 +1,347 @@
...
+    end
+
+    def render?
+      main.present? && sidebar.present?
+    end
+
+    private
+
+    def wrapper
+      if @container == :full
```
joelhawksley, line 310: Isn't this more akin to being `none`, than `full`?

app/components/primer/layout.rb
```diff
@@ -0,0 +1,347 @@
...
+    def initialize(
+      density: DENSITY_DEFAULT,
+      container: CONTAINER_DEFAULT,
+      gutter: GUTTER_DEFAULT,
+      flow_row_until: FLOW_ROW_UNTIL_DEFAULT,
+      sidebar_width: SIDEBAR_WIDTH_DEFAULT,
+      sidebar_placement: SIDEBAR_PLACEMENT_DEFAULT,
+      sidebar_flow_row_placement: SIDEBAR_FLOW_ROW_PLACEMENT_DEFAULT,
+      main_width: MAIN_WIDTH_DEFAULT,
+      divider: false,
```
joelhawksley, line 273: What if we called this `show_divider`? It might more clearly indicate it is a boolean.

app/components/primer/layout.rb
```diff
@@ -0,0 +1,347 @@
...
+    # @param flow_row_until [Symbol] When the `Layout` should change from a row flow into a column flow. <%= one_of(Primer::Layout::FLOW_ROW_UNTIL_OPTIONS) %>
+    # @param sidebar_width [Symbol] <%= one_of(Primer::Layout::SIDEBAR_WIDTH_OPTIONS) %>
+    # @param sidebar_placement [Symbol] <%= one_of(Primer::Layout::SIDEBAR_PLACEMENT_OPTIONS) %>
+    # @param sidebar_flow_row_placement [Symbol] Sidebar placement when `Layout` is flowing as row. <%= one_of(Primer::Layout::SIDEBAR_FLOW_ROW_PLACEMENT_OPTIONS) %>
+    # @param main_width [Symbol] <%= one_of(Primer::Layout::MAIN_WIDTH_OPTIONS) %>
+    # @param divider [Boolean] Wether or not to add a divider between `main` and `sidebar`.
+    # @param divider_flow_row_variant [Symbol] Variants for the divider when `Layout` is flowing as row. <%= one_of(Primer::Layout::DIVIDER_FLOW_ROW_VARIANT_OPTIONS) %>
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    def initialize(
+      density: DENSITY_DEFAULT,
```
joelhawksley, line 265: @yaili is `density` a common term in our system? Maybe this would be more conventionally named as `variant`?

app/components/primer/layout_component.rb
```diff
@@ -3,6 +3,8 @@
 module Primer
   # Use `Layout` to build a main/sidebar layout.
   class LayoutComponent < Primer::Component
+    status :deprecated
```
joelhawksley, line 6: What do you think about providing a deprecation message suggesting the new component, with a brief description of why it's preferred?

app/components/primer/layout.rb
```diff
@@ -0,0 +1,347 @@
...
+        system_arguments[:classes]
+      )
+
+      Primer::BaseComponent.new(**system_arguments)
+    }
+
+    # @example Default
+    #
+    #   <%= render(Primer::Layout.new) do |c| %>
+    #     <% c.main(border: true) { "Main" } %>
```
joelhawksley, line 83: What do you think about having an example with _only_ defaults? (so no border)

joelhawksley: (Or do we want `border` to default to `true`?)

app/components/primer/layout.rb
```diff
@@ -0,0 +1,347 @@
...
+      end
+
+      render Primer::BaseComponent.new(tag: :div, container: @container) do
+        yield
+      end
+    end
+
+    # The layout's main content.
+    class Main < Primer::Component
+      def initialize(width:, **system_arguments)
```
joelhawksley, line 322: Might we add docs for `width`?

joelhawksley: (maybe it's time to add a linter that would have caught this?)

demo/test/components/previews/primer/docs/layout_preview/default.html.erb
```diff
@@ -0,0 +1,5 @@
+
```
joelhawksley, line 1: Might we remove these empty lines in the previews?

demo/test/components/previews/primer/docs/layout_preview.rb
```diff
@@ -0,0 +1,29 @@
+module Primer
+  module Docs
+    class LayoutPreview < ViewComponent::Preview
+      def default; end
```
joelhawksley, line 4: Wouldn't it be nice if we didn't have to define these methods? (Like controllers...)

app/components/primer/layout.rb
yaili: What is it relating to? If it's relating to the amount of items and space between within a screen, density would be correct.

demo/test/components/previews/primer/docs/layout_preview/default.html.erb
manuelpuyol: we don't commit them anymore since https://github.com/primer/view_components/pull/603 :)

app/components/primer/layout.rb
manuelpuyol: I'm adding borders to the examples so it is easier to **see** the sidebar and main blocks. Otherwise it'd just be text in the page.

- with border

image

- without border

image

app/components/primer/layout.rb
manuelpuyol: changed it to `with_divider` in https://github.com/primer/view_components/pull/600/commits/d26e8558e6fcf34b97b099db53d3a30f56c7abc3

app/components/primer/layout.rb
manuelpuyol: `width` is internal here, the parent component sets the correct `width` according to the `main_width` argument.

app/components/primer/layout_component.rb
manuelpuyol: we can do that in a follow up PR (to all deprecated components)

app/components/primer/layout.rb
joelhawksley: 👍🏻 is there a way in yard to explicitly mark something as not documented?

app/components/primer/layout.rb
manuelpuyol: I think `full` makes sense given the doc: "If `full`, the Layout component is applied edge-to-edge"

PR #603, opened 2021-05-26
Don't commit auto-generated previews

khiga8:
**What**
This PR splits out the auto-generation of previews into a `docs:preview` task.
This also updates the `ci.yml` actions workflow so that previews are generated before tests are run.
This also modifies `.gitignore` to ensure that generated previews are not committed.

**Notes**
`script/test` will generate preview & run tests.


JoseInTheArena:
> Notes
With this change, individuals should remember to call bundle exec rake docs:preview before running tests locally.

Would it be challenging for that command to run before the tests automatically?

khiga8:
> > > Notes
> > > With this change, individuals should remember to call bundle exec rake docs:preview before running tests locally.
> > 
> > 
> > Would it be challenging for that command to run before the tests automatically?
> 
> I'm pretty sure we can invoke rake tasks manually in Ruby! something like `Rake::Task["docs:preview"].invoke` probably.
> We could also extract the running code from the Rake task into a separate file which can be included by our tests and run by them

I ended up adding a new `script/test`. What do you think?

manuelpuyol:
> @manuelpuyol I ended up adding a new `script/test`. What do you think?

I'd prefer not to have a separate script for testing 😬

khiga8:
> > @manuelpuyol I ended up adding a new `script/test`. What do you think?
> 
> I'd prefer not to have a separate script for testing 😬



> Rake::Task["docs:preview"].invoke

ahh, it didn't occur to me to build previews inside the test...I'll look into that, thanks!!

khiga8:
> > > @manuelpuyol I ended up adding a new `script/test`. What do you think?
> > 
> > 
> > I'd prefer not to have a separate script for testing 😬
> 
> > Rake::Task["docs:preview"].invoke
> 
> ahh, it didn't occur to me to build previews inside the test...I'll look into that, thanks!!

Actually, I'm not so sure about doing this... Building previews actually depends on having yard documentation built first (part of the `docs:build` task). It feels kind of odd to run both of these tasks from within a test.. 🤔

manuelpuyol:
When I try to run all the tests without the generated previews, `IntegrationDocExamplesAxeTest` fails with `uninitialized constant Primer::Docs`

That _may_ lead to some friction, so I think we could do one of:

1. Declaring an empty `Primer::Docs` module in so it does not raise
2. Check if module exists before running the test

In both cases it would be nice to have the test print a warning saying it is being skipped and explaining how to generate the previews.

.github/workflows/ci.yml
```diff
@@ -157,7 +157,9 @@ jobs:
         gem install bundler:2.2.9
         bundle config path vendor/bundle
         bundle update actionview activemodel activesupport railties
+        bundle exec rake docs:preview
         bundle exec rake
+        rm -rf demo/test/components/previews/primer/docs/
```
manuelpuyol, line 162: do we have to remove them? I think this is ephemeral 🤔

test/system/doc_examples_axe_test.rb
```diff
@@ -3,20 +3,22 @@
...
   NOT_STANDALONE = [:AutoCompleteItemPreview].freeze
 
   # Starting point violations to be addressed. Do not add to this list!
   STARTING_POINT_VIOLATIONS = [:MarkdownPreview, :AutoCompletePreview, :TabComponentPreview, :NavigationTabComponentPreview].freeze
 
   def test_accessibility_of_doc_examples
+    # Workaround to ensure that all component previews are loaded.
+    visit("/rails/view_components")
+
+    raise "You must generate previews with `bundle exec rake docs:preview` before running this test" unless defined? Primer::Docs
```
khiga8, line 16: I think it might be better to raise here instead of silently skipping in case something changes in  in the CI workflow where previews isn't generated.

.github/workflows/ci.yml
khiga8: hrmmm I actually wasn't sure about this but looking into it a bit more, it seems we don't need this line:

```
each GitHub-hosted runner is always a clean isolated virtual machine, and it is 
destroyed at the end of the job execution.
```

https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories

manuelpuyol (approved):
✨

Issue #463, opened 2021-04-19
Children of `role="menu"` required to have `menuitem`, `menuitemcheckbox`, `menuitemradio`


khiga8:
Our `DropdownMenu` component has an ARIA role of `menu`. An ARIA role of `menu` MUST have a child of `menuitem`, `menuitemcheckbox`, or `menuitemradio`. 

<img width="658" alt="Screenshot of wai-aria menu required owned element documentation" src="https://user-images.githubusercontent.com/16447748/115291941-35d79d80-a10a-11eb-98b5-125bde99e65d.png">

 [wai-aria menu](https://www.w3.org/TR/wai-aria/#menu)

However, the example shown in the docs does not have these required roles set on the child elements. 
If indeed, this `DropdownMenu` component should always have children elements with role of `menuitem`, `menuitemcheckbox`, or `menuitemradio`, we should make that clear in the docs. However, if this dropdown is meant to be general purpose and won't contain menu items, then we should remove `role="menu"` OR refer them to a more general `DetailsComponent`.

khiga8:
Additionally, the current markup of `Dropdown` component looks something like: 

```
<details-menu role="menu">
  <div>
    <ul>
      <li>
      <li>
      <li>
    </ul>
  </div>
</details-menu>
```

While a list seems appropriate semantically, the `role="menu"` on `<details-menu>` requires a direct child to be one of `menuitem`, `menuitemcheckbox`, and `menuitemradio`. See violation:

<img width="629" alt="Screen Shot 2021-04-20 at 4 14 02 PM" src="https://user-images.githubusercontent.com/16447748/115480801-1fac0900-a200-11eb-93c3-e2d894ec9f14.png">

When using the `<details-menu>` element, we should be sure to follow and advocate following the [recommended markup](https://github.com/github/details-menu-element#markup) listed in the repo.

khiga8:
I'm closing this because the `action-menu` work should take care of this when we've upstreamed it

PR #606, opened 2021-05-28
Dropdown API updates

manuelpuyol:
I'm preparing `Dropdown` component to be used in dotcom and later promoting it to beta. WIth that in mind, I renamed it removing the suffix.
I've also updated the `button` slot to be called `summary` and use `system_arguments` instead of relying in the `summary_classes` attribute.


joelhawksley (approved):

khiga8:
Could we also update the examples to have proper roles on the menu items? 

Related issue: https://github.com/primer/view_components/issues/463
Related w3 link: https://www.w3.org/TR/wai-aria/#menu

manuelpuyol:
> Could we also update the examples to have proper roles on the menu items?
> 
> Related issue: #463
> Related w3 link: https://www.w3.org/TR/wai-aria/#menu

@khiga8 done in https://github.com/primer/view_components/pull/606/commits/a7bb02c2c54b6e3c206ce8c2c4399423ad9fc979

PR #610, opened 2021-06-01
Raise an error if `class` is used as a system argument

manuelpuyol:
The error is only raised in non-prod environments and redirects the user towards `classes`.


app/components/primer/base_component.rb
```diff
@@ -148,6 +148,8 @@ def initialize(tag:, classes: nil, **system_arguments)
       @tag = tag
       @system_arguments = system_arguments
 
+      raise ArgumentError, "`class` is an invalid argument. You should use `classes` instead." if system_arguments.key?(:class) && !Rails.env.production?
```
joelhawksley, line 151:
```suggestion
      raise ArgumentError, "`class` is an invalid argument. Use `classes` instead." if system_arguments.key?(:class) && !Rails.env.production?
```

joelhawksley (approved):

PR #611, opened 2021-06-01
Allow customizing `TabContainer` when using `TabNav` and `UnderlineNav` components.

manuelpuyol:
There are some cases in dotcom where we add custom classes to the `tab-container` wrapper, so we should support customizing it.


koddsson (approved):
This will unblock me in some dotcom work I'm doing! ❤️

Issue #594, opened 2021-05-21
Primer::Classify updates


jonrohan:
# Current problems

The `Primer::Classify` module in its current state has worked and got us where we are today, but could stand a rethinking. 

Some of the shortcomings of how it's implemented are:

1. Current [classify objects](https://github.com/primer/view_components/blob/main/app/lib/primer/classify.rb) are messy. We have a ton of constants in the class that try and piece together a bunch of logic to output the correct class names. It's a bit dizzying to look at what it's all doing and understand.
1. Because of its constants we have to keep it up to date with primer/css manually or [risk generating classes that don't exist anymore](https://github.com/primer/view_components/issues/579). 
1. We can't make good suggestions in an automated way as to what arguments people should be using when [warning them not to use utility classes](https://github.com/primer/view_components/blob/main/app/lib/primer/classify.rb#L132).
1. All of this logic lookup is slow and requires extra caching.

# Solutions

I would like to propose the solution that we build the API in a rake task at publish time. This rake task could read the [output of @primer/css utility classes](https://unpkg.com/@primer/css@17.0.3/dist/stats/utilities.json) and build a hash (stored in json/yaml?) that would be a very fast lookup.

```yaml
- "display: :none": "d-none"
- "display: :flex": "d-flex"
- "mr: 1": "mr-1"
```

# Questions?

**Q: CSS class naming isn't always consistent.**

In some of the utility classes we have inconsistent naming that we had to build logic around in primer::classify. Example, `anim-pulse` and `hover-grow` both are considered animation classes, but aren't consistent. 

I propose nothing is off the table to rename. If something is inconsistent here, that points to a problem in primer/css that we should fix.

# What this unblocks

With a rewrite of primer::classify in this way, we could make these future ideas possible.

* We could validate arguments properly by suggesting the correct alternative.
* Assuring that the VC api matches what's possible

PR #612, opened 2021-06-01
Create rake task to build out classify utilities file

jonrohan:
Closes https://github.com/primer/view_components/issues/594

This PR begins work on https://github.com/primer/view_components/issues/594 paving a way to convert all our arguments over to this method. I would consider that issue closed when this merges, not because it's 100% but gives us a clear migration path for including all of what classify does.

To include new classes add the class prefix to the [SUPPORTED_KEYS Array in utilities.rake](https://github.com/primer/view_components/pull/612/files#diff-b06ff8dce079663cbf388633860d951ae726544317372b7a049e5cef5c85c602R10) then rerun `bundle exec rake utilities:build`. If the class key doesn't match the system argument, add an exception to the `REPLACEMENT_KEYS` hash in the same file.

Even though this PR isn't all utilities available yet, you can already see improvements to the system:

* **Benchmark speed** Caching the spacing keys seems no longer needed, I tested a bit and found that it didn't effect the benchmark score. [As you can see in the diff both cached and uncached calls to Classify went down in the benchmark test](https://github.com/primer/view_components/pull/612/files#diff-0452827697d47b2ba296f89d3c73f37986d0f811d82a41de195e339231333daf). I think this will improve further as we move more keys into the class. 📉 😱 
* **Missing arguments** [If you check the built docs file in the diff, there are arguments that weren't documented](https://github.com/primer/view_components/pull/612/files#diff-45b24b2ca9179785a1c4dbef3d694b5fd94784674d96d55b143ba208a740dc1b) but now are simply because they existed in primer/css but weren't supported here. 
* **Reverse classname search** You can now call classify with a string of classnames and get the corresponding system arguments. This is most useful for migration efforts and will allow us to write scripts to migrate dotcom faster to 100% view component coverage.


app/lib/primer/classify.rb
```diff
@@ -3,11 +3,20 @@
...
+    # Load the utilities.yml file.
+    # Disabling because we want to load symbols, strings, and integers from the .yml file
+    # rubocop:disable Security/YAMLLoad
+    UTILITIES = YAML.load(
+      File.read(
+        File.join(File.dirname(__FILE__), "./classify/utilities.yml")
+      )
+    ).freeze
+    # rubocop:enable Security/YAMLLoad
+
```
jonrohan, line 15: Does this seem like the best place to load the file?

app/lib/primer/classify/cache.rb
```diff
@@ -19,26 +19,6 @@ def clear!
...
-          preload(
-            keys: Primer::Classify::Spacing::AUTO_MAPPINGS.keys,
-            values: Primer::Classify::Spacing::AUTO_OPTIONS
-          )
-
-          preload(
-            keys: Primer::Classify::Spacing::RESPONSIVE_MAPPINGS.keys,
-            values: Primer::Classify::Spacing::RESPONSIVE_OPTIONS
-          )
-
```
jonrohan, line 41: I could add these back, but spent a little time trying and found that they didn't improve the bench tests.

test/benchmarks/bench_classify.rb
```diff
@@ -51,7 +51,7 @@ def bench_allocations_with_cache
     Primer::Classify::Cache.preload!
     Primer::Classify.call(**@values)
 
-    assert_allocations 35 do
+    assert_allocations 30 do
```
jonrohan, line 54: 😱 was not expecting that

test/lib/classify/utilities_test.rb
```diff
@@ -0,0 +1,40 @@
...
+      assert_equal "", Primer::Classify::Utilities.classname(:mr, :foo)
+    end
+  end
+
+  def test_classes_to_hash
+    assert_equal({ mr: 1, mb: 2, classes: "foo" }, Primer::Classify::Utilities.classes_to_hash("mr-1 mb-2 foo"))
+    assert_equal({ classes: "foo" }, Primer::Classify::Utilities.classes_to_hash("foo"))
+    assert_equal({ mx: :auto }, Primer::Classify::Utilities.classes_to_hash("mx-auto"))
+    assert_equal({ mr: [1, 2], classes: "baz bin" }, Primer::Classify::Utilities.classes_to_hash("mr-1 mr-sm-2 baz bin"))
+    assert_equal({ mr: [1, nil, 2], classes: "foo bar" }, Primer::Classify::Utilities.classes_to_hash("mr-1 mr-md-2 foo bar"))
```
jonrohan, line 38: This is my favorite bit 😎

docs/content/system-arguments.md
```diff
@@ -103,7 +103,7 @@ System arguments include most HTML attributes. For example:
 | Name | Type | Description |
 | :- | :- | :- |
 | `bottom` | Boolean | If `false`, sets `bottom: 0`. |
-| `float` | Symbol | One of `:left` and `:right`. |
+| `float` | Symbol | One of `:left`, `:right`, or `:none`. |
```
jonrohan, line 106: `float-none` was missing in the docs

docs/content/system-arguments.md
```diff
@@ -114,19 +114,19 @@ System arguments include most HTML attributes. For example:
...
 | :- | :- | :- |
 | `m` | Integer | Margin. One of `0`, `1`, `2`, `3`, `4`, `5`, `6`, or `:auto`. |
-| `mb` | Integer | Margin bottom. One of `-6`, `-5`, `-4`, `-3`, `-2`, `-1`, `0`, `1`, `2`, `3`, `4`, `5`, or `6`. |
-| `ml` | Integer | Margin left. One of `-6`, `-5`, `-4`, `-3`, `-2`, `-1`, `0`, `1`, `2`, `3`, `4`, `5`, or `6`. |
-| `mr` | Integer | Margin right. One of `-6`, `-5`, `-4`, `-3`, `-2`, `-1`, `0`, `1`, `2`, `3`, `4`, `5`, or `6`. |
-| `mt` | Integer | Margin top. One of `-6`, `-5`, `-4`, `-3`, `-2`, `-1`, `0`, `1`, `2`, `3`, `4`, `5`, or `6`. |
+| `mb` | Integer | Margin bottom. One of `0`, `1`, `-1`, `2`, `-2`, `3`, `-3`, `4`, `-4`, `5`, `-5`, `6`, `-6`, `7`, `-7`, `8`, `-8`, `9`, `-9`, `10`, `-10`, `11`, `-11`, `12`, or `-12`. |
+| `ml` | Integer | Margin left. One of `0`, `1`, `-1`, `2`, `-2`, `3`, `-3`, `4`, `-4`, `5`, `-5`, `6`, or `-6`. |
+| `mr` | Integer | Margin right. One of `0`, `1`, `-1`, `2`, `-2`, `3`, `-3`, `4`, `-4`, `5`, `-5`, `6`, or `-6`. |
+| `mt` | Integer | Margin top. One of `0`, `1`, `-1`, `2`, `-2`, `3`, `-3`, `4`, `-4`, `5`, `-5`, `6`, `-6`, `7`, `-7`, `8`, `-8`, `9`, `-9`, `10`, `-10`, `11`, `-11`, `12`, `-12`, or `:auto`. |
```
jonrohan, line 120: https://github.com/primer/view_components/pull/624 should improve this output

jonrohan (commented):
@joelhawksley @BlakeWilliams @manuelpuyol I believe I've reached a point where I'm ready for a review here. Here's some feedback I'm looking for:

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main
 
+* Replacing Classify::Spacing class with primer utilities class and generation
```
joelhawksley, line 5:
```suggestion
* Replace Classify::Spacing class with pre-generated mappings.
```

app/lib/primer/classify.rb
joelhawksley: 🤷🏻 works for me!

app/lib/primer/classify/cache.rb
joelhawksley: 🤔 it's curious to me that this didn't change the benchmark. I guess that means your new solution is just as fast as preloading? 🤯

app/lib/primer/classify/utilities.rb
```diff
@@ -0,0 +1,114 @@
...
+        private
+
+        def find_selector(selector)
+          Primer::Classify::UTILITIES.each do |k, v|
+            v.each do |vk, vv|
+              next unless vv.include?(selector)
+
+              return [k, vk, vv.index(selector)]
+            end
+          end
```
joelhawksley, line 85: What might some intermediate variable names be that could make this code easier to understand?

joelhawksley (commented):
👏🏻

app/lib/primer/classify/cache.rb
BlakeWilliams: I don't think we'll need to preload anything generated from YAML files since it should be implicitly cached, and won't impact our object allocation tests.

test/benchmarks/bench_classify.rb
BlakeWilliams: I think this will be one of the benefits of auto-generating a lot of our API. Since the values are already determined in the YAML file, it's implicitly caching them.

BlakeWilliams (commented):
This looks great overall! I'll 👍 Joel's suggestions.

Would it be possible to run a benchmark against the old and new implementations using [Benchmark.ips](https://github.com/evanphx/benchmark-ips)? It would be great to validate any performance improvements made here and celebrate them!

app/lib/primer/classify/utilities.rb
manuelpuyol, line 85: If I understand correctly, we are traversing the whole YAML file looking for the class and then returning the keys. Performance could be an issue here (not sure if we care about it though).

jonrohan: Yeah I suspect it would be slow, but it's not something we'll do outside of ci in prod?

jonrohan: Maybe I can guard that method with a Rails.env.production? check to return a {} just so we don't accidentally use it in production.

manuelpuyol (approved):

jonrohan:
With the consultation of @BlakeWilliams, I ran a bench test between the old classify and this new codepath. The results were **1.63333725** faster!

## Test paths run

```rb
require "benchmark/ips"

Benchmark.ips do |x|
  x.report("old classify") do
    Primer::ClassifyOld.call({
      float: :left,
      m: [0, nil, 1],
      p: 4,
      pr: 2,
      hide: :sm
    })
  end
  x.report("new classify") do
    Primer::Classify.call({
      float: :left,
      m: [0, nil, 1],
      p: 4,
      pr: 2,
      hide: :sm
    })
  end
end
```

## Results

```
        old classify     6.089k i/100ms
        new classify     9.616k i/100ms
        old classify     59.281k (± 3.7%) i/s -    298.361k in   5.040182s
        new classify     93.601k (± 2.5%) i/s -    471.184k in   5.037053s
```

PR #613, opened 2021-06-01
Update `Dropdown` API

manuelpuyol:
I'm preparing `Dropdown` component to be used in dotcom and later promoting it to beta. With that in mind, 

### Breaking changes

- Renamed `Dropdown` and `Dropdown::Menu` removing the suffix.
- Updated the `button` slot to be called `summary` and use `system_arguments` instead of relying in the `summary_classes` attribute.

### Dropdown::Menu

Before this update, we always built the menu using:

```html
<ul>
  <li class="dropdown-item">text</li>
...
```

but this isn't always how we use Dropdows, so I updated `Dropdown::Menu` to either render items using a `<ul><li><item_tag>` structure or just adding the tags inside `details-menu`


manuelpuyol:
@joelhawksley @khiga8 I made significant changes from https://github.com/primer/view_components/pull/606 so I'd love a new reviews 😄 
The main changes from that PR is the ability to customize how to render the `Menu` and its `Items`

app/components/primer/dropdown.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+
+module Primer
+  # `Dropdown` is a lightweight context menu for housing navigation and actions.
+  # They're great for instances where you don't need the full power (and code) of the select menu.
```
joelhawksley, line 5:
```suggestion
  # They're great for instances where you don't need the full power (and code) of the SelectMenu.
```

app/components/primer/dropdown.rb
```diff
@@ -0,0 +1,127 @@
...
+    # Required trigger for the dropdown. Has the same arguments as <%= link_to_component(Primer::ButtonComponent) %>,
+    # but it is locked as a `summary` tag.
+    renders_one :summary, lambda { |**system_arguments, &block|
+      @summary_arguments = system_arguments
+      @summary_arguments[:button] = true
+
+      view_context.capture { block&.call }
+    }
+
+    # Required context menu for the dropdown
```
joelhawksley, line 16:
```suggestion
    # Required context menu for the dropdown.
```

app/components/primer/dropdown.rb
```diff
@@ -0,0 +1,127 @@
...
+    # @param direction [Symbol] <%= one_of(Primer::Dropdown::Menu::DIRECTION_OPTIONS) %>
+    # @param scheme [Symbol] Pass `:dark` for dark mode theming
+    # @param header [String] Optional string to display as the header
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+    renders_one :menu, "Primer::Dropdown::Menu"
+
+    # @example Default
+    #   <div>
+    #     <%= render(Primer::Dropdown.new) do |c| %>
+    #       <% c.summary do %>
```
joelhawksley, line 28: In what cases do you think we should use `with_content` in our examples? I think it would be nice to use here.

app/components/primer/dropdown.rb
```diff
@@ -0,0 +1,127 @@
...
+    #         menu.item { "Item 1" }
+    #         menu.item { "Item 2" }
+    #         menu.item(divider: true)
+    #         menu.item { "Item 3" }
+    #         menu.item { "Item 4" }
+    #       end %>
+    #     <% end %>
+    #   </div>
+    #
+    # @example With Direction
```
joelhawksley, line 42:
```suggestion
    # @example With direction
```

app/components/primer/dropdown.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+
+module Primer
+  # `Dropdown` is a lightweight context menu for housing navigation and actions.
+  # They're great for instances where you don't need the full power (and code) of the select menu.
+  class Dropdown < Primer::Component
+    # Required trigger for the dropdown. Has the same arguments as <%= link_to_component(Primer::ButtonComponent) %>,
+    # but it is locked as a `summary` tag.
+    renders_one :summary, lambda { |**system_arguments, &block|
```
joelhawksley, line 9: What do you think about calling this `button`? `summary` refers to the underlying implementation, not the function.

app/components/primer/dropdown.rb
```diff
@@ -0,0 +1,127 @@
...
+    #         menu.item { "Item 2" }
+    #         menu.item(divider: true)
+    #         menu.item { "Item 3" }
+    #         menu.item { "Item 4" }
+    #       end %>
+    #     <% end %>
+    #   </div>
+    #
+    # @example Customizing menu items
+    #   <div>
```
joelhawksley, line 94: Is there any way to avoid these wrapping `div`s in these examples?

app/components/primer/dropdown.rb
```diff
@@ -0,0 +1,127 @@
...
+    #         menu.item(tag: :button) { "Item 2" }
+    #         menu.item(divider: true)
+    #         menu.item(classes: "custom-class") { "Item 3" }
+    #         menu.item { "Item 4" }
+    #       end %>
+    #     <% end %>
+    #   </div>
+    #
+    # @param overlay [Symbol] <%= one_of(Primer::DetailsComponent::OVERLAY_MAPPINGS.keys) %>
+    # @param reset [Boolean] Whether to hide the default caret on the button
```
joelhawksley, line 111: What if this was something like `show_caret`?

khiga8:
Could we also import the necessary javascript file? I am trying to test the component in previews but the javascript bundle is missing 😭

app/components/primer/dropdown.rb
manuelpuyol: I'm pretty sure we need them for the dropdown to render correctly, but I'll double check

app/components/primer/dropdown.rb
manuelpuyol: The problem of using `with_content` here is how slots work :/
Since we aren't using any arguments in this slot, the method acts as a getter, so it returns `nil`

joelhawksley: Ah yes, we are tracking a bug for that.

app/components/primer/dropdown/menu.rb
```diff
@@ -0,0 +1,99 @@
...
+
+        def call
+          component = if BUTTON_TAGS.include?(@system_arguments[:tag])
+                        Primer::ButtonComponent.new(scheme: :link, **@system_arguments)
+                      else
+                        Primer::BaseComponent.new(**@system_arguments)
+                      end
+
+          # divider has no content
+          render(component) if divider?
```
joelhawksley, line 84: What if there is a `divider` _and_ `content`? Do we raise? Is this documented?

app/components/primer/dropdown.rb
```diff
@@ -0,0 +1,134 @@
...
+      view_context.capture { block&.call }
+    }
+
+    # Required context menu for the dropdown.
+    #
+    # @param as [Symbol] When `as` is `:list`, wraps the menu in a `<ul>` with a `<li>` for each item.
+    # @param direction [Symbol] <%= one_of(Primer::Dropdown::Menu::DIRECTION_OPTIONS) %>
+    # @param scheme [Symbol] Pass `:dark` for dark mode theming
+    # @param header [String] Optional string to display as the header
+    # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
```
joelhawksley, line 22: Does it feel off to you that we write these comments here and not for the initializer on the slot class?

app/components/primer/dropdown/menu.rb
manuelpuyol: if someone uses
```rb
c.item(divider: true) { "some content" }
```

the content is ignored. I don't think we need to raise here

app/components/primer/dropdown.rb
manuelpuyol: it's just how we structure our docs. Adding the comments here will add a `### Slots` section and show the `menu` slot args in the same page as `Dropdown`.
We could restructure how we build it though

CHANGELOG.md
```diff
@@ -26,6 +40,7 @@
 
 ### Breaking changes
 
+
```
manuelpuyol, line 43:
```suggestion
```

app/components/primer/dropdown.rb
joelhawksley: 👍🏻 that's what I thought. I think this is fine for now! Was just curious.

app/components/primer/dropdown/menu.rb
joelhawksley: Do we need to document that behavior anywhere?

manuelpuyol: ah, that gets tricky since the items are menu's slots, I'm not sure how to handle that

app/components/primer/dropdown/menu.rb
```diff
@@ -0,0 +1,99 @@
...
+        TAG_OPTIONS = [TAG_DEFAULT, *BUTTON_TAGS].freeze
+
+        def initialize(as:, tag: TAG_DEFAULT, divider: false, **system_arguments)
+          @divider = divider
+          @as = as
+
+          @system_arguments = system_arguments
+          @system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, TAG_DEFAULT)
+          @system_arguments[:tag] = :li if list? && divider?
+          @system_arguments[:role] ||= :menuitem
```
khiga8, line 67: Is this `Dropdown` meant to be used with checkbox and radio buttons like in the `details-menu` repo example? Or maybe just links to start with?

manuelpuyol: Looking at current usages, I see some `role=menuitemradio` and `role=button` in existing dropdowns

khiga8: Ah okay, thus the `||= :menuitem`. Might be good to note this in the docs!

app/components/primer/dropdown/menu.rb
joelhawksley: Maybe in the param docs for `divider`?

app/components/primer/dropdown.rb
```diff
@@ -0,0 +1,134 @@
...
+    #   <% end %>
+    #
+    # @example Customizing menu items
+    #   <%= render(Primer::Dropdown.new) do |c| %>
+    #     <% c.button do %>
+    #       Dropdown
+    #     <% end %>
+    #
+    #     <%= c.menu(header: "Options") do |menu|
+    #       menu.item(tag: :summary) { "Item 1" }
```
khiga8, line 107: I think we might want to leave `summary` out as a menu item since it should rarely/never be used.

docs/content/components/dropdown.md
```diff
@@ -1,85 +1,164 @@
...
+<Example src="<details data-view-component='true' class='dropdown details-overlay details-reset'>  <summary role='button' data-view-component='true' class='btn'>            Dropdown      </summary>  <div data-view-component='true'>    <details-menu role='menu' data-view-component='true' class='dropdown-menu dropdown-menu-se'>    <div class='dropdown-header'>      Options    </div>      <a role='menuitem' data-view-component='true' class='dropdown-item'>Item 1</a>      <a role='menuitem' data-view-component='true' class='dropdown-item'>Item 2</a>      <a role='separator' data-view-component='true' class='dropdown-divider'></a>      <a role='menuitem' data-view-component='true' class='dropdown-item'>Item 3</a>      <a role='menuitem' data-view-component='true' class='dropdown-item'>Item 4</a></details-menu></div></details>" />
```
khiga8, line 22: It looks like there's an extra div being wrapping the `details-menu`. 
Previously I think we tested and saw that the extra div was causing issues with the js selection.
Might want to verify this!

app/components/primer/dropdown/menu.ts
```diff
@@ -0,0 +1 @@
+import '@github/details-menu-element'
```
khiga8, line 1: Hrmm I'm still unable to test the behavior in previews.. 🤔

docs/content/components/dropdown.md
manuelpuyol: Do you know what was the issue? docs/storybook/previews are working fine here. To remove that `div` we may need to do bigger changes to `DetailsComponent`

app/components/primer/dropdown/menu.ts
manuelpuyol: here they are looking fine 🤔

app/components/primer/dropdown/menu.rb
manuelpuyol: I added a `With dividers` section to the docs and added a description for it

docs/content/components/dropdown.md
khiga8: When I add an extra div to the [details-menu-repo examples](https://github.github.io/details-menu-element/examples/), it looks like the summary text isn't changing as expect based on the selection. I'm not sure how this will impact dropdown though.

manuelpuyol: I think that this is related to the `data-menu-button` prop, which we don't use in dropdowns
Given that we currently ship this component with the wrapping `div`, I'm inclined to not change that behavior in this PR and try to fix it in a follow-up. would that be ok with you?

khiga8 (approved):
Let's 🚢 it!

PR #622, opened 2021-06-04
Namespace non stable components

manuelpuyol:
As discussed, we're going to experiment using namespaces to indicate the component's status.
As a first experiment, I'm updating:

* `Primer::ButtonMarketingComponent` to `Primer::Alpha::ButtonMarketing`.
* `Primer::TextComponent` to `Primer::Beta::Text`.


app/components/primer/beta/text.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Primer
+  module Beta
+    # `Text` is a wrapper component that will apply typography styles to the text inside.
+    class Text < Primer::Component
+      status :beta
```
joelhawksley, line 7: I assume we'd eventually get rid of the `status` macro, right?

joelhawksley (approved):

app/components/primer/beta/text.rb
manuelpuyol: yeah :)

PR #624, opened 2021-06-05
Sort the component arguments in the docs

jonrohan:
This PR sorts the arguments found in the docs like https://primer.style/view-components/system-arguments. 

I believe this is useful because as we automate more of the arguments, they will show up sorted automatically in the docs. And sorted them will keep them consistent between pages.

https://github.com/primer/view_components/pull/612#discussion_r646073592


docs/content/components/blankslate.md
```diff
@@ -129,7 +129,7 @@ There are a few variations of how the Blankslate appears: `narrow` adds a maximu
 | `title` | `String` | `""` | Text that appears in a larger bold font. |
 | `title_tag` | `Symbol` | `:h3` | HTML tag to use for title. |
 | `icon` | `Symbol` | `""` | Octicon icon to use at top of component. |
-| `icon_size` | `Symbol` | `:medium` | One of `:small` (`16`) and `:medium` (`24`). |
+| `icon_size` | `Symbol` | `:medium` | One of `:medium` (`24`) and `:small` (`16`). |
```
joelhawksley, line 132: IMO it is nicer to have these from smallest to largest. What do you think?

jonrohan: Agreed, but that would require a lot more customized sorting comparator than here right? https://github.com/primer/view_components/pull/624/files#diff-b3a4f33411bc8a22b787fc10f0f06694f1487183b49e9add40bf17929a4c2209R7-R9

joelhawksley: Yeah, I don't know how we'd even write that. What if we added a `sort` option `one_of` and had it be false for just these cases?

lib/yard/docs_helper.rb
```diff
@@ -3,7 +3,13 @@
 module YARD
   # Helper methods to use for yard documentation
   module DocsHelper
-    def one_of(enumerable, lower: false)
+    def one_of(enumerable, lower: false, sort: false)
```
joelhawksley, line 6: What do you think about defaulting `sort` to `true`?

joelhawksley: Also, maybe we should write a test for this change?

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
 
     *Manuel Puyol*
 
+### Misc
+
+* Sort the docs arguments
```
joelhawksley, line 15:
```suggestion
* Sort documentation arguments.
```

joelhawksley (approved):

Issue #623, opened 2021-06-04
Improve linters


manuelpuyol:
Currently, our linters only count the number of offenses and work as a Pass/Fail.

It'd be nice for linters to suggest how to call the component to generate the same HTML it replaces. We can work first with component arguments and work our way through System Arguments.

manuelpuyol:
Autocorrecting linters are implemented now :)

PR #625, opened 2021-06-07
Linter suggestions

manuelpuyol:
Related to https://github.com/primer/view_components/issues/623

### Summary

This PR starts to work on linter suggestions, providing a better feedback for users on how to use our components.

| Before | After |
|--------|--------|
| image | image  |

### When does it work?

For now, the linter suggestion supports:

1. classes that the component generate (not System Arguments / custom classes)
2. `aria-` attributes
3. `data-` attributes
4. `test_selector` ERB block
5. `tag` attribute
6. `disabled` and `type` (for Button only)

### TODOs

1. I'm not looking at the content, so the "suggestion" is not auto-correctable since I still don't know how to look at the whole element block (any AST experts here? 😄 ).
2. Integrating with https://github.com/primer/view_components/pull/612
3. Dealing with slots

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@
 
     *Manuel Puyol*
 
+### Misc
+
+* Add linter suggestions for `Button` component.
+
+    *Manuel Puyol*
+
 ## 0.0.43
 
 * Upgrade primer/css to 17.2.1
--- a/lib/primer/view_components/linters/argument_mappers/button.rb
+++ b/lib/primer/view_components/linters/argument_mappers/button.rb
@@ -20,6 +20,8 @@ module ERBLint
           "btn-large" => ":large"
         }.freeze
 
+        TYPE_OPTIONS = %w[button reset submit].freeze
+
         def initialize(tag)
           @tag = tag
         end
@@ -36,6 +38,13 @@ module ERBLint
               args = args.merge(classes_to_args(attribute))
             elsif attr_name == "disabled"
               args[:disabled] = true
+            elsif attr_name == "type"
+              # button is the default type, so we don't need to do anything.
+              next if attribute.value == "button"
+
+              raise ConversionError, "Button component does not support type \"#{attribute.value}\"" unless TYPE_OPTIONS.include?(attribute.value)
+
+              args[:type] = ":#{attribute.value}"
             elsif attr_name.start_with?("aria-", "data-")
               # if attribute has no value_node, it means it is a boolean attribute.
               args["\"#{attr_name}\""] = attribute.value_node ? "\"#{attribute.value}\"" : true
--- a/lib/primer/view_components/linters/argument_mappers/button.rb
+++ b/lib/primer/view_components/linters/argument_mappers/button.rb
@@ -60,7 +60,7 @@ module ERBLint
         end
 
         def classes_to_args(classes)
-          hash = classes.value.split(" ").each_with_object({}) do |class_name, acc|
+          classes.value.split(" ").each_with_object({}) do |class_name, acc|
             next if class_name == "btn"
 
             if SCHEME_MAPPINGS[class_name] && acc[:scheme].nil?
@@ -75,8 +75,6 @@ module ERBLint
               raise ConversionError, "Cannot convert class \"#{class_name}\""
             end
           end
-
-          hash
         end
       end
     end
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 ### Misc
 
+* Add linter suggestions for `Button` component.
```
joelhawksley, line 15: This seems like a bigger deal than `misc` 😄

lib/primer/view_components/linters/argument_mappers/button.rb
```diff
@@ -0,0 +1,84 @@
...
+
+require_relative "conversion_error"
+require_relative "system_arguments"
+
+module ERBLint
+  module Linters
+    module ArgumentMappers
+      # Maps classes in a button element to arguments for the Button component.
+      class Button
+        SCHEME_MAPPINGS = {
```
joelhawksley, line 11: Do you think we could reuse the scheme mappings from the ButtonComponent? (I can also see why we'd rather not include the component for dependency reasons)

My concern would be about keeping these in sync.

lib/primer/view_components/linters/argument_mappers/button.rb
```diff
@@ -0,0 +1,84 @@
...
+            elsif class_name == "btn-block"
+              acc[:block] = true
+            elsif class_name == "BtnGroup-item"
+              acc[:group_item] = true
+            else
+              raise ConversionError, "Cannot convert class \"#{class_name}\""
+            end
+          end
+
+          hash
```
joelhawksley, line 79: Does this work?

```suggestion
          classes.value.split(" ").each_with_object({}) do |class_name, acc|
            next if class_name == "btn"

            if SCHEME_MAPPINGS[class_name] && acc[:scheme].nil?
              acc[:scheme] = SCHEME_MAPPINGS[class_name]
            elsif VARIANT_MAPPINGS[class_name] && acc[:variant].nil?
              acc[:variant] = VARIANT_MAPPINGS[class_name]
            elsif class_name == "btn-block"
              acc[:block] = true
            elsif class_name == "BtnGroup-item"
              acc[:group_item] = true
            else
              raise ConversionError, "Cannot convert class \"#{class_name}\""
            end
          end
```

lib/primer/view_components/linters/argument_mappers/conversion_error.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module ERBLint
+  module Linters
+    module ArgumentMappers
+      # Error when converting atguments.
```
joelhawksley, line 6:
```suggestion
      # Error when converting arguments.
```

lib/primer/view_components/linters/argument_mappers/system_arguments.rb
```diff
@@ -0,0 +1,46 @@
...
+        TEST_SELECTOR_REGEX = /test_selector\((?<selector>.+)\)$/.freeze
+
+        attr_reader :attribute
+        def initialize(attribute)
+          @attribute = attribute
+        end
+
+        def to_args
+          if attribute.erb?
+            _, _, code_node, = *attribute.node
```
joelhawksley, line 20:
```suggestion
            _, _, code_node = *attribute.node
```

lib/primer/view_components/linters/argument_mappers/system_arguments.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require_relative "conversion_error"
+
+module ERBLint
+  module Linters
+    module ArgumentMappers
+      # Maps classes in a button element to arguments for the Button component.
```
joelhawksley, line 8: Is this the right description?

lib/primer/view_components/linters/argument_mappers/button.rb
manuelpuyol: the problem is everything that we need to load to get access to those constants :/

joelhawksley (approved):
👏🏻 fantastic!

Do you think it's worth waiting on the system arguments mapping work from @jonrohan before we roll this out?

manuelpuyol:
> 👏🏻 fantastic!
> 
> Do you think it's worth waiting on the system arguments mapping work from @jonrohan before we roll this out?

I think this is really good for a PoC. Once that work ships I can integrate it too :)

PR #626, opened 2021-06-09
Move docs around

manuelpuyol:
Changes the information architecture of the Docs to follow the order:

1. Component description
2. Deprecation warning
3. Accessibility notes
4. Arguments
5. Slots
6. Examples

I feel like most users don't ever reach the `Arguments` session now and ask things that are in the docs, like "how do I use a different tag for the component?". So it is nice to have it in the top.

Extra:

Added a way to manually set defaults to params, which will help slots documentations and cases (like button) where a parameter is hidden in the code.

### DIFF
--- a/app/components/primer/button_component.rb
+++ b/app/components/primer/button_component.rb
@@ -78,8 +78,8 @@ module Primer
     #
     # @param scheme [Symbol] <%= one_of(Primer::ButtonComponent::SCHEME_OPTIONS) %>
     # @param variant [Symbol] <%= one_of(Primer::ButtonComponent::VARIANT_OPTIONS) %>
-    # @param tag [Symbol] <%= one_of(Primer::BaseButton::TAG_OPTIONS) %>
-    # @param type [Symbol] <%= one_of(Primer::BaseButton::TYPE_OPTIONS) %>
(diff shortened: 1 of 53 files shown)


app/components/primer/button_component.rb
```diff
@@ -78,8 +78,8 @@ class ButtonComponent < Primer::Component
     #
     # @param scheme [Symbol] <%= one_of(Primer::ButtonComponent::SCHEME_OPTIONS) %>
     # @param variant [Symbol] <%= one_of(Primer::ButtonComponent::VARIANT_OPTIONS) %>
-    # @param tag [Symbol] <%= one_of(Primer::BaseButton::TAG_OPTIONS) %>
-    # @param type [Symbol] <%= one_of(Primer::BaseButton::TYPE_OPTIONS) %>
+    # @param tag [Symbol] (Primer::BaseButton::DEFAULT_TAG) <%= one_of(Primer::BaseButton::TAG_OPTIONS) %>
```
joelhawksley, line 81: What do you think about evaluating these constants to get their value?

manuelpuyol: the `docs.rake` takes care of that 😉 
image

joelhawksley (approved):

PR #631, opened 2021-06-10
Retry ferrum tests up to 3 times

manuelpuyol:
Based on https://github.com/y-yagi/minitest-retry

This module allows us to rerun failed tests according to their error. In this case, every time a test fails with `Ferrum::TimeoutError`, it will retry up to 3 times. If it fails all 3 times, the test suite will fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## main
 
+### Misc
+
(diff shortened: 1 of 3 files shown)


khiga8:
Thank you for looking into this 🙏

khiga8 (approved):

PR #633, opened 2021-06-11
Button linter autocorrect

manuelpuyol:
Adds autocorrects for (simple) buttons! For this to work, I had to re-analyze the AST to find out where each element starts and ends. With that info, I was able to transform HTML blocks into ERB blocks

https://user-images.githubusercontent.com/11280312/121713209-3a815800-caa2-11eb-8635-95a8d4d4271b.mov


joelhawksley (approved):
🤯 nice work!

PR #634, opened 2021-06-11
 Add API for restricting system arguments with helpful suggestions 

joelhawksley:
### Problem

In pairing with @scotttesler we ended up with a `p: 4` on a BorderBox, an option we should not allow to be set on BorderBox. Only by reading the docs did we find the `padding` option that accomplished what we were trying to do.

### Proposed solution

Introduce a way to restrict which system arguments can be passed to a Primer ViewComponent, starting with padding arguments for `BorderBox`. I've set us up with a nice error message:

```console
ArgumentError: Found 1 violations:

The p system argument is not allowed here. Perhaps you could consider using :padding options of :default, :condensed, and :spacious?
```


app/components/primer/base_component.rb
```diff
@@ -152,6 +150,35 @@ def initialize(tag:, classes: nil, **system_arguments)
...
+      if (denylist = system_arguments[:system_arguments_denylist])
+        if Rails.application.config.primer_view_components.force_system_arguments && !ENV["PRIMER_WARNINGS_DISABLED"]
+          # Convert denylist from:
+          # { [:p, :pt] => "message" } to:
+          # { p: "message", pt: "message" }
+          unpacked_denylist =
+            denylist.each_with_object({}) do |(keys, value), memo|
+              keys.each { |key| memo[key] = value }
+
+              memo
```
manuelpuyol, line 162:
```suggestion
              keys.each { |key| memo[key] = value }
```

I'm pretty sure that you don't need to return `memo` in `each_with_object`

manuelpuyol (commented):
I'm not sure how I feel about having a `denylist`, I was thinking of actually having an allowlist instead, that way we can have more control on what actually is acceptable (which will allow us docummenting it)

app/components/primer/base_component.rb
```diff
@@ -152,6 +150,35 @@ def initialize(tag:, classes: nil, **system_arguments)
...
+              message << "\n The #{violation} system argument is not allowed here. #{unpacked_denylist[violation]}"
+            end
+
+            raise(ArgumentError, message)
+          end
+        end
+
+        # Remove :system_arguments_denylist key and any denied keys from system arguments
+        @system_arguments.except!(:system_arguments_denylist)
+        @system_arguments.except!(*denylist.keys.flatten)
```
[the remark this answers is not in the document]
joelhawksley: Totally. I could see us combining this functionality with other argument filtering.

joelhawksley:
@manuelpuyol given the conversation during our team meeting this morning, what direction do you think we should take this PR? Maybe we land it as-is as a stopgap?

manuelpuyol:
> @manuelpuyol given the conversation during our team meeting this morning, what direction do you think we should take this PR? Maybe we land it as-is as a stopgap while we consider better long-term options?

I think that's a good strategy :)

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Updates
 
+* Restrict usage of padding system arguments on BorderBox, recommending use of `padding` density instead.
```
manuelpuyol, line 7: mind moving this up once you sync with `main`?

app/components/primer/base_component.rb
```diff
@@ -152,6 +152,33 @@ def initialize(tag:, classes: nil, **system_arguments)
 
       raise ArgumentError, "`class` is an invalid argument. Use `classes` instead." if system_arguments.key?(:class) && !Rails.env.production?
 
+      if (denylist = system_arguments[:system_arguments_denylist])
+        if Rails.application.config.primer_view_components.force_system_arguments && !ENV["PRIMER_WARNINGS_DISABLED"]
```
manuelpuyol, line 156: I'd like to extract the config check to `Primer::Component`, since it could be reused in other components

```rb
def force_system_arguments?
  Rails.application.config.primer_view_components.force_system_arguments
end
```

app/components/primer/border_box_component.rb
```diff
@@ -111,6 +112,8 @@ def initialize(padding: DEFAULT_PADDING, **system_arguments)
         PADDING_MAPPINGS[fetch_or_fallback(PADDING_MAPPINGS.keys, padding, DEFAULT_PADDING)],
         system_arguments[:classes]
       )
+
+      @system_arguments[:system_arguments_denylist] = { [:p, :pt, :pb, :pr, :pl] => PADDING_SUGGESTION }
```
manuelpuyol, line 116: could we add a test for `BorderBox` using denied args?

manuelpuyol (approved):

PR #635, opened 2021-06-11
Codify tag restrictions explorations

khiga8:
**What**
This PR introduces a custom rubocop rule against the `[:tag] ||= ` syntax which is used to allow flexible tags, and instead recommends the developer to use `fetch_or_fallback` or have a fixed tag. See Notes below on why I went with linter (for now). 

**Notes**
When it comes to tags, our components and slots fall under one of the three:

1. ✅ Fixed tag and not accepting a `tag:` keyword

```
def initialize(**system_arguments)
  system_arguments[:tag] = :h1
end
```

2. ✅ Restricted set of tags and accepting a `tag:`

```
def initialize(tag: DEFAULT_TAG, **system_arguments)
   system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, DEFAULT_TAG)
end

```

3. ⛔️ No tag restrictions and having a default via `||=`

```
def initialize(**system_arguments)
  system_arguments[:tag] ||= :h2
end

```
(Slot scenario)
```
   system_arguments[:tag] ||= :div
}
```
We probably rarely (if ever) want the last `[:tag] ||=`.

At first, I approached this by brainstorming what we could do at an API level to restrict tags. However, I found there isn't really a straightforward API we can have to enforce this. One option I explored was requiring an allowlist of tags for every component that is named `TAG_ALLOWLIST`, and having a test in `component_test` that always ensures that variable is defined. However, having this allowlist feels awkward for our components that are never changed like in scenario 1 and also won't address tag issues in slots. It also doesn't mesh well with components like  `Button` which don't need a tag allowlist defined because it renders a `BaseButton`.

Maybe with more exploration, there is an API that allows us to be more restrictive with both slot and component arguments (and not just for tag).

But for now, I think restricting the `[:tag] ||=` syntax will achieve what we want for both components and slots.


rubocop/cop/primer/no_tag_memoize.rb
```diff
@@ -0,0 +1,47 @@
...
+            _
+          )
+        PATTERN
+
+        def on_block(node)
+          add_offense(node, message: INVALID_MESSAGE) if tag_memoized?(node)
+        end
+
+        def on_def(node)
+          add_offense(node, message: INVALID_MESSAGE) if tag_memoized?(node)
```
khiga8, line 42: TODO: flagging the line with the issue, instead of the entire def or block.

CONTRIBUTING.md
```diff
@@ -0,0 +1,58 @@
+# Contributing
```
manuelpuyol, line 1: should this file be here?

rubocop/cop/primer/no_tag_memoize.rb
```diff
@@ -0,0 +1,42 @@
...
+      # This cop ensures that tags are not set with ||=
+      #
+      # bad
+      # @system_arguments[:tag] ||= :h1
+      #
+      # good
+      # @system_arguments = fetch_or_fallback(TAG_OPTIONS, tag, DEFAULT_TAG)
+      #
+      # good
+      # @system_arguments = :h2
```
manuelpuyol, line 17:
```suggestion
      # @system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, DEFAULT_TAG)
      #
      # good
      # @system_arguments[:tag] = :h2
```

CONTRIBUTING.md
khiga8: 😱

docs/content/contributing.md
```diff
@@ -24,6 +24,27 @@ To declare a dependency on an `npm` package, pass `js` to the generator:
...
+system_arguments[:tag] = :h1
+```
+
+2) Having an allowed list of tags that are set by the consumer via the `tag:` argument using the
+`fetch_or_fallback` helper.
+
+```rb
+system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, DEFAULT_TAG)
+```
+
```
joelhawksley, line 47:
```suggestion
Components and slots should restrict which HTML tags they will render. For example, an `Image` component should never be rendered as an `<h1>`.

Consider which HTML tags make sense. Components typically fall under one of the following patterns:

1) Fixed tag that cannot be updated by the consumer:

```rb
system_arguments[:tag] = :h1
```

2) Allowed list of tags that are set with the `tag:` argument using the `fetch_or_fallback` helper:

```rb
system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, DEFAULT_TAG)
```

```

khiga8: Less wordy!

manuelpuyol (approved):

PR #640, opened 2021-06-15
Adding ability for utilities class to check if a class is a classify selector

jonrohan:
This adds a `supported_selector?` method to the Primer::Classify::Utilities class. The goal is to make the check for classnames that should be arguments more explicit.

Currently, this code 👇🏻 loosely checks prefixes and throws the error. After my PR, the class will check directly if the whole selector exists. This code still uses the slow `find_selector` method so I'm limiting use to non-production. I have some ideas to make it faster, but could come at a tradeoff.

```rb
INVALID_CLASS_NAME_PREFIXES =      (["bg-", "color-", "text-", "d-", "v-align-", "wb-", "box-shadow-"] + CONCAT_KEYS.map { |k| "#{k}-" }).freeze
```

### DIFF
--- a/test/lib/classify/utilities_test.rb
+++ b/test/lib/classify/utilities_test.rb
@@ -20,8 +20,8 @@ class PrimerClassifyUtilitiesTest < Minitest::Test
 
   def test_supported_selector_returns_nil_in_production
     Rails.stub(:env, ActiveSupport::StringInquirer.new("production")) do
-      refute Primer::Classify::Utilities.supported_selector?("m-1")
-      refute Primer::Classify::Utilities.supported_selector?("foo")
+      assert_nil Primer::Classify::Utilities.supported_selector?("m-1")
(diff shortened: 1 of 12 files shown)


test/lib/classify/utilities_test.rb
```diff
@@ -13,6 +13,18 @@ def test_supported_value
...
   end
 
+  def test_supported_selector
+    assert Primer::Classify::Utilities.supported_selector?("m-1")
+    refute Primer::Classify::Utilities.supported_selector?("foo")
+  end
+
+  def test_supported_selector_returns_nil_in_production
+    Rails.stub(:env, ActiveSupport::StringInquirer.new("production")) do
+      assert_nil Primer::Classify::Utilities.supported_selector?("m-1")
```
joelhawksley, line 23: Generally, Ruby predicate methods should always return a boolean. Perhaps we could update this to `return false` in the early return case?

joelhawksley (approved):

PR #649, opened 2021-06-16
move classify from app/lib to lib/

manuelpuyol:


lib/primer/classify/utilities.yml
```diff
@@ -18,8 +18,6 @@
   - anim-pulse
   :pulse_in:
   - anim-pulse-in
-  :hover_grow:
-  - anim-hover-grow
```
jonrohan, line 22: This shouldn't have been removed 🤔

jonrohan (approved):
Looks good 🤘🏻 Added a couple of commits

PR #650, opened 2021-06-16
Restrict avatar stack body and image crop tags

khiga8:
**What**
This PR restricts the `AvatarStack` body slot to `:div` and `span`.
This PR also restricts the `ImageCrop` spinner tag to `:div`.

Both these were determined based on tag usage in dotcom.

**Notes**

Part of #491

### DIFF
--- a/app/components/primer/avatar_stack_component.rb
+++ b/app/components/primer/avatar_stack_component.rb
@@ -10,6 +10,9 @@ module Primer
 
     DEFAULT_TAG = :div
     TAG_OPTIONS = [DEFAULT_TAG, :span].freeze
+
+    DEFAULT_BODY_TAG = :div
+    BODY_TAG_OPTIONS = TAG_OPTIONS = [DEFAULT_TAG, :span].freeze
     # Required list of stacked avatars.
     #
     # @param kwargs [Hash] The same arguments as <%= link_to_component(Primer::AvatarComponent) %>.
@@ -40,6 +43,8 @@ module Primer
     # @param align [Symbol] <%= one_of(Primer::AvatarStackComponent::ALIGN_OPTIONS) %>
     # @param tooltipped [Boolean] Whether to add a tooltip to the stack or not.
     # @param body_arguments [Hash] Parameters to add to the Body. If `tooltipped` is set, has the same arguments as <%= link_to_component(Primer::Tooltip) %>.
+    #   The default tag is <%= pretty_value(Primer::AvatarStackComponent::DEFAULT_BODY_TAG) %> but can be changed using `tag:` to 
+    #    <%= one_of(Primer::AvatarStackComponent::BODY_TAG_OPTIONS).downcase %>
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     def initialize(tag: DEFAULT_TAG, align: ALIGN_DEFAULT, tooltipped: false, body_arguments: {}, **system_arguments)
       @align = fetch_or_fallback(ALIGN_OPTIONS, align, ALIGN_DEFAULT)
@@ -47,7 +52,8 @@ module Primer
       @tooltipped = tooltipped
       @body_arguments = body_arguments
 
-      @body_arguments[:tag] ||= :div # rubocop:disable Primer/NoTagMemoize
+      body_tag = @body_arguments[:tag] || DEFAULT_BODY_TAG
+      @body_arguments[:tag] = fetch_or_fallback(BODY_TAG_OPTIONS, body_tag, DEFAULT_BODY_TAG)
       @body_arguments[:classes] = class_names(
         "AvatarStack-body",
         @body_arguments[:classes]
--- a/app/components/primer/avatar_stack_component.rb
+++ b/app/components/primer/avatar_stack_component.rb
@@ -43,8 +43,8 @@ module Primer
     # @param align [Symbol] <%= one_of(Primer::AvatarStackComponent::ALIGN_OPTIONS) %>
     # @param tooltipped [Boolean] Whether to add a tooltip to the stack or not.
     # @param body_arguments [Hash] Parameters to add to the Body. If `tooltipped` is set, has the same arguments as <%= link_to_component(Primer::Tooltip) %>.
-    #   The default tag is <%= pretty_value(Primer::AvatarStackComponent::DEFAULT_BODY_TAG) %> but can be changed using `tag:` to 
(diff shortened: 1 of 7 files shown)


app/components/primer/avatar_stack_component.rb
```diff
@@ -10,6 +10,9 @@ class AvatarStackComponent < Primer::Component
 
     DEFAULT_TAG = :div
     TAG_OPTIONS = [DEFAULT_TAG, :span].freeze
+
+    DEFAULT_BODY_TAG = :div
+    BODY_TAG_OPTIONS = TAG_OPTIONS = [DEFAULT_TAG, :span].freeze
```
joelhawksley, line 15: Should we mention this in the docs?

khiga8: Yeah, I think we should. `tag` isn't explicitly part of the arguments and I'm a little blocked by the constraints of how we generate documentation. I'll have to revisit this with fresh eyes on Monday :)

lib/tasks/docs.rake
```diff
@@ -178,8 +178,8 @@ namespace :docs do
             "default" => default_value,
             "description" => view_context.render(inline: tag.text)
           }
-
-          f.puts("| `#{tag.name}` | `#{tag.types.join(', ')}` | #{default_value} | #{view_context.render(inline: tag.text)} |")
+          
+          f.puts("| `#{tag.name}` | `#{tag.types.join(', ')}` | #{default_value} | #{view_context.render(inline: tag.text.squish)} |")
```
khiga8, line 182: When [multi-lining the tag][0], a newline gets added. This newline causes issues with table formatting:

[0]: https://www.rubydoc.info/gems/yard/file/docs/Tags.md#multi-line-tags

Arguments table with description cell being split into two as a result of a newline

Added `.squish` to get rid of newlines within tag text. 

Screen Shot 2021-06-21 at 12 59 43 PM

khiga8:
Mind another look? @joelhawksley :)

app/components/primer/avatar_stack_component.rb
```diff
@@ -40,14 +43,17 @@ class AvatarStackComponent < Primer::Component
     # @param align [Symbol] <%= one_of(Primer::AvatarStackComponent::ALIGN_OPTIONS) %>
     # @param tooltipped [Boolean] Whether to add a tooltip to the stack or not.
     # @param body_arguments [Hash] Parameters to add to the Body. If `tooltipped` is set, has the same arguments as <%= link_to_component(Primer::Tooltip) %>.
+    #   The default tag is <%= pretty_value(Primer::AvatarStackComponent::DEFAULT_BODY_TAG) %> but can be changed using `tag:`
+    #   to <%= one_of(Primer::AvatarStackComponent::BODY_TAG_OPTIONS).downcase %>
```
joelhawksley, line 47: What is the `downcase` for here?

app/components/primer/avatar_stack_component.rb
```diff
@@ -40,14 +43,17 @@ class AvatarStackComponent < Primer::Component
...
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     def initialize(tag: DEFAULT_TAG, align: ALIGN_DEFAULT, tooltipped: false, body_arguments: {}, **system_arguments)
       @align = fetch_or_fallback(ALIGN_OPTIONS, align, ALIGN_DEFAULT)
       @system_arguments = system_arguments
       @tooltipped = tooltipped
       @body_arguments = body_arguments
 
-      @body_arguments[:tag] ||= :div # rubocop:disable Primer/NoTagMemoize
+      body_tag = @body_arguments[:tag] || DEFAULT_BODY_TAG
```
joelhawksley, line 55: 🤔 Why do we need this line? Isn't fetch_or_fallback meant to handle this behavior?

app/components/primer/avatar_stack_component.rb
khiga8: `one_of` renders `One of _____` because we most often use use it on it's own.
In this case, we're combining it with another sentence so we'd want to downcase.

khiga8: Oh! actually I just noticed there's a `lower` argument on the `one_of` method 🙏. I'll use that :)

app/components/primer/avatar_stack_component.rb
khiga8: This is definitely a little awkward but necessary in this scenario with `body_arguments`. 

`fetch_or_fallback` only falls back in prod and will raises an error otherwise.

If `@body_arguments[:tag]` is nil, the `fetch_or_fallback` method will raise in an error and cause our development tests to fail with:

```
Expected one of: [:div, :span]
Got: nil
```

Alternatively, I considered setting the tag default within the hash like:

```
def initialize(...body_arguments: { tag: DEFAULT_BODY_TAG}, **system_arguments)
```

However, it seemed like the default would easily get lost when a user passes in their own `body_arguments` for any other attribute.

joelhawksley: @khiga8 ah, that's helpful context. Do you think this might be a sign that `body` should be a slot?

khiga8: I was wondering that too. @manuelpuyol, would you have some context of why `body_arguments` over having a `body` slot?

manuelpuyol: We chose to use arguments instead of slots here because customizing `body` was rare, so having a slot for the edge case would make it more complicated for eveyone using it:

```
render Primer::AvatarStack.new do |c|
  c.body do |b|
    b.avatar
    b.avatar
...
```

khiga8: ah, gotcha! It is unexpected to me that the avatar stack is composed with an extra element like:
```
<div class="AvatarStack AvatarStack--three-plus">
  <div class="AvatarStack-body"> 
     <img>
     <img>
 ```
vs

```
<div class="AvatarStack AvatarStack--three-plus">
   <img>
   <img>
``` 
It looks like there's some [historical context][0] and a css reason for why it's like this.

I see what you're saying @manuelpuyol that maybe the consumer doesn't need to know exactly how this component is composed. Doing a quick grep for`AvatarStack-body` in dotcom, I see several instances with a tooltip and `aria-label` set on the `body` which makes me think it *could* be worth exposing `body` as a slot.  

I think we should revisit this API question outside of this PR! What do you think of moving forward with this tag change for now? @joelhawksley 

[0]: https://primer.style/css/components/avatars#avatar-stack

joelhawksley (approved):

Issue #641, opened 2021-06-15
Extend tag support for `Truncate`


khiga8:
Currently `Truncate` only supports `div`, `span`, `p`. Ideally, it also supports `strong`.

See [slack discussion](https://github.slack.com/archives/CGJTTJ738/p1623427050297100?thread_ts=1623426315.295400&cid=CGJTTJ738)

PR #652, opened 2021-06-18
Update the `Truncate` component to accept `:strong` as a tag.

Artemisia-Absinthium:
# Changes
Added the `:strong` tag in the `Truncate` component tag options.
Added a test for the `:strong` tag and one other for the `:p` tag.

# Why
This change follows an internal slack discussion about adding strong as a tag option. (See issue https://github.com/primer/view_components/issues/641)

# Linked issues
Close https://github.com/primer/view_components/issues/641


test/components/truncate_test.rb
```diff
@@ -17,9 +17,22 @@ def test_renders_truncated_span
...
+    render_inline(Primer::Truncate.new(tag: :p)) { "content" }
+
+    assert_selector("p.css-truncate.css-truncate-overflow", text: "content")
+  end
+
+  def test_renders_truncated_strong
+    render_inline(Primer::Truncate.new(tag: :strong)) { "content" }
+
+    assert_selector("strong.css-truncate.css-truncate-overflow", text: "content")
+  end
```
manuelpuyol, line 30: what do you think of having a more generic test, so we don't have to keep adding a test every time we change a tag?

```rb
def test_renders_multiple_tags
  Primer::Truncate::TAG_OPTIONS.each do |tag|
    render_inline(Primer::Truncate.new(tag: tag)) { "content" }

    assert_selector("#{tag}.css-truncate.css-truncate-overflow", text: "content")
  end
end
```

Artemisia-Absinthium: Yes! I like it!

Still learning a lot with Ruby. I will change to that one :)

app/components/primer/truncate.rb
```diff
@@ -6,7 +6,7 @@ class Truncate < Primer::Component
     status :beta
 
     DEFAULT_TAG = :div
-    TAG_OPTIONS = [DEFAULT_TAG, :span, :p].freeze
+    TAG_OPTIONS = [DEFAULT_TAG, :span, :p, :strong].freeze
```
khiga8, line 9: While we're here, do you think we should consider other similar text elements like, `<b>` and `<i>`?

Artemisia-Absinthium: I was looking up the [phrasing content tags](https://html.spec.whatwg.org/#phrasing-content) and I feel there is a lot to add if we want to be exhaustive.
`<b>` and `<i>` for sure!

I feel `<em>`, `<samp>`, `<small>` and `<u>` can also be used with truncated text.

And `<bdi>` and `<bdo>` for bidirecitonal text?

I don't see anything bad about giving more semantic options to the user, but I don't know how we want to tackle that in the project in general?

khiga8: Good point. Let's leave it out for now :)

CHANGELOG.md
```diff
@@ -36,6 +36,10 @@
 
     *Kate Higa*
 
+* Update the `Truncate` component to accept `:strong` as a tag.
+
+    *Amélia Chavot*
+
```
manuelpuyol, line 42: I think this could be in the `Updates` section! :)

manuelpuyol (approved):

CHANGELOG.md
Artemisia-Absinthium: To the update section then!

Do we have some rules for where changes should be written down? Or do we do it by intuition like I did? ^^'

joelhawksley: @Artemisia-Absinthium good question! Would you be up for updating our contributing guidelines to mention these sections?

Artemisia-Absinthium: Yep! I will draft a new PR for that. :)

joelhawksley (approved):
❤️ 🧡 💛 💚 💙 💜

PR #653, opened 2021-06-22
Restrict allowed tags for TabNav and Tab

khiga8:
**What**
This PR removes the flexible tag syntax for `TabNav` and `Tab` and updates documentation to be more descriptive with what is expected by setting `with_panel`. 

**Notes**
I found the API for these components quite complex. It's hard to know what gets rendered, what each of the parts are, and what API attributes cause what change.

From what I understand, `TabNav` and `UnderlineNav` are both dependent on `Navigation::Tab`. If `with_panel` is set on these components, `Navigation::Tab` will be rendered as a button and may also render a panel. Also, if `with_panel` is set, `Primer::TabContainer` is rendered as the wrapper and this comes with associated javascript behavior that allow panel navigation. However, by default, `Navigation::Tab` simply renders as an anchor. For now, I'm updating the documentation to be more descriptive but I think we should explore if it's possible to simplify the API. 
  
Part of #491

### DIFF
--- a/app/components/primer/navigation/tab_component.rb
+++ b/app/components/primer/navigation/tab_component.rb
@@ -96,7 +96,7 @@ module Primer
 
         @system_arguments = system_arguments
 
-        if with_panel
+        if with_panel || @system_arguments[:tag] == :button
           @system_arguments[:tag] = :button
           @system_arguments[:type] = :button
           @system_arguments[:role] = :tab
--- a/app/components/primer/navigation/tab_component.rb
+++ b/app/components/primer/navigation/tab_component.rb
@@ -18,7 +18,7 @@ module Primer
       #
       # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
       renders_one :panel, lambda { |**system_arguments|
-        system_arguments[:tag] ||= :div # rubocop:disable Primer/NoTagMemoize
+        system_arguments[:tag] = :div
         system_arguments[:role] ||= :tabpanel
         system_arguments[:hidden] = true unless @selected
 
@@ -97,11 +97,11 @@ module Primer
         @system_arguments = system_arguments
 
         if with_panel
-          @system_arguments[:tag] ||= :button # rubocop:disable Primer/NoTagMemoize
+          @system_arguments[:tag] = :button
           @system_arguments[:type] = :button
           @system_arguments[:role] = :tab
         else
-          @system_arguments[:tag] ||= :a # rubocop:disable Primer/NoTagMemoize
+          @system_arguments[:tag] = :a
         end
 
         @wrapper_arguments = wrapper_arguments
--- a/app/components/primer/tab_nav_component.rb
+++ b/app/components/primer/tab_nav_component.rb
@@ -130,7 +130,7 @@ module Primer
       @body_arguments = body_arguments
       @wrapper_arguments = wrapper_arguments
 
-      @system_arguments[:tag] ||= :div # rubocop:disable Primer/NoTagMemoize
+      @system_arguments[:tag] = :div
       @system_arguments[:classes] = class_names(
         "tabnav",
         system_arguments[:classes]
--- a/app/components/primer/tab_nav_component.rb
+++ b/app/components/primer/tab_nav_component.rb
@@ -10,7 +10,8 @@ module Primer
     DEFAULT_EXTRA_ALIGN = :left
     EXTRA_ALIGN_OPTIONS = [DEFAULT_EXTRA_ALIGN, :right].freeze
 
-    # Tabs to be rendered. For more information, refer to <%= link_to_component(Primer::Navigation::TabComponent) %>.
+    # Tabs to be rendered.  Set `with_panel` on the parent to render these as a button for panel navigation. Otherwise,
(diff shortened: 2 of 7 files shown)


app/components/primer/navigation/tab_component.rb
```diff
@@ -97,11 +98,11 @@ def initialize(list: false, selected: false, with_panel: false, icon_classes: ""
         @system_arguments = system_arguments
 
         if with_panel
-          @system_arguments[:tag] ||= :button # rubocop:disable Primer/NoTagMemoize
+          @system_arguments[:tag] = :button
```
khiga8, line 101: TODO: verify whether we should allow `button` to be rendered without `with_panel` being set [like done here][0], or whether this linked case should actually be using `with_panel`.


[0]: https://github.com/github/github/blob/master/app/components/discussions/templates/edit_form_component.html.erb#L28

khiga8: Setting `with_panel` will render `tab-container` which comes with it's own keyboard behavior which conflicts with the one specified [here](https://github.com/github/github/blob/master/app/assets/modules/github/behaviors/commenting/preview.ts#L50). This appears to be why `with_panel` is not being used. 

[here]: https://github.com/github/github/blob/master/app/assets/modules/github/behaviors/commenting/preview.ts#L50

app/components/primer/tab_nav_component.rb
```diff
@@ -119,7 +120,8 @@ class TabNavComponent < Primer::Component
     #   <% end %>
     #
     # @param label [String] Used to set the `aria-label` on the top level `<nav>` element.
-    # @param with_panel [Boolean] Whether the TabNav should navigate through pages or panels.
+    # @param with_panel [Boolean] Whether the TabNav should navigate through pages or panels. When true, <%= link_to_component(Primer::TabContainerComponent) %>
+    #   which comes with JavaScript behavior.
```
khiga8, line 124: From the docs alone, it's currently hard to know that setting `with_panel` comes with associated JS behavior. Took me a bit to trace down in the code so making docs more descriptive.

app/components/primer/navigation/tab_component.rb
khiga8: ANSWER: `with_panel` renders `tab-container` which comes with it's own keyboard behavior which conflicts with the behaviors specified [here](https://github.com/github/github/blob/master/app/assets/modules/github/behaviors/commenting/preview.ts#L50). Therefore, it looks like we want to support tag manually being set to `:button` outside of `with_panel`. 

[here]: https://github.com/github/github/blob/master/app/assets/modules/github/behaviors/commenting/preview.ts#L50

app/components/primer/navigation/tab_component.rb
```diff
@@ -18,7 +18,7 @@ class TabComponent < Primer::Component
       #
       # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
       renders_one :panel, lambda { |**system_arguments|
-        system_arguments[:tag] ||= :div # rubocop:disable Primer/NoTagMemoize
+        system_arguments[:tag] = :div
```
khiga8, line 21: Doesn't look like we ever render panel as a non-div element.

app/components/primer/navigation/tab_component.rb
```diff
@@ -96,12 +96,12 @@ def initialize(list: false, selected: false, with_panel: false, icon_classes: ""
...
 
-        if with_panel
-          @system_arguments[:tag] ||= :button # rubocop:disable Primer/NoTagMemoize
+        if with_panel || @system_arguments[:tag] == :button
+          @system_arguments[:tag] = :button
           @system_arguments[:type] = :button
           @system_arguments[:role] = :tab
         else
-          @system_arguments[:tag] ||= :a # rubocop:disable Primer/NoTagMemoize
+          @system_arguments[:tag] = :a
```
khiga8, line 104: Anchor and button seem to be the only valid options.

app/components/primer/navigation/tab_component.rb
manuelpuyol: I think the "correct" solution would be changing that case to stop relying in custom JS and start using `TabContainer`

app/components/primer/underline_nav_component.rb
```diff
@@ -135,7 +136,8 @@ class UnderlineNavComponent < Primer::Component
     #   <% end %>
     #
     # @param label [String] The `aria-label` on top level `<nav>` element.
-    # @param with_panel [Boolean] Whether the TabNav should navigate through pages or panels.
+    # @param with_panel [Boolean] Whether the `TabNav` should navigate through pages or panels. When true, <%= link_to_component(Primer::TabContainerComponent) %> is
```
manuelpuyol, line 139:
```suggestion
    # @param with_panel [Boolean] Whether the `UnderlineNav` should navigate through pages or panels. When true, <%= link_to_component(Primer::TabContainerComponent) %> is
```

app/components/primer/navigation/tab_component.rb
khiga8: Agreed! I'll look into how "do-able" that is!

manuelpuyol (approved):

Issue #654, opened 2021-06-22
Correctly render boolean attributes


manuelpuyol:
### Problem

We currently guide our users to use `true` for boolean attributes.

E.g:

```rb
Primer::ButtonComponent.new("data-disable-with": true)
```

Since we use `content_tag` as our base, when passing `true` to an attribute, it will render it as the attribute value. In the example above, it will render:

```html
<button data-disable-with="true">
```

Which is different from the expected:

```html
<button data-disable-with>
```

### Solution?

We could either guide users to use `""` instead of `true`, which should render correctly. Otherwise, we could treat the `trues` ourselves in `BaseComponent`.

jonrohan:
> We could either guide users to use `""`

I think a problem with this approach is if someone has some code like `Component.new(hidden: user_can_see_component_helper?)`

khiga8:
For the particular example you have above, I think that's expected behavior since `data-disable-with` should be set to a string.

manuelpuyol:
> > We could either guide users to use `""`
> 
> I think a problem with this approach is if someone has some code like `Component.new(hidden: user_can_see_component_helper?)`

I'm actually leaning towards only adding guidance to users. If an user has something like `Component.new(data-attr: some_boolean_check?)`, we can't know if they are depending on having the `"true"` value or not. Changing how this works could be a massive breaking change.

PR #655, opened 2021-06-22
Update linter autocorrection to use `""` instead of `true` for boolean attributes

manuelpuyol:
While we don't fix https://github.com/primer/view_components/issues/654 it is better for the autocorrection to not use invalid values


koddsson (approved):
Looks good to me ✨

PR #657, opened 2021-06-22
Restrict tags for `Popover`

khiga8:
**What**
This PR restricts allowed tags for `Popover`.

**Notes**
Part of #491

### DIFF
--- a/app/components/primer/tooltip.rb
+++ b/app/components/primer/tooltip.rb
@@ -10,6 +10,9 @@ module Primer
     MULTILINE_DEFAULT = false
     DELAY_DEFAULT = false
 
+    TAG_DEFAULT = :span
+    TAG_OPTIONS = [TAG_DEFAULT, :div].freeze
+
     ALIGN_MAPPING = {
       ALIGN_DEFAULT => "",
       :left_1 => "tooltipped-align-left-1",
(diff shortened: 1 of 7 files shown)


app/components/primer/tooltip.rb
```diff
@@ -70,7 +75,7 @@ def initialize(
       **system_arguments
     )
       @system_arguments = system_arguments
-      @system_arguments[:tag] ||= :span # rubocop:disable Primer/NoTagMemoize
+      @system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, TAG_DEFAULT)
```
khiga8, line 78: @manuelpuyol 

On dotcom I see a lot of something like:
`<button class="btn disabled tooltipped tooltipped-n"`  (and a whole bunch of other elements).

Is [the wrapping syntax](https://primer.style/view-components/components/tooltip#wrapping-another-component) a direct replacement?

manuelpuyol: I think this would be a case-by-case scenario, since the wrapped element would decide the `tag` for the tooltip wouldn't it?

khiga8:
Gonna leave `Tooltip` out of this PR, based on ambiguity and #663

khiga8:
Mind another look over? :) @manuelpuyol

joelhawksley (approved):

PR #666, opened 2021-06-25
Remove starting point violation list from `test/system/doc_examples_axe_test.rb`

khiga8:
After working on https://github.com/primer/view_components/pull/653, I realized that two of the components we currently have listed under `STARTING_POINT_VIOLATIONS` actually belong in `NOT_STANDALONE`. These are the components that don't make sense to run axe-core checks on since they require being examined as part of a larger component.

I also decided to fix some issues in the `Markdown` example so we can eliminate the `STARTING_POINT_VIOLATIONS` list entirely.


app/components/primer/markdown.rb
```diff
@@ -241,9 +241,9 @@ class Markdown < Primer::Component
     #
     #     <pre><code>var foo = "bar";</code></pre>
     #
-    #     <pre><code>Long, single-line code blocks should not wrap. They should horizontally scroll if they are too long. This line should be long enough to demonstrate this.</code></pre>
+    #     <pre tabindex="0"><code>Long, single-line code blocks should not wrap. They should horizontally scroll if they are too long. This line should be long enough to demonstrate this.</code></pre>
     #
-    #     <pre><code>var foo = "The same thing is true for code with syntax highlighting. A single line of code should horizontally scroll if it is really long.";</code></pre>
+    #     <pre tabindex="0"><code>var foo = "The same thing is true for code with syntax highlighting. A single line of code should horizontally scroll if it is really long.";</code></pre>
```
khiga8, line 246: axe was flagging this as violating the `scrollable-region-focusable` rule.

This tabindex ensures that keyboard uses can access this scrollable region and use right/left arrow keys to scroll.

https://dequeuniversity.com/rules/axe/4.1/scrollable-region-focusable

joelhawksley (approved):

PR #668, opened 2021-06-28
Always generate navigation links within list

khiga8:
## What 
This PR makes it so we always render links in `TabNav` and `UnderlineNav` within a `<ul>`. 
For instance, the markup for `TabNav` for page navigation which currently looks like:

```.html
<div data-view-component="true" class="tabnav">
  <nav aria-label="Default" data-view-component="true" class="tabnav-tabs">          
    <a href="#" aria-current="page" data-view-component="true" class="tabnav-tab">Tab 1</a>          
    <a href="#" data-view-component="true" class="tabnav-tab">Tab 2</a>          
    <a href="#" data-view-component="true" class="tabnav-tab">Tab 3</a>
  </nav>   
</div>
```

will be updated to:

```.html
<nav aria-label="Default" data-view-component="true" class="tabnav">
  <ul data-view-component="true" class="tabnav-tabs">
    <li data-view-component="true" class="d-flex">  
      <a href="#" aria-current="page" data-view-component="true" class="tabnav-tab">Tab 1</a>
    </li>
    <li data-view-component="true" class="d-flex">
      <a href="#" data-view-component="true" class="tabnav-tab">Tab 2</a>
    </li>
    <li data-view-component="true" class="d-flex">  
      <a href="#" data-view-component="true" class="tabnav-tab"> Tab 3</a>
    </li>
  </ul>    
</nav>
```

When panel navigation is set with `with_panel`, the `li` element will have [role="presentation"](https://www.w3.org/TR/wai-aria-practices/#presentation_role).

```.html
<div data-view-component="true" class="tabnav">
  <ul aria-label="With panels" role="tablist" data-view-component="true" class="tabnav-tabs">
    <li role="presentation" data-view-component="true" class="d-inline-flex">
      <button type="button" role="tab" aria-selected="true" data-view-component="true" class="tabnav-tab" tabindex="0"> <span data-view-component="true">Tab 1</span></button>
    </li>        
    <li role="presentation" data-view-component="true" class="d-inline-flex">
      <button type="button" role="tab" data-view-component="true" class="tabnav-tab" aria-selected="false" tabindex="-1"><span data-view-component="true">Tab 2</span> </button>
    </li>
  </ul>
</div>
```

## Why

- It'd help to simplify the API  if we always rendered the link items within a list rather than having it optional with `list: true`
- There are semantic benefits to rendering these as a list such as allowing the screen reader to announce the number of items available.
- Rendering the links within a list will help to provide semantic structure in places where it's not appropriate to be relying on a `nav` for rendering page links. For instance, in the [Explore page](https://github.com/explore) there's a bunch of `TabNav` being used for rendering repository links. We currently always render page links within a `nav` element. However, `nav` comes with an implicit role of `navigation` which should be reserved for MAJOR links. If we were to eliminate the use of `nav` here, list will provide semantic structure.

## References
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav
https://www.w3.org/TR/wai-aria-practices/examples/landmarks/navigation.html

### DIFF
--- a/app/components/primer/navigation/tab_component.rb
+++ b/app/components/primer/navigation/tab_component.rb
@@ -71,11 +71,6 @@ module Primer
       #     <% c.counter(count: 10) %>
       #   <% end %>
       #
-      # @example Inside a list
-      #   <%= render(Primer::Navigation::TabComponent.new(list: true)) do |c| %>
-      #     <% c.text { "Tab" } %>
-      #   <% end %>
-      #
       # @example With custom HTML
       #   <%= render(Primer::Navigation::TabComponent.new) do %>
       #     <div>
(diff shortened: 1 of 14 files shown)


app/components/primer/navigation/tab_component.rb
```diff
@@ -106,6 +99,7 @@ def initialize(list: false, selected: false, with_panel: false, icon_classes: ""
 
         @wrapper_arguments = wrapper_arguments
         @wrapper_arguments[:tag] = :li
+        @wrapper_arguments[:role] = :presentation if with_panel
```
khiga8, line 102: [Read more about presentation role](https://www.w3.org/TR/wai-aria-practices/#presentation_role)

app/components/primer/tab_nav_component.rb
```diff
@@ -44,23 +47,6 @@ class TabNavComponent < Primer::Component
     #     <% c.tab(href: "#") { "Tab 3" } %>
     #   <% end %>
     #
-    # @example With icons and counters
```
khiga8, line 47: Moving this down so `with panel` example comes second

app/components/primer/navigation/tab_component.rb
joelhawksley: This issue comment might make a good code comment ❤️

app/components/primer/tab_nav_component.rb
```diff
@@ -132,19 +135,23 @@ def initialize(label:, with_panel: false, body_arguments: {}, wrapper_arguments:
       @body_arguments = body_arguments
       @wrapper_arguments = wrapper_arguments
 
-      @system_arguments[:tag] = :div
+      @system_arguments[:tag] = navigation_tag(with_panel)
```
khiga8, line 138: I don't want to bloat this PR so in a follow up, I will change this so consumers have option to render page navigation links NOT in a `nav`.

Currently page navigation always gets rendered in a `nav`.`nav` may not be appropriate when the links aren't for main navigation. I noticed the GitHub explore page uses `TabNav` to list repository links, and these are showing up as navigation landmark which we probably don't want here. Navigation landmark should be reserved for primary/secondary navigation.

NVDA landmark menu for github explore page showing multiple navigations

We should have option to render `div` instead of `nav so the markup is:
```
<div> 
  <ul>
    <li>
    <li>
  </ul>
</div>
```

khiga8:
Closing this in favor of https://github.com/primer/view_components/pull/689

PR #671, opened 2021-06-29
Update `Navigation::Tab` to ensure panel has correct accessibility attributes

khiga8:
Depends on https://github.com/primer/view_components/pull/668

## What

- Ensures that tab panel has `tabindex=0` set which puts the panel in the tab sequence and allows assistive technology users to tab into the panel.
- Enforces that tab panel is properly labeled by the tab button text (or otherwise labeled with an `aria-label`).

## Why 

Both these updates were made following recommendations in: [WAI-ARIA tab example](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html) as well as [WAI-ARIA tab panel patterns](https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties-20).


khiga8:
Closing and will likely reopen later.

PR #674, opened 2021-07-01
Linter to enforce usage of System Arguments

manuelpuyol:
Uses `Primer::Classify::Utilities` to figure out if a component should be using system arguments instead of `classes`. This will provide live feedback while developing and also help us migrate usage as we rollout new utilities.

I had to update `Primer::Classify::Utilities` to not depend on `Rails` to make the linter easier to use.

# Warnings

image

# Autofix

image


rubocop/cop/primer/system_argument_instead_of_class.rb
```diff
@@ -0,0 +1,75 @@
...
+      # This cop ensures that components use System Arguments instead of CSS classes.
+      #
+      # bad
+      # Component.new(classes: "mr-1")
+      #
+      # good
+      # Component.new(mr: 1)
+      class SystemArgumentInsteadOfClass < RuboCop::Cop::Cop
+        INVALID_MESSAGE = <<~STR
+          Avoid using CSS classes when you can use System Arguments.
```
joelhawksley, line 19: Perhaps we could include a link to the systems arguments docs in this message?

```suggestion
          Avoid using CSS classes when you can use System Arguments: https://primer.style/view-components/system-arguments.
```

joelhawksley (approved):

Issue #676, opened 2021-07-02
Standardize PVC component names


manuelpuyol:
## Overview

We don't have a standard for our component names, having three different variants:

- `Primer::ButtonComponent` - with suffix
- `Primer::AutoComplete` - no suffix
- `Primer::Alpha::ButtonMarketing` - status as a module

This causes confusion and a not ideal developer experience, so we should have a single naming standard.

## Proposal

Rename all components to have their status as a module and remove the `Component` suffix.

Examples:

- `Primer::ButtonComponent` -> `Primer::Beta::Button`
- `Primer::AutoComplete` -> `Primer::Beta::AutoComplete`

Removing the suffix will make working with components less verbose. Using the status as a module will help with status promotion because we can keep the previous version for some time, giving users time to migrate.

## How to migrate a component

The majority of the work is done through the `component_status_migrator` thor command. This command will move files, update references, create a deprecated component in place of the original, fix navigation in storybook and the doc site, and more.

### Run a migration

As an example, the above referenced `Primer::ButtonComponent` can be migrated with the following command line:

```shell
bundle exec thor component_status_migrator ButtonComponent --status beta
```

Note that this command specifically calls out the target status for the component in `--status beta`. If this option is left off, it will default to `alpha` and move the component into the corresponding `alpha` folder and `Alpha::` namespace. 

Be sure to set the correct status when migrating a component. Valid status entries include:

* alpha
* beta
* stable
* deprecated (not fully supported yet)

### Testing a migration

Prior to creating a pull request and marking it as ready for review, be sure to run all tests, check storybook, and check the docs site to ensure all updates have been made:

* run `script/test`
* run `script/dev`
  * go to `http://localhost:5000` to check storybook
  * go to `localhost:5400` to check the doc site

### Common errors

You may run into errors with storybook or the docs site, due to previously built site information. For example, storybook may still have the original `Primer::ButtonComponent` entry while also having the updated `Primer::Beta::Button` entry. If this happens, you will need to clean out the build cache and rebuild. One way to do that is to run the following:

1. `git clean -dfx`
2. `script/setup`

The first call to clean will remove all untracked files and folders that are not part of `.gitignore`. The second call will remove all untracked files and folders that are part of `.gitignore`. This will remove everything, including nod_modules and bundle vendor folders. After that, you'll need to run `script/setup` to reinstall and rebuild before running `script/dev` again.
 
 ### Example pull requests
 
See the component migration work that is already completed, below, for example pull requests.

[...]

PR #678, opened 2021-07-12
Move AutoComplete components to Beta module

manuelpuyol:
Part of https://github.com/primer/view_components/issues/676

Renames:
 
1. `Primer::AutoComplete` to `Primer::Beta::AutoComplete`
2. `Primer::AutoComplete::Item` to `Primer::Beta::AutoComplete::Item`

I had to update a bunch of configuration files since having a `stories/primer/beta/auto_complete` directory was making zeitwerk assume that `Primer::Beta::AutoComplete` was a module.
The changes are:

1. Stop loading stories files when generating docs (prevents the zeitwerk issue)
2. Update `docs-build` CI to set an env var that will do the above
3. Creating `stories/primer/beta/auto_complete.rb` to explicity set `Primer::Beta::AutoComplete` as a class for the stories build


joelhawksley (approved):
Do you think the autoloading issues might come up in other environments?

manuelpuyol:
> Do you think the autoloading issues might come up in other environments?

I don't think so, the issue is caused by the stories, which aren't included when we package the gem

Issue #680, opened 2021-07-13
Update tab navigation API and make accessibility fixes


khiga8:
### Problem
Our Tab navigation components are missing some necessary accessibility attributes. In addition, the tab navigation component APIs are becoming really complex. For example, `UnderlineNav` can be rendered as page navigation links or panel navigation depending on a single `with_panel` attribute. Based on whether panels are rendered, the markup is totally different yet the documentation is missing the nuances for these differences. We should break down `UnderlineNav`and `TabNav` into two components, one that is specific to page navigation links, and one that is specific to panel navigation links.

### User Story
As a developer, I want an accessible tab component in my project, and I want to be aware of all the accessibility considerations.

### Exit Criteria
- [x] Ensure tab panels are labelled and have tabindex set to 0. #683 
- [x] Ensure tab panels and tab are associated with aria-controls #702 
- [x] Break UnderlineNav into UnderlineNav and UnderlinePanel #689 
- [ ] Break TabNav into TabNav and TabPanel #785

khiga8:
**1. Markup for Navigating links**

- **Case 1:** As MAIN navigation. `nav` has implicit role of `navigation` which will make it accessible via landmarks shortcuts. This should be used sparingly.

```.html
<nav aria-label="something">
  <ul>
    <li><a>Link 1</a></li>
    <li><a>Link 2</a></li>
    <li><a>Link 3</a></li>
  <ul>
</nav>
```

- **Case 2:** As regular list of links. `div` tag when group of links are NOT main links.

```.html
<div>
  <ul>
    <li><a>Link 1</a></li>
    <li><a>Link 2</a></li>
    <li><a>Link 3</a></li>
  <ul>
</div>
```

**2. Markup for Panel navigation**

- **Case 1:** Only case. Contained within a web component which comes with associated js behavior.

```.html
<tab-container>
  <div>
    <ul role="tablist">
      <li role="presentation"><button id="tab-1" role="tab">Tab 1</li>
      <li role="presentation"><button id="tab-2" role="tab">Tab 2</li>
      <li role="presentation"><button id="tab-3" role="tab">Tab 3</li>
    <ul>
  </div>
</tab-container>
<div role="tabpanel" aria-labelledby="tab-1" tabindex="0">Panel content</div>
<div role="tabpanel" aria-labelledby="tab-2" tabindex="0">Panel content</div>
<div role="tabpanel" aria-labelledby="tab-3" tabindex="0">Panel content</div>
```

[...]

PR #683, opened 2021-07-19
Ensure tab panels are labelled and have `tabindex` set to 0.

khiga8:
## What
- Updates panel slot in `Navigation::Tab` to ensure there's a label. For ease, if `id` is set on the tab, we auto-label the panel with the tab text using `aria-labelledby`. Consumers may alternatively pass in an `aria-label`.
- Sets panel `tabindex="0"`. 

## Why
According to [WAI-ARIA Roles, States, and Properties for tabs](https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties-20) and [Example of Tabs](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-2/tabs.html), tab panels should be labelled. Additionally, tab panel should have a tabindex set to 0.

## Notes

- There are a lot of updates I want to make with the tab components but I also want to keep the diffs small for ease of review. Ultimately, I think we need to move towards separating the current `UnderlineNav` into `UnderlinePanelNav` and `UnderlinePageNav`. Same thing with `TabNav`. [Related issue](https://github.com/primer/view_components/issues/680)

### DIFF
--- a/app/components/primer/navigation/tab_component.rb
+++ b/app/components/primer/navigation/tab_component.rb
@@ -18,9 +18,21 @@ module Primer
       #
       # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
       renders_one :panel, lambda { |**system_arguments|
+        return unless @with_panel
+
         system_arguments[:tag] = :div
         system_arguments[:role] ||= :tabpanel
+        system_arguments[:tabindex] = 0
         system_arguments[:hidden] = true unless @selected
+
+        label_present = system_arguments[:"aria-label"] || system_arguments.dig(:aria, :label) || system_arguments[:"aria-labelledby"] || system_arguments.dig(:aria, :labelledby)
+        unless label_present
+          if @id.present?
+            system_arguments[:"aria-labelledby"] = @id
+          elsif !Rails.env.production?
+            raise ArgumentError, "Panels must be labelled. Either set a unique `id` on the tab, or set an `aria-label` directly on the panel"
+          end
+        end
 
         Primer::BaseComponent.new(**system_arguments)
       }
@@ -93,8 +105,10 @@ module Primer
         @selected = selected
         @icon_classes = icon_classes
         @list = list
+        @with_panel = with_panel
 
         @system_arguments = system_arguments
+        @id = @system_arguments[:id]
 
         if with_panel || @system_arguments[:tag] == :button
           @system_arguments[:tag] = :button
--- a/app/components/primer/navigation/tab_component.rb
+++ b/app/components/primer/navigation/tab_component.rb
@@ -25,7 +25,7 @@ module Primer
         system_arguments[:tabindex] = 0
         system_arguments[:hidden] = true unless @selected
 
-        label_present = aria("label", system_arguments) || system_arguments[:"aria-labelledby"] || system_arguments.dig(:aria, :labelledby)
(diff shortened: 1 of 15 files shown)


manuelpuyol:
I'm pretty sure that `tab-container` should manage that for us 🤔

khiga8:
> I'm pretty sure that `tab-container` should manage that for us 🤔

@manuelpuyol  Nope :( 

The labeling is recommended in the [example markup](https://github.com/github/tab-container-element#usage) for `tab-container`.

manuelpuyol:
> > I'm pretty sure that `tab-container` should manage that for us 🤔
> 
> @manuelpuyol Nope :(
> 
> The labeling is recommended in the [example markup](https://github.com/github/tab-container-element#usage) for `tab-container`.

what about the `tabindex`? I think the element sets it after selecting a tab

khiga8:
> > > I'm pretty sure that `tab-container` should manage that for us 🤔
> > 
> > 
> > @manuelpuyol Nope :(
> > The labeling is recommended in the [example markup](https://github.com/github/tab-container-element#usage) for `tab-container`.
> 
> what about the `tabindex`? I think the element sets it after selecting a tab

Hrmm yeah. The `tabindex` gets set if tab is selected but we should make sure it's set always.

app/components/primer/navigation/tab_component.rb
```diff
@@ -18,10 +18,22 @@ class TabComponent < Primer::Component
...
       # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
       renders_one :panel, lambda { |**system_arguments|
+        return unless @with_panel
+
         system_arguments[:tag] = :div
         system_arguments[:role] ||= :tabpanel
+        system_arguments[:tabindex] = 0
         system_arguments[:hidden] = true unless @selected
 
+        label_present = system_arguments[:"aria-label"] || system_arguments.dig(:aria, :label) || system_arguments[:"aria-labelledby"] || system_arguments.dig(:aria, :labelledby)
```
manuelpuyol, line 28: I think it is time to write helpers to get aria values, this double check is getting propagated everywhere and it's not great 😅 

We could do something like this in `Primer::Component`:
```rb
def aria(val, system_arguments)
  system_arguments[:"aria-#{val}"] || system_arguments.dig(:aria, val)
end
```

khiga8: great idea!

khiga8: I may add a linter in a followup.

app/components/primer/navigation/tab_component.rb
```diff
@@ -18,10 +18,22 @@ class TabComponent < Primer::Component
...
       # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
       renders_one :panel, lambda { |**system_arguments|
+        return unless @with_panel
+
         system_arguments[:tag] = :div
         system_arguments[:role] ||= :tabpanel
+        system_arguments[:tabindex] = 0
         system_arguments[:hidden] = true unless @selected
 
+        label_present = aria("label", system_arguments) || system_arguments[:"aria-labelledby"] || system_arguments.dig(:aria, :labelledby)
```
manuelpuyol, line 28:
```suggestion
        label_present = aria("label", system_arguments) || aria("labelledby", system_arguments)
```

manuelpuyol (approved):

PR #688, opened 2021-07-19
Update docs generation to point to the correct urls

manuelpuyol:
The `source` and `storybook` urls weren't taking the status module into account, so I'm updating `docs.rake` to take them into consideration


docs/src/@primer/gatsby-theme-doctocat/components/head.js
```diff
@@ -18,7 +18,7 @@ function Head(props) {
 
   let primerViewComponentsSrc
 
-  if(process.env.NODE_ENV == 'development') {
+  if(process.env.NODE_ENV === 'development') {
```
srt32, line 21: :)

lib/tasks/docs.rake
```diff
@@ -385,4 +384,55 @@ namespace :docs do
 
     pretty_value(constant_value)
   end
+
+  def status_module_and_short_name(component)
+    name_with_status = component.name.gsub(/Primer::|Component/, "")
+
+    m = name_with_status.match(/(?<status>Beta|Alpha)?(::)?(?<name>.*)/)
```
srt32, line 391: We don't need deprecated here?

srt32 (approved):
Much nicer ;)

lib/tasks/docs.rake
manuelpuyol: we do 😬

PR #690, opened 2021-07-20
Add ENV flag to dump linter data into a file.

manuelpuyol:
Dumping the data into a file will be helpful to report that data to services like Datadog

### DIFF
--- a/lib/primer/view_components/linters/helpers.rb
+++ b/lib/primer/view_components/linters/helpers.rb
@@ -13,10 +13,13 @@ module ERBLint
         link menuitem meta param source track wbr img
       ].freeze
 
+      DUMP_FILE = ".erblint-counter-ignore.json"
+
       def self.included(base)
         base.include(ERBLint::LinterRegistry)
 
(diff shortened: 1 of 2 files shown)


joelhawksley (approved):

lib/primer/view_components/linters/helpers.rb
```diff
@@ -13,10 +13,13 @@ module Helpers
         link menuitem meta param source track wbr img
       ].freeze
 
+      DUMP_FILE = ".erblint-counter-ignore.json"
```
srt32, line 16: Could we put this in the /static dir?

manuelpuyol: We could, but the goal here isn't to keep the file. The idea is that this file will be generated when reporting data to Datadog and then removed. This allow us to enable specific linters when reporting metrics without having them always enables (and causing noise with ignore comments in all files)
We also don't want to commit this file since it will generate **tons** of merge conflicts

PR #692, opened 2021-07-20
Expose custom cops and default config for erblint

manuelpuyol:
Moves custom cops to `lib/` dir so they are packed into the gem and customers are able to use the cops in their apps


joelhawksley (approved):
Do you think we should add anything to the docs for this new capability in the gem?

manuelpuyol:
> Do you think we should add anything to the docs for this new capability in the gem?

Going to add in a followup!

PR #693, opened 2021-07-20
Adding `Primer::Beta::Truncate` component to reflect changes in primer/css component

jonrohan:
The [primer-css `.Truncate` classes have changed](https://primer.style/css/components/truncate) as of PR https://github.com/primer/css/pull/1358, and I wanted to update the `Primer::Truncate` component to reflect that change, but there was so much of a departure I thought I should create a `Primer::Beta::Truncate` component.

I have documented all the changes under the [`/components/beta/truncate`](https://view-components-gjo80b1y7-primer.vercel.app/view-components/components/beta/truncate) path on primer.style. There's not a lot to change on dotcom, but the structure uses slots now instead of content.


stories/primer/beta/truncate/truncate_text_stories.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class Primer::Beta::Truncate::TruncateTextStories < ViewComponent::Storybook::Stories
```
manuelpuyol, line 3:
```suggestion
# frozen_string_literal: true

require "primer/beta/truncate"

class Primer::Beta::Truncate::TruncateTextStories < ViewComponent::Storybook::Stories
```

manuelpuyol: For some reason the stories can't autoload subdirectories 🤷

stories/primer/beta/truncate_stories.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "primer/alpha/button_marketing"
```
manuelpuyol, line 3:
```suggestion
require "primer/beta/truncate"
```

app/components/primer/beta/truncate.rb
```diff
@@ -0,0 +1,116 @@
...
+      #   <% end %>
+      #   <br/>
+      #   <%= render(Primer::Beta::Truncate.new) do |component| %>
+      #     <% component.item(max_width: 100, expandable: true) do %>branch-name-that-is-really-long-branch-name-that-is-really-long-branch-name-that-is-really-long<% end %>
+      #   <% end %>
+      #
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+      def initialize(**system_arguments)
+        @system_arguments = system_arguments
+        @system_arguments[:tag] = system_arguments[:tag] || :span
```
manuelpuyol, line 70: does this component support **all** tags? We were trying to restrict usable tags for components

cc @khiga8

app/components/primer/beta/truncate.rb
```diff
@@ -0,0 +1,116 @@
...
+      def before_render
+        return unless content.present? && items.empty?
+
+        # Get the slot arguments out of the system arguments
+        slot_arguments = @system_arguments.slice(:priority, :expandable, :max_width)
+        @system_arguments.delete(:priority)
+        @system_arguments.delete(:expandable)
+        @system_arguments.delete(:max_width)
+
+        set_slot(:items, **slot_arguments) { content }
```
manuelpuyol, line 86: would `item(**slot_arguments) { content }` work? I'd try to avoid calling direct ViewComponent APIs

app/components/primer/beta/truncate.rb
```diff
@@ -0,0 +1,116 @@
...
+          "Truncate",
+          system_arguments[:classes]
+        )
+      end
+
+      def before_render
+        return unless content.present? && items.empty?
+
+        # Get the slot arguments out of the system arguments
+        slot_arguments = @system_arguments.slice(:priority, :expandable, :max_width)
```
manuelpuyol, line 81: I think it can be confusing that those arguments only work if you pass content, I can see people trying to use

```rb
render Primer::Beta::Truncate.new(max_width: 100) do |c|
  c.item ...
  c.item ...
```

expecting that `max_width` would propagate to all the slots. In this case, the arguments also won't be sliced so they'll be rendered in html as `max-width="100"`.

Maybe we shouldn't allow customization through the parent class and if someone wants to change any of those arguments they should use the slot.

app/components/primer/beta/truncate.rb
```diff
@@ -0,0 +1,116 @@
...
+      def render?
+        items.any?
+      end
+
+      # This component is part of `Primer::Beta::Truncate` and should not be
+      # used as a standalone component.
+      class TruncateText < Primer::Component
+        def initialize(priority: false, expandable: false, max_width: nil, **system_arguments)
+          @system_arguments = system_arguments
+          @system_arguments[:tag] = system_arguments[:tag] || :span
```
manuelpuyol, line 98: I'd like to see which tags we accept here too @khiga8

app/components/primer/beta/truncate.rb
jonrohan: > Maybe we shouldn't allow customization through the parent class and if someone wants to change any of those arguments they should use the slot.

I'm down with this 👍🏻 I was trying to allow for the shorter hand when not using the slot

manuelpuyol (approved):

PR #695, opened 2021-07-20
Fix double constant assign

manuelpuyol:


app/components/primer/avatar_stack_component.rb
```diff
@@ -12,7 +12,7 @@ class AvatarStackComponent < Primer::Component
     TAG_OPTIONS = [DEFAULT_TAG, :span].freeze
 
     DEFAULT_BODY_TAG = :div
-    BODY_TAG_OPTIONS = TAG_OPTIONS = [DEFAULT_TAG, :span].freeze
+    BODY_TAG_OPTIONS = [DEFAULT_BODY_TAG, :span].freeze
```
srt32, line 15: Is this not a functional change?

srt32: or, hm, not it's not since `DEFAULT_TAG` == `DEFAULT_BODY_TAG`. We need both?

srt32 (approved):

app/components/primer/avatar_stack_component.rb
manuelpuyol: no, the arrays end up being the same

PR #696, opened 2021-07-21
Ensure preview generator parses example tag correctly

khiga8:
## What 

- Ensures preview generation code accounts for when `@example` has `@code` and `@description`. 
- Extracts example tag parsing into a method. 

## Why

Currently the generated preview view code can look something like:

```
@description
  Add an `icon` to give additional context. Refer to the [Octicons](https://primer.style/octicons/) documentation to choose an icon.
@code
  <%= render Primer::BlankslateComponent.new(
    icon: :globe,
    title: "Title",
    description: "Description",
  ) %>
``` 

which we don't want. We only care about the code.

### DIFF
--- a/lib/tasks/docs.rake
+++ b/lib/tasks/docs.rake
@@ -227,23 +227,12 @@ namespace :docs do
         f.puts("## Examples")
 
         initialize_method.tags(:example).each do |tag|
-          name = tag.name
-          description = nil
-          code = nil
-
-          if tag.text.include?("@description")
-            splitted = tag.text.split(/@description|@code/)
-            description = splitted.second.gsub(/^[ \t]{2}/, "").strip
-            code = splitted.last.gsub(/^[ \t]{2}/, "").strip
-          else
-            code = tag.text
-          end
-
+          name, description, code = parse_example_tag(tag)
           f.puts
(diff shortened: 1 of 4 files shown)


lib/yard/docs_helper.rb
```diff
@@ -29,7 +29,7 @@ def one_of(enumerable, lower: false, sort: true)
     end
 
     def link_to_accessibility
-      "[Accessibility](#system-arguments)"
+      "[Accessibility](#accessibility)"
```
khiga8, line 32: (unrelated but minor fix)

lib/tasks/docs.rake
```diff
@@ -227,23 +227,12 @@ namespace :docs do
...
-            code = tag.text
-          end
-
+          name, description, code = parse_example_tag(tag)
           f.puts
           f.puts("### #{name}")
           if description
             f.puts
-            f.puts(description)
+            f.puts(view_context.render(inline: description.squish))
```
khiga8, line 235: Makes sure that if inline erb in the description gets parsed properly. 

What we don't want:
```
This is a description. See <%= link_to_accessibility %>
```

manuelpuyol (approved):

PR #698, opened 2021-07-21
Improve classify utilities performance

manuelpuyol:
`classes_to_hash` now uses the `utilties.rake` logic to infer a possible key for the selector being analyzed instead of looping through the whole utilities hash. Since the new `SystemArgumentInsteadOfClass` linter uses that utility, it makes sense to improve it where we can 😄.
This changes speeds the helper up ~3 times.

# Benchmark

Benchmark code here https://gist.github.com/manuelpuyol/a0f1b5a2840315e3c7d6fc3144d86ca5

```
Running 10000 times.
       user     system      total        real
Old  1.072237   0.000000   1.072237 (  1.072257)
New  0.375978   0.000000   0.375978 (  0.375981)
Running 50000 times.
       user     system      total        real
Old  5.366870   0.000000   5.366870 (  5.366916)
New  1.890638   0.000000   1.890638 (  1.890662)
Running 100000 times.
       user     system      total        real
Old 10.710371   0.000000  10.710371 ( 10.710519)
New  3.762228   0.000000   3.762228 (  3.762282)
```

### DIFF
--- a/lib/primer/classify/utilities.rb
+++ b/lib/primer/classify/utilities.rb
@@ -16,6 +16,15 @@ module Primer
       # rubocop:enable Security/YAMLLoad
       BREAKPOINTS = ["", "-sm", "-md", "-lg", "-xl"].freeze
 
+      # Replacements for some classnames that end up being a different argument key
+      REPLACEMENT_KEYS = {
+        "^anim" => "animation",
+        "^v-align" => "vertical_align",
+        "^d" => "display",
+        "^wb" => "word_break",
+        "^v" => "visibility"
+      }.freeze
+
       class << self
         def classname(key, val, breakpoint = "")
           if (valid = validate(key, val, breakpoint))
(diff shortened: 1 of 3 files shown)


jonrohan (approved):
Wow! Nice improvements.

lib/primer/classify/utilities.rb
```diff
@@ -104,23 +115,33 @@ def classes_to_hash(classes)
...
+            # has index of classname, so we can match it up with responsvie array `mr: [nil, 0]`
+            return [key, key_argument, classnames.index(selector)]
           end
 
           nil
         end
 
+        def infer_selector_key(selector)
+          REPLACEMENT_KEYS.each do |k, v|
+            return v.to_sym if selector.match?(Regexp.new(k))
```
jhawthorn, line 139: Building a new regex each iteration is going to be slow. Instead we could probably have the keys to this hash be regexes, or even better, since these all seem to be testing prefixes, we could use `String#start_with?`.

Is the to_sym necessary, or could we keep these values as strings?

lib/primer/classify/utilities.rb
```diff
@@ -104,23 +115,33 @@ def classes_to_hash(classes)
...
+
+          # Each value hash will also contain an array of classnames for breakpoints
+          # Key argument `0`, classes `[ "mr-0", "mr-sm-0", "mr-md-0", "mr-lg-0", "mr-xl-0" ]`
+          value_hash.each do |key_argument, classnames|
+            # Skip each value hash until we get one with the selector
+            next unless classnames.include?(selector)
+
+            # Return [:mr, 0, 1]
+            # has index of classname, so we can match it up with responsvie array `mr: [nil, 0]`
+            return [key, key_argument, classnames.index(selector)]
```
jhawthorn, line 131: `include?` and `index` are each going to do a linear search through the array. We could probably make this faster by precomputing a hash of `selector` to `index` (or nil if missing)

Ideally, if we flattened all those classnames to a single level hash, we could even do that without iterating over the `value_hash` 🤞

jhawthorn (commented):
I saw this method in a profile! Glad to see it's already on the radar for improvements ❤️. It looks like we could still remove a few more loops here, let me know if you'd like to pair on this 😁.

PR #700, opened 2021-07-21
Add linter suggestions for `Label` component.

manuelpuyol:
The linter already supports auto correction


lib/primer/view_components/linters/argument_mappers/base.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require_relative "conversion_error"
+require_relative "system_arguments"
+
+module ERBLint
+  module Linters
+    module ArgumentMappers
+      # Maps classes in a label element to arguments for the Label component.
```
joelhawksley, line 9: Is this comment still accurate?

lib/primer/view_components/linters/argument_mappers/label.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require_relative "base"
+
+module ERBLint
+  module Linters
+    module ArgumentMappers
+      # Maps classes in a label element to arguments for the Label component.
+      class Label < Base
+        SCHEME_MAPPINGS = {
```
joelhawksley, line 10: I'd love to be able to reuse these from the components, as copying them here will likely lead to them being out of date. What do you think?

lib/primer/view_components/linters/argument_mappers/base.rb
manuelpuyol:
```suggestion
      # Provides the base interface to implement an `ArgumentMapper`.
      # To use, inherit this class and implement a `attribute_to_args(attribute)` method.
```

lib/primer/view_components/linters/argument_mappers/base.rb
```diff
@@ -0,0 +1,33 @@
...
+          args = {}
+
+          args[:tag] = ":#{@tag.name}" unless @tag.name == self.class::DEFAULT_TAG
+
+          @tag.attributes.each do |attribute|
+            args.merge!(attribute_to_args(attribute))
+          end
+
+          args
+        end
```
manuelpuyol, line 29:
```suggestion
        end
        
        def attribute_to_args(attribute); end
```

lib/primer/view_components/linters/argument_mappers/label.rb
manuelpuyol: I'd love to have access to the component's constants here, but that'd require including the component in this file, which will basically load the whole Rails app.
I thought about having a separate `Constants` namespace that could hold that information and could be shared between the component definition and the linters, but that would make development harder

joelhawksley: @manuelpuyol gotcha. Can you describe the overhead and downsides of loading the whole Rails app?

manuelpuyol: Since we rely on autoloading to load our components (and helpers), we can't simply require the components and be good to go. We'd need to manually load everything needed - ViewComponent framework, component, helpers...
That can get very complicated once the library grows.

joelhawksley (approved):
👍🏻 let's look at extracting the constants automatically in a follow up PR ❤️

PR #701, opened 2021-07-21
WIP script for standardizing component naming

srt32:
In support of https://github.com/primer/view_components/issues/676

Usage

```
bundle exec rake renamer:add_status[AvatarStackComponent,beta]
```

The script will
* update the class namespace to be within the proper status module - `module Beta`
* update the component's filepath and name to drop the `Component` suffix
* find and replace all instances to include the status prefix and drop the component suffix

The script is far from perfect but should get a dev most of the way to handling each renaming. There are inline TODO's we could work on improving but this might be good enough to have us chug through the list of renames.

### DIFF
--- a/lib/tasks/renamer.rake
+++ b/lib/tasks/renamer.rake
@@ -57,6 +57,8 @@ namespace :renamer do
     puts mv_result
 
     # TODO: rename template filename too
+    # TODO: rename story too
+    # TODO: rename nav.yml
 
     # rename in codebase for status and suffix
     if target_component_name.end_with?("Component")
--- a/lib/tasks/renamer.rake
+++ b/lib/tasks/renamer.rake
@@ -2,12 +2,12 @@
 
 # bundle exec rake renamer:add_status[AvatarStack,beta]
 namespace :renamer do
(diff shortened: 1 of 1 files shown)


manuelpuyol:
A few things that we could add:

1. moving the test file into `test/components/STATUS/`
  i. Could also update the test class name from `PrimerComponentNameTest` to `PrimerSTATUSComponentNameTest`
2. moving the stories into the `status` folder too
  i. We need to add a `require "primer/STATUS/component"` at the top for the stories to work properly
3. Run rubocop with autocorrect (which will fix indentation errors)
4. Updating `nav.yml` with the new path to docs. From `components/componentname` to `components/STATUS/componentname`

I think that's all we need to update the module 😅

srt32:
I pushed two more commits to rename the test file (I did not change the test class name) and run rubocop

srt32:
Remaining TOOD:'s

* moving the stories into the status folder too
  i. We need to add a require "primer/STATUS/component" at the top for the stories to work properly
* move the template file too
* don't mess up the class definition in the component file

srt32:
superseded by https://github.com/primer/view_components/pull/704

PR #689, opened 2021-07-20
Break `UnderlineNav` into `UnderlineNav` and `UnderlinePanel`

khiga8:
#### What
- Breaks `UnderlineNav` into two components:
  - `UnderlineNav` for page links.
  - `UnderlinePanel` for panels and JavaScript.

#### Why
- `UnderlineNav`  API is getting bloated and will benefit from being broken down. While there's some overlap between the two, splitting this into 2 components allows a simpler API and more tailored documentation.

#### Notes
- Part of https://github.com/primer/view_components/issues/680
- I plan to break down `TabNav` similarly.

PR #702, opened 2021-07-22
Ensure tab panels and tab are associated with aria-controls

khiga8:
#### What
This PR requires `panel_id` to be set when the tab has an associated panel for panel navigation.

#### Why
This is necessary for accessibility. In order for tabs to be associated with the corresponding panel, the tab's `aria-controls` must be set to the `id` of the corresponding panel. 

##### References
-  [Deque ARIA Tab Panel Accessibility: A11y Support Series](https://www.deque.com/blog/a11y-support-series-part-1-aria-tab-panel-accessibility/)
-  [W3 Example of Tabs with Automatic Activation](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html)

#### Notes

The API for tab components is getting increasingly complex because we're supporting both page navigation and panel navigation within a single component. See issue #680. 

PR #689  will break down the `UnderlineNavComponent` into `UnderlineNav` and `UnderlinePanel`. While there's some overlap between the two, I think there are greater benefits to splitting this into two components including more tailored documentation, and a simpler API.

### DIFF
--- a/app/components/primer/tab_nav_component.rb
+++ b/app/components/primer/tab_nav_component.rb
@@ -13,7 +13,7 @@ module Primer
     # Tabs to be rendered. When `with_panel` is set on the parent, a button is rendered for panel navigation. Otherwise,
     # an anchor tag is rendered for page navigation. For more information, refer to <%= link_to_component(Primer::Navigation::TabComponent) %>.
     #
-    # @param panel_id [String] Only applies if `with_panel` is `true`. Unique id of panel.
+    # @param panel_id [String] Only applies if `with_panel` is `true`. Unique ID of panel.
     # @param selected [Boolean] Whether the tab is selected.
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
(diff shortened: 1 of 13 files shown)


app/components/primer/tab_nav_component.rb
```diff
@@ -13,6 +13,7 @@ class TabNavComponent < Primer::Component
     # Tabs to be rendered. When `with_panel` is set on the parent, a button is rendered for panel navigation. Otherwise,
     # an anchor tag is rendered for page navigation. For more information, refer to <%= link_to_component(Primer::Navigation::TabComponent) %>.
     #
+    # @param panel_id [String] Only applies if `with_panel` is `true`. Unique id of panel.
```
joelhawksley, line 16:
```suggestion
    # @param panel_id [String] Only applies if `with_panel` is `true`. Unique ID of panel.
```

joelhawksley (approved):

PR #703, opened 2021-07-22
update Text component path in sidebar

manuelpuyol:


joelhawksley (approved):
How can we make it so this doesn't happen again?

manuelpuyol:
> How can we make it so this doesn't happen again?

Probably we need to [automate the name migration](https://github.com/primer/view_components/pull/701)

PR #704, opened 2021-07-22
Thor component migrator

manuelpuyol:
Related to https://github.com/primer/view_components/issues/676

### Summary

Since we want to migrate all of our components into modules indicating their status, automating the migration task will save us a lot of time.
This PR creates a `thor` runner that does everything needed to move a component in the main `app/components` directory into `app/components/STATUS`.

### How to use

```sh
bundle exec thor component_status_migrator MyComponentName
bundle exec thor component_status_migrator MyComponentName --status beta
```

### What it does

1. Moves component files into a `/STATUS` subdirectory. This includes:
  i. Controller
  ii. Template
  iii. Test
  iv. Story
2. Adds the `STATUS` module to the component and removes the `Component` suffix.
   i. E.g.: `Primer::AvatarStackComponent` becomes `Primer::Beta::AvatarStack`.
3. Renames the test class from `PrimerComponentTest` to `PrimerStatusComponentTest`
4. Adds a `require "primer/status/component_name"` to the story.
  i. This is needed because of a limitation on the storybook gem.
5. Changes the `nav.yml` entry of the component to reflect the new docs path.
6. Updates all references to the component in the project to use the new name with module.
7. Runs rubocop to fix any linting issues.
8. Regenerates the docs.
9. Regenerates static files.

### How it works

This is based on Thor [generators](https://github.com/rails/thor/wiki/Generators. When running the script, thor will call all the **public** methods in the order they are defined in the file.
The big difference between using `thor` and a shell script is the helpers `thor` provides. See the [`Thor::Actions`](https://rubydoc.info/github/wycats/thor/master/Thor/Actions) documentation. It is also easier to read the script since it is all written in ruby and most of our contributors are Ruby developers.

### DIFF
--- a/component_status_migrator.thor
+++ b/component_status_migrator.thor
@@ -1,0 +1,152 @@
+# frozen_string_literal: true
+
+require "thor"
+require "active_support/core_ext/string/inflections"
+
+# Migrates components to their new namespace.
+#
+# Usage:
+#
+# bundle exec thor component_status_migrator MyComponentName
+# bundle exec thor component_status_migrator MyComponentName --status beta
+class ComponentStatusMigrator < Thor::Group
+  include Thor::Actions
+
+  STATUSES = %w[alpha beta].freeze
+
+  # Define arguments and options
+  argument :name
+  class_option :status, default: "alpha", desc: "Status of the component. Either alpha or beta", required: true, type: :string
+
+  def self.source_root
+    File.dirname(__FILE__)
+  end
+
+  def validate_status
+    raise unless STATUSES.include?(status)
+  end
+
+  def move_controller
+    raise unless File.exist?(controller_path)
+
+    copy_file(controller_path, controller_path_with_status)
+    remove_file(controller_path)
+  end
+
+  def move_template
+    if File.exist?(template_path)
+      copy_file(template_path, template_path_with_status)
+      remove_file(template_path)
+    else
+      puts "No template found"
+    end
+  end
+
+  def move_test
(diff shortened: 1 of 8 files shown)


component_status_migrator.thor
```diff
@@ -0,0 +1,152 @@
...
+# Migrates components to their new namespace.
+#
+# Usage:
+#
+# bundle exec thor component_status_migrator MyComponentName
+# bundle exec thor component_status_migrator MyComponentName --status beta
+class ComponentStatusMigrator < Thor::Group
+  include Thor::Actions
+
+  STATUSES = %w[alpha beta].freeze
```
srt32, line 15: no need for deprecated?

test/lib/yard/docs_helper_test.rb
```diff
@@ -13,7 +13,7 @@ def test_sorts_one_of
 
   def test_link_to_component
     assert_equal("[Button](/components/button)", link_to_component(Primer::ButtonComponent))
-    assert_equal("[AutoCompleteItem](/components/autocompleteitem)", link_to_component(Primer::Beta::AutoComplete::Item))
+    assert_equal("[AutoCompleteItem](/components/beta/autocompleteitem)", link_to_component(Primer::Beta::AutoComplete::Item))
```
srt32, line 16: This was a typo fix?

srt32 (approved):
This is excellent. Let's give it a go for a few and see how we like it.

PR #715, opened 2021-07-29
add support for width: :full and height: :full

### LINKED ISSUE #638: Add support for `width-full`
### Problem

System arguments do not support `width-full`.

### Proposed solution

Map `width: :full` to `width-full`.

joelhawksley:
[Closes #638]

### DIFF
--- a/lib/primer/classify/utilities.yml
+++ b/lib/primer/classify/utilities.yml
@@ -18,6 +18,8 @@
   - anim-pulse
   :pulse_in:
   - anim-pulse-in
+  :hover_grow:
+  - anim-hover-grow
   :rotate:
   - anim-rotate
 :position:
@@ -142,6 +144,8 @@
   - m-md-6
   - m-lg-6
   - m-xl-6
+  :auto:
+  - m-auto
 :mt:
   0:
   - mt-0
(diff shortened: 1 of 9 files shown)


manuelpuyol (commented):
Instead of adding the `:full` key, could we move this to use `Primer::Classify::Utilities`?

joelhawksley:
@manuelpuyol I just tried that and ran into issues with needing `width` and `height` to support falling back to integer values. 

Is there a way to support that in Utilities? If not, I guess I could add a special case to Classify for the integers and let Utilities handle the mapping to CSS.

manuelpuyol:
> I guess I could add a special case to Classify for the integers and let Utilities handle the mapping to CSS.

I think this would be the ideal case 🤔

lib/primer/classify/utilities.yml
```diff
@@ -1269,3 +1269,13 @@
...
+:height:
+  :fit:
+  - height-fit
+  :full:
+  - height-full
+:width:
+  :fit:
+  - width-fit
+  :full:
+  - width-full
```
manuelpuyol: did you add those manually? I think you are supposed to add the keys to https://github.com/primer/view_components/blob/ce6e757ca91729f201faeee641ef7967568ba5fa/lib/tasks/utilities.rake#L11-L21 and run the task :)

manuelpuyol (approved):

PR #718, opened 2021-08-02
update to octicons 15, removing open-ended dependency

joelhawksley:
This PR updates our `octicons` dependency, fixing a case where CI builds failed due to a breaking change in `octicon` `v15`.


app/components/primer/clipboard_copy.html.erb
```diff
@@ -2,7 +2,7 @@
   <% if content.present? %>
     <%= content %>
   <% else %>
-    <%= render Primer::OcticonComponent.new("clippy") %>
+    <%= render Primer::OcticonComponent.new("paste") %>
     <%= render Primer::OcticonComponent.new("check", color: :icon_success, style: "display: none;") %>
```
manuelpuyol, line 6: mind updating here to use symbols instead of strings?

manuelpuyol (approved):

PR #719, opened 2021-08-03
Create linter for CloseButton

manuelpuyol:
This linter is not autocorrectable (yet), because I'm not sure how to deal with cases where the `aria-label` is being set in the button's content. Since the component has a default aria-label, it may override the one being set.

```erb
<button class="close-button"><%= primer_octicon(:x, "aria-label": "Some Label") %></button>
```

This would become:

```erb
<%= render Primer::CloseButton.new do %>
  <%= primer_octicon(:x, "aria-label": "Some Label") %>
<% end %>
```

Which translates to:

```erb
<button class="close-button" aria-label="Close"><%= primer_octicon(:x, "aria-label": "Some Label") %></button>
```

### DIFF
--- a/lib/primer/view_components/linters/close_button_component_migration_counter.rb
+++ b/lib/primer/view_components/linters/close_button_component_migration_counter.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative "helpers"
+
+module ERBLint
+  module Linters
+    # Counts the number of times a HTML clipboard-copy is used instead of the component.
+    class CloseButtonComponentMigrationCounter < Linter
+      include Helpers
+
+      TAGS = %w[button].freeze
+      CLASSES = %w[close-button].freeze
+      MESSAGE = "We are migrating clipboard-copy to use [Primer::CloseButton](https://primer.style/view-components/components/closebutton), please try to use that instead of raw HTML."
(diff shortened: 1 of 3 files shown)


lib/primer/view_components/linters/close_button_component_migration_counter.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module ERBLint
+  module Linters
+    # Counts the number of times a HTML clipboard-copy is used instead of the component.
+    class CloseButtonComponentMigrationCounter < Linter
+      include Helpers
+
+      TAGS = %w[button].freeze
+      CLASSES = %w[close-button].freeze
+      MESSAGE = "We are migrating clipboard-copy to use [Primer::CloseButton](https://primer.style/view-components/components/closebutton), please try to use that instead of raw HTML."
```
joelhawksley, line 13: Do we want to update this copy?

manuelpuyol:
```suggestion
      MESSAGE = "We are migrating close-button to use [Primer::CloseButton](https://primer.style/view-components/components/closebutton), please try to use that instead of raw HTML."
```

joelhawksley (approved):

PR #727, opened 2021-08-05
Create linter to use `primer_octicon` instead of `octicon`

manuelpuyol:
This linter will check all calls to `octicon()` and suggest to use `primer_octicon` instead.
It will only suggest if:

1. no custom classes are used
2. no arguments other than `size, width, height, class` are used
3. `size, width, height` are either String or Int
4. `class` is a String

### DIFF
--- a/lib/rubocop/cop/primer/primer_octicon.rb
+++ b/lib/rubocop/cop/primer/primer_octicon.rb
@@ -20,7 +20,7 @@ module RuboCop
 
         SIZE_ATTRIBUTES = %i[height width size].freeze
         VALID_ATTRIBUTES = [*SIZE_ATTRIBUTES, :class].freeze
-        INVALID_CLASSES = -1
+        INVALID_ATTRIBUTE = -1
 
         def on_send(node)
           return unless node.method_name == :octicon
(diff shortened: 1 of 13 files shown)


lib/rubocop/cop/primer/primer_octicon.rb
```diff
@@ -0,0 +1,136 @@
...
+        private
+
+        def transform_sizes(kwargs)
+          attributes = octicon_size_attributes(kwargs)
+
+          attributes.transform_values do |size|
+            if size.between?(10, 16)
+              ""
+            elsif size.between?(22, 26)
+              ":medium"
```
joelhawksley, line 77: Are we sure about this mapping rule? I'm not so sure about automating this kind of change without visually verifying the results.

jonrohan: This is something I was doing on the dotcom side

lib/rubocop/cop/primer/primer_octicon.rb
```diff
@@ -0,0 +1,136 @@
...
+module RuboCop
+  module Cop
+    module Primer
+      # This cop ensures that components use System Arguments instead of CSS classes.
+      #
+      # bad
+      # octicon("icon")
+      #
+      # good
+      # primer_octicon("icon")
```
jonrohan, line 16: In the dotcom erblint I was transforming the icon to a symbol `octicon("icon") => primer_octicon(:icon)` thought that was preferable from a performance standpoint

lib/rubocop/cop/primer/primer_octicon.rb
joelhawksley: @jonrohan if you're OK with this change, I am!

lib/rubocop/cop/primer/primer_octicon.rb
manuelpuyol: I decided to keep the icon as-is because I'm actually supporting ruby code here (ivar, method calls, ternaries) and it was kinda hard to figure out if the icon was a plain string or something else, I'll give it another try

manuelpuyol: I updated it to deal with strings when passed directly, but kept ternaries untouched

jonrohan (approved):

PR #735, opened 2021-08-10
Convert ERB interpolations in linters

manuelpuyol:
Before, any time we encountered an ERB interpolation, the linter would skip the autocorrection.
Now, we'll support it for **some** values:

1. ClipboardCopy#value
2. LabelComponent#title
3. aria- values
4. data- values
5. test_selector


joelhawksley (approved):
🎊

PR #736, opened 2021-08-10
Add message about autocorrection

manuelpuyol:
Adds a message informing the user about running erblint in autocorrect mode when the offense is autocorrectable:

image


jonrohan (approved):
WOO! 🥇

PR #740, opened 2021-08-11
Promote ClipboardCopy to beta

manuelpuyol:
Since our goal is to reach 90% usage on this component, it makes sense to move it to beta


koddsson (approved):
Looks good to me!

PR #742, opened 2021-08-12
Support ruby 3.0

camertron:
Necessary to upgrade rubocop to do this, which of course led to a bunch of style fixes.


docs/content/system-arguments.md
```diff
@@ -55,57 +55,57 @@ System arguments include most HTML attributes. For example:
...
+| `border_color` | Symbol | Border color. One of `:primary`, `:secondary`, `:tertiary`, `:info`, `:success`, `:warning`, `:danger`, or `:inverse`. |
+| `color` | Symbol | Text color. One of `:icon_primary`, `:icon_secondary`, `:icon_tertiary`, `:icon_info`, `:icon_success`, `:icon_warning`, `:icon_danger`, `:text_primary`, `:text_secondary`, `:text_tertiary`, `:text_link`, `:text_success`, `:text_warning`, `:text_danger`, `:text_white`, or `:text_inverse`. |
 
 ## Flex
 
 | Name | Type | Description |
 | :- | :- | :- |
-| `align_items` | Symbol | One of `:baseline`, `:center`, `:flex_end`, `:flex_start`, or `:stretch`. |
-| `align_self` | Symbol | One of `:auto`, `:baseline`, `:center`, `:end`, `:start`, or `:stretch`. |
+| `align_items` | Symbol | One of `:flex_start`, `:flex_end`, `:center`, `:baseline`, or `:stretch`. |
```
joelhawksley, line 73: Do you think we should change the way we're sorting these arguments?

lib/yard/docs_helper.rb
```diff
@@ -6,7 +6,7 @@ module DocsHelper
     def one_of(enumerable, lower: false, sort: true)
       # Sort the array if requested
       if sort
-        compare = ->(a, b) { a.class == b.class ? a <=> b : a.class.to_s <=> b.class.to_s }
+        compare = ->(a, b) { a.class.instance_of?(b.class) ? a <=> b : a.class.to_s <=> b.class.to_s }
```
jonrohan, line 9: It seems like this broke the one of sorting? I'm not sure why though

camertron: Ah ok my bad, this should be `a.instance_of?(b.class)`. I think I misread the rubocop output.

docs/content/system-arguments.md
camertron: Definitely not, thanks for catching this! As @jonrohan mentioned, this happened as a result of changing the `one_of` helper. Should be fixed now.

jonrohan (approved):
Nice!

test/benchmarks/bench_classify.rb
```diff
@@ -40,7 +42,7 @@ def bench_allocations_without_cache
     Primer::Classify::Cache.clear!
     Primer::Classify.call(**@values)
 
-    assert_allocations 68 do
+    assert_allocations 69 do
```
joelhawksley, line 45: Interesting. Any ideas why these allocations increased?

camertron: I'm not sure! I ran the benchmarks on ruby 2.7 with the rubocop fixes and they passed, which indicates to me the extra allocations were introduced in ruby 3.0. The numbers also aren't particularly consistent between test runs, but I'm not sure if that's because of ruby 3 or if the benchmarks are inherently "noisy.:

PR #744, opened 2021-08-12
Break out of select loop; publish artifacts in script/release

camertron:
Right now, script/release loops because no `break` exists to escape the `select` command loop 😄

### DIFF
--- a/script/release
+++ b/script/release
@@ -81,10 +81,15 @@ push() {
   git push origin release-v$1-$2-$3
 
   echo "####################################################"
-  echo "Now, open a PR with this branch and merge it to main"
-  echo "Then, run script/publish on main to release the gem and the NPM package"
+  echo "Now, open a PR with this branch and merge it to main."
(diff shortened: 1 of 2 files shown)


script/release
```diff
@@ -159,6 +164,11 @@ main() {
     add_changed_files $major $minor $patch
     commit $major $minor $patch
     push $major $minor $patch
+
+    # publish artifacts
+    script/publish
```
manuelpuyol, line 169: I'm not sure if we want to publish artifacts here since we haven't merged the release PR at this point

camertron: Hmm yeah I see what you mean. The `push` function prints a message asking you to merge the PR, then waits for you to press a key. Is that enough or should we keep the steps separated?

manuelpuyol: I think it is safer to keep them separated

camertron: Ok I'm 🆒 with that 😎

manuelpuyol (approved):

PR #750, opened 2021-08-16
Naming improvements for linters

manuelpuyol:
Renames `Helpers` to `BaseLinter` and updates `classes_to_args` to receive a list of classes instead of a string

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @primer/view-components-reviewers
+* @primer/rails-reviewers
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,16 +78,37 @@ jobs:
   benchmark:
     needs: lint
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rails_version: [5.2.3, 6.1.1, main]
+        ruby_version: [2.5.x, 2.6.x, 2.7.x, 3.0.x]
+        exclude:
+          - rails_version: 5.2.3
+            ruby_version: 2.6.x
+          - rails_version: 5.2.3
+            ruby_version: 2.7.x
+          - rails_version: 5.2.3
+            ruby_version: 3.0.x
+          - rails_version: 6.1.1
+            ruby_version: 2.5.x
+          - rails_version: 6.1.1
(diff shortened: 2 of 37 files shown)


camertron (changes requested):
Hmm, I'm not sure `BaseLinter` communicates what this module does. The comment above it says "Helper methods for linting ERB." Have you considered `ERBLintingHelpers` or maybe `ERBLinting`? In my experience it's more common to see _classes_ named `Base*` and for them to be inherited directly instead of mixed-in.

manuelpuyol:
> Hmm, I'm not sure `BaseLinter` communicates what this module does. The comment above it says "Helper methods for linting ERB." Have you considered `ERBLintingHelpers` or maybe `ERBLinting`? In my experience it's more common to see _classes_ named `Base*` and for them to be inherited directly instead of mixed-in. 😄 😄

hmm maybe the comment should be updated, that module actually holds **all** the linter logic, the classes that include it only have to set some configuration variables
I agree it _should_ be a class, but I had to make it a module because of some `erb_lint` quirks (custom linters have to inherit from `ERBLint::Linters::Base`) so the only way of sharing the logic was through a module :/

going to update the module docs to reflect that change :)

camertron:
@manuelpuyol ah ok that's good to know! Not to completely change the architecture of our linters, but what if `BaseLinter` inherited from `ERBLint::Linters::Base` and each individual linter then inherited from `BaseLinter`?

manuelpuyol:
> @manuelpuyol ah ok that's good to know! Not to completely change the architecture of our linters, but what if `BaseLinter` inherited from `ERBLint::Linters::Base` and each individual linter then inherited from `BaseLinter`?

I've tried that in the past but it didn't work 😢

camertron:
@manuelpuyol shoot 😞  What went wrong?

manuelpuyol:
> @manuelpuyol shoot 😞 What went wrong?

I'll give it a try again, I remember that some configs weren't being applied when subclassing like that
But `erb_lint` received some good updates since then, maybe I can figure it out this time!

manuelpuyol:
@camertron I've updated the PR using inheritance 🎉

camertron (approved):
🎉 🎉 🎉

Issue #717, opened 2021-08-02
[Tracking] Color Modes v2 Support


jonrohan:
This is a tracking issue to track the support of Color Modes V2 coming to primer. https://github.com/github/design-infrastructure/issues/1388

# Todo

- [x] Drop support for presentational deprecated arguments. We currently have [presentational overrides](https://github.com/primer/view_components/blob/223aebfc97276a9489185366f96fa7dbee8dc5ec/lib/primer/classify/functional_background_colors.rb#L28-L40) that need to be cleaned up. These were added in color modes v1 work.
    - [x] Write a rubocop linter to look for old arguments
    - [x] Remove all of these old arguments from dotcom.
        * https://github.com/github/github/pull/187574
        * https://github.com/github/github/pull/187532
- [x] Migrate color classes to use `Classify::Utilities`
- [x] Workarounds for static color options
- [ ] Fix cases in dotcom of static workarounds

PR #752, opened 2021-08-16
Creating rubocop for deprecated arguments

jonrohan:
This PR aims to create a generic rubocop linter for deprecated arguments. We need this work to support https://github.com/primer/view_components/issues/717 which will be replacing color arguments again.

Full information on the project here https://github.com/github/design-infrastructure/issues/1388


.rubocop.yml
```diff
@@ -65,3 +65,15 @@ Lint/MissingSuper:
...
+  Enabled: true
+  Deprecated:
+    color:
+      blue: ":link"
+      gray_dark: ":primary"
+      gray: ":secondary"
+      gray_light: ":tertiary"
+      green: ":success"
+      yellow: ":warning"
+      red: ":danger"
```
jonrohan, line 79: I thought this was a good place to put the arguments this way it can be configured by consumers.

docs/content/deprecated.md
```diff
@@ -0,0 +1,19 @@
...
+
+| argument | value | replacement |
+| :- | :- | :- |
+| `color:` | `:blue` | `:link` |
+| `color:` | `gray_dark:` | `:primary` |
+| `color:` | `gray:` | `:secondary` |
+| `color:` | `gray_light:` | `:tertiary` |
+| `color:` | `green:` | `:success` |
+| `color:` | `yellow:` | `:warning` |
+| `color:` | `red:` | `:danger` |
```
jonrohan, line 19: I wanted to create a url we could link to with more instructions https://view-components-git-linterfordeprecatingarguments-primer.vercel.app/view-components/deprecated

I tried to leave this in a format like this so we could repeat for other things like components, arguments.

```
Version
- Types of deprecations
   - Table of deprecations
Version
- Types of deprecations
   - Table of deprecations
Version
- Types of deprecations
   - Table of deprecations
```

.rubocop.yml
camertron: Is this something we want consumers to be able to configure? I'm just a bit concerned about requiring teams to copy/paste mappings into their configs. Seems like we might end up with a hodgepodge of correct and incorrect mappings in various repos 🤔

camertron (changes requested):
Super cool! 🎉  Just the one comment.

.rubocop.yml
jonrohan: I think most consumers would extend this config so they'll get what we have listed anyways, but will have the ability to change if it's taking them longer to migrate.

camertron: Hmm ok I see what you mean. Sounds good to me 👍

camertron (approved):

lib/rubocop/config/default.yml
```diff
@@ -12,3 +12,15 @@ Primer/NoTagMemoize:
...
+  Enabled: true
+  Deprecated:
+    color:
+      blue: ":link"
+      gray_dark: ":primary"
+      gray: ":secondary"
+      gray_light: ":tertiary"
+      green: ":success"
+      yellow: ":warning"
+      red: ":danger"
```
manuelpuyol, line 26: ❤️

lib/rubocop/cop/primer/deprecated_arguments.rb
```diff
@@ -0,0 +1,69 @@
...
+          kwargs = node.arguments.last
+
+          return unless kwargs.type == :hash
+
+          @deprecated_arguments = cop_config["Deprecated"] || {}
+
+          kwargs.pairs.each do |pair|
+            # Skip if we're not dealing with a symbol
+            next if pair.key.type != :sym
+            next if pair.value.type != :sym
```
manuelpuyol, line 38: could we support strings like `color: "blue"`? Or do we never support strings for colors?

lib/rubocop/cop/primer/deprecated_arguments.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        # We only verify SystemArguments if it's a `.new` call on a component or
+        # a ViewHleper call.
+        def valid_node?(node)
+          view_helpers.include?(node.method_name) || (node.method_name == :new && ::Primer::ViewComponents::STATUSES.key?(node.receiver.const_name))
+        end
+
+        def view_helpers
+          ::Primer::ViewHelper::HELPERS.keys.map { |key| "primer_#{key}".to_sym }
+        end
```
manuelpuyol, line 65: don't have to do it now, but it'd be nice to extract this logic to a module that's included here and in the SysArgs cop

test/rubocop/deprecated_arguments_test.rb
```diff
@@ -0,0 +1,65 @@
...
+    assert_equal 2, cop.offenses.count
+  end
+
+  def test_deprecated_argument_autocorrected
+    investigate(cop, <<-RUBY)
+      Primer::BaseComponent.new(foo: :deprecated)
+    RUBY
+
+    assert_equal 1, cop.offenses.count
+    assert_equal "Primer::BaseComponent.new(foo: :new_argument)", cop.offenses.first.corrector.rewrite.strip
```
manuelpuyol, line 54: ohhh that's so cool, I wasn't able to figure out how to test corrections!

manuelpuyol (approved):

lib/rubocop/cop/primer/deprecated_arguments.rb
jonrohan: Was thinking the same, we need a rubocop helpers class

test/rubocop/deprecated_arguments_test.rb
jonrohan: Took a bunch of irb probing to figure it out. 😅

docs/content/deprecated.md
```diff
@@ -0,0 +1,19 @@
...
+## Deprecated in v0.0.28
+
+### System Arguments
+
+These presentational system arguments were deprecated in favor of functional color names.
+
+| argument | value | replacement |
+| :- | :- | :- |
+| `color:` | `:blue` | `:link` |
+| `color:` | `gray_dark:` | `:primary` |
```
joelhawksley, line 14:
```suggestion
| `color:` | `:gray_dark` | `:primary` |
```

etc

docs/content/deprecated.md
joelhawksley: @camertron @manuelpuyol what do you think about adding the CHANGELOG to the docs site and just publishing this as part of the release notes? We do something similar for ViewComponent.

.rubocop.yml
joelhawksley: 🤔 I have a similar hesitation to @camertron. My vote would be to hardcode these mappings inside the linter for now and to revisit configuration later as an added feature.

docs/content/deprecated.md
camertron: Hmm not sure I understand, can you expand a little more on what that would mean? 🤔

.rubocop.yml
jonrohan: ok, I've moved them there 👍🏻

PR #754, opened 2021-08-17
PrimerOcticon linter supports `aria-` and `data-` attributes.

manuelpuyol:
The linter will autocorrect cases that are using `aria-` and `data-` attributes. It will also convert `data-test-selector` to `test_selector`

### DIFF
--- a/lib/rubocop/cop/primer/primer_octicon.rb
+++ b/lib/rubocop/cop/primer/primer_octicon.rb
@@ -27,8 +27,12 @@ module RuboCop
           Replace the octicon helper with primer_octicon. See https://primer.style/view-components/components/octicon for details.
         STR
 
-        SIZE_ATTRIBUTES = %i[height width size].freeze
-        VALID_ATTRIBUTES = [*SIZE_ATTRIBUTES, :class].freeze
+        SIZE_ATTRIBUTES = %w[height width size].freeze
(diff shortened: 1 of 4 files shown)


test/cop_test.rb
```diff
@@ -21,4 +21,9 @@ def investigate(cop, src, filename = nil)
     commissioner.investigate(processed_source)
     commissioner
   end
+
+  def assert_correction(correction)
+    refute_empty cop.offenses
+    assert_equal correction, cop.offenses.first.corrector.rewrite.strip
+  end
```
jonrohan, line 28: 🤘🏻

lib/rubocop/cop/primer/primer_octicon.rb
```diff
@@ -142,6 +145,19 @@ def size_attributes_to_string(size_attributes)
...
+
+          args = kwargs.pairs.each_with_object([]) do |pair, acc|
+            next unless pair.key.value.to_s.match?(STRING_ATTRIBUTE_REGEX)
+
+            key =  pair.key.value.to_s == "data-test-selector" ? "test_selector" : "\"#{pair.key.value}\""
+            acc << "#{key}: #{pair.value.source}"
+          end
+
+          args.join(",")
+        end
```
jonrohan, line 159: Another method that might be good in a rubocop helper class

jonrohan (approved):

PR #757, opened 2021-08-17
Don't run benchmarks against Ruby 3 for now

camertron:
Unfortunately adding Ruby 3 support seems to have broken the benchmark CI jobs. The number of allocations doesn't appear to be "stable," i.e. can change between test runs. I'm going to spend some time this week looking into it, but for now let's stop running benchmarks on Ruby 3 so our CI isn't broken.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,14 +81,12 @@ jobs:
     strategy:
       matrix:
         rails_version: [5.2.3, 6.1.1, main]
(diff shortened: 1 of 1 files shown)


jonrohan:
Is this the same fix as https://github.com/primer/view_components/pull/753 ?

camertron:
@jonrohan I don't think so? #753 includes ruby 3.0 in the list of ruby versions.

camertron:
Oh wait, I missed that version-specific code. I suppose it is the same, yeah.

PR #759, opened 2021-08-18
Pulling in text color utilities from primer/css and removing functional_text_colors class

jonrohan:
The purpose of the PR is to move color arguments into `Primer::Classify::Utilities` I'm pulling these files directly from primer/css like the rest of the utilities.

One change to the way Utilities work in this PR is now, when an unsupported value in production is used, instead of returning an empty string, I'm returning a dasherized `key-value` class for the argument. This is important as it lets us still support deprecated classes while moving forward with the library.


test/lib/classify_test.rb
```diff
@@ -159,81 +159,15 @@ def test_underline
...
-    assert_generated_class("text-inherit",       { color: :inherit })
-
-    assert_generated_class("color-blue-5",       { color: :blue_5 })
-    assert_generated_class("color-gray-9",       { color: :gray_9 })
-    assert_generated_class("color-purple-3",     { color: :purple_3 })
-
     assert_generated_class("color-text-inverse",   { color: :text_inverse })
-
     assert_generated_class("color-text-primary",   { color: :text_primary })
-    assert_generated_class("color-text-primary",   { color: :gray_dark })
```
joelhawksley, line 215: So just to be clear, we're removing this fallback?

jonrohan: The fallbacks will still work because of this change https://github.com/primer/view_components/pull/759/files#diff-03a91cc7818b023b5a2caa70ded0df6ebbb735a0cd9c3187c487c5d568e9a7beR178

The plan is:

1. Deprecate argument in linter
2. Ship linter change out to dotcom
3. remove support from view components
4. Move old classes to dotcom from primer/css
5. Cleanup classes and delete old css

joelhawksley: Makes sense. If the fallbacks still work, can we keep this test in place?

jonrohan: They don't

joelhawksley (approved):

manuelpuyol:
@jonrohan may we fix the conflict so we can merge this? 😄

PR #760, opened 2021-08-18
Update linters override configuration

manuelpuyol:
Adds a new configuration to linters `override_ignores_if_correctable` that will decide whether or not to override the `erblint:counter XYZ N` if there is a correctable case in the file.

This change is nice to have for future updates on linters, so we can ship a new version of PVC in dotcom without having to fix all the new cases that autocorrection covers.


jonrohan (approved):
Nice improvement

PR #762, opened 2021-08-19
Map `for` argument in ClipboardCopy migration linter

koddsson:
The `ClipboardCopyComponetMigrationCounter` linter would fail to lint `<clipboard-copy>` elements with a `for` argument. The reason is that the `for` argument isn't mapped at all. Only `value` is mapped.

This PR adds the `for` argument to the argument mapper for ClipboardCopy so that the linter will auto fix instances of `<clipboard-copy for>`.


manuelpuyol (approved):
🎉

PR #763, opened 2021-08-19
Extract shared linter tests

manuelpuyol:
Most of the tests were exactly the same, so I extracted `BasicLinterSharedTests` and `AutocorrectableLinterSharedTests` that will test shared behaviors between linters


camertron (approved):
Very cool! 🎉

PR #764, opened 2021-08-19
Stabilize allocation counts

camertron:
This PR makes allocation counting more stable, as test order can affect the numbers and lead to flaky tests. I also refactored things a little to move the allocation counts per Ruby version inside the test.

Worryingly, it appears Ruby 3 allocates more objects in every benchmark than Ruby 2.7. In the classify without cache case, Ruby 3 produces _many_ more allocations, 127 vs 70 😱  Accordingly, I'm going to look into improving classify.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,7 +169,7 @@ jobs:
       uses: actions/upload-artifact@master
       if: always()
       with:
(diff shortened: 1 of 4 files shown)


jonrohan (approved):
> 127 vs 70 😱 Accordingly, I'm going to look into improving classify.

yikes!

PR #769, opened 2021-08-23
rename BreadcrumbComponent to Breadcrumbs

joelhawksley:
This PR renames `BreadcrumbComponent` to `Breadcrumbs`, per our decision to use the plural form to better reflect the fact that this component almost always displays multiple items.


app/components/primer/breadcrumbs.rb
```diff
@@ -2,7 +2,7 @@
 
 module Primer
   # Use `Breadcrumb` to display page hierarchy within a section of the site. All of the items in the breadcrumb "trail" are links except for the final item, which is a plain string indicating the current page.
-  class BreadcrumbComponent < Primer::Component
+  class Breadcrumbs < Primer::Component
```
manuelpuyol, line 5: since we are already doing a name change, wdyt of adding the `Beta` module?

joelhawksley:
> So, this means `BreadcrumbComponent` will no longer be available to, for example, dotcom. Is that ok? What's the upgrade path?

@camertron that's correct. Whoever updates PVC in dotcom will need to rename the existing uses of the component.

PR #777, opened 2021-08-26
Add support for color utilities in linters

manuelpuyol:
This won't affect how system arguments generate classes, so it still supports deprecated colors. But the utilities change will allow linters to know how to convert classes like `color-text-primary` into arguments. This will enable tons of migrations without breaking the user API.

I've also added an extra migration for the PrimerOcticon linter, which will convert `color-text-x` classes into their respective `color: :icon_x` argument


manuelpuyol:
Closing in favor of https://github.com/primer/view_components/pull/759

PR #778, opened 2021-08-29
Add `break_word` as an option for the `word_break` system argument

antn:
When working on https://github.com/github/communities/issues/741, I noticed that we didn't have the ability to specific `break_word` with `word_break`, even though it's part of Primer.

This PR adds that, and also fixes a typo in CONTRIBUTING.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,9 +30,9 @@ The category for changes related to documentation, testing and tooling. Also, fo
 
 ## main
 
-### New
+### Updates
 
-* Add `break_word` as an option for the `word_break` system argument.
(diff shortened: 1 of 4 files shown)


jonrohan (changes requested):
The utilities.yml file gets generated on build, and will be overwritten when edited manually. The root of the problem is how we named the `.break-word` class in primer doesn't match naming convention. This PR should fix that problem https://github.com/primer/css/pull/1555 and next time we bump the primer/css version in this library it will pull in the new class.

antn:
Ah, thanks! Closing this then!

PR #785, opened 2021-09-02
Split `TabNav` into `TabNav` and `TabPanels`

### LINKED ISSUE #680: Update tab navigation API and make accessibility fixes

khiga8:
Closes #680 

### What
- Splits `TabNav` into `TabNav` and `TabPanels` to simplify API and tailors documentation for each component.

### Why
`TabNav` API is getting complex with markup and behavioral changes that are based on the single `with_panel` attribute. This makes appropriate documentation difficult.

### Related
- https://github.com/github/primer/issues/265 - something to keep in mind.

### Discussion
Something we should follow up separately:
- we currently support `align: left` and `align: right` to place the extra content on left or right of the tabs in the DOM. However, with the default `align: left`, the styling is off and the examples show it set with`float-right` which visually makes it the same as `align: right`. What are the usecases for this?


manuelpuyol:
> we currently support align: left and align: right to place the extra content on left or right of the tabs in the DOM. However, with the default align: left, the styling is off and the examples show it set withfloat-right which visually makes it the same as align: right. What are the usecases for this?

`align` is a terrible name (sorry about that), the goal of that attribute is only to change the position of the element in the HTML code, which changes the priority on tab navigation

manuelpuyol (approved):

PR #787, opened 2021-09-03
Stabilize Ruby 3 allocations

BlakeWilliams:
I don't have a great understanding of this, but `GC.compact` likely
invalidates caches which seem to be causing our Ruby 3.0 allocations to
appear larger than they actually are.

This change removes the call to `GC.compact` which drops our allocation
numbers. I think the new numbers should be closer to what we expect when
using these classes in an application.

Thanks @jhawthorn for pointing this out!

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,6 +17,8 @@ Rake::TestTask.new(:test) do |t|
 end
 
 Rake::TestTask.new(:bench) do |t|
+  # Disable parallel tests since it can interfere with benchmark results and allocation counts
+  ENV["MT_CPU"] = "0"
   t.libs << "test"
   t.test_files = FileList["test/benchmarks/**/bench_*.rb"]
   t.verbose = true
--- a/test/benchmarks/bench_classify.rb
+++ b/test/benchmarks/bench_classify.rb
@@ -38,11 +38,8 @@ class BenchClassify < Minitest::Benchmark
   end
 
   def bench_allocations_without_cache
-    # Preload cache, warm up benchmark
-    Primer::Classify::Cache.clear!
-    Primer::Classify.call(**@values)
-
-    assert_allocations "3.0" => 127, "2.7" => 68, "2.6" => 69, "2.5" => 70 do
+    assert_allocations "3.0" => 68, "2.7" => 68, "2.6" => 69, "2.5" => 70 do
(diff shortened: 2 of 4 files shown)


camertron:
Hey @BlakeWilliams thanks for taking this on! I actually added the `GC.compact` call in recently because without it our benchmark CI job would fail intermittently (which appears to have happened to this PR 😓 ). From my testing I discovered allocation counting can be pretty non-deterministic, even with GC disabled. Adding in `GC.compact` made our test suite pass 100% of the time, although I'll admit I have no idea why. Maybe a way forward would be to pass a flag or environment variable to skip running it? Are you guys getting stable numbers?

BlakeWilliams:
> Hey @BlakeWilliams thanks for taking this on! I actually added the `GC.compact` call in recently because without it our benchmark CI job would fail intermittently (which appears to have happened to this PR 😓 ). From my testing I discovered allocation counting can be pretty non-deterministic, even with GC disabled. Adding in `GC.compact` made our test suite pass 100% of the time, although I'll admit I have no idea why. Maybe a way forward would be to pass a flag or environment variable to skip running it? Are you guys getting stable numbers?

Hmm, I wonder why compacting made it consistent. 🤔 

It's been very consistent locally and I've been unable to make it fail, so I'm not sure what is different between the CI environment and my local environment besides OS.

Rakefile
```diff
@@ -17,6 +17,8 @@ Rake::TestTask.new(:test) do |t|
 end
 
 Rake::TestTask.new(:bench) do |t|
+  # Disable parallel tests since it can interfere with benchmark results and allocation counts
+  ENV["MT_CPU"] = "1"
```
BlakeWilliams, line 21: Minitest was parallelizing our benchmarks by default. This was the source of the extra random allocations.

https://github.com/seattlerb/minitest/blob/master/lib/minitest.rb#L27
https://github.com/seattlerb/minitest/blob/master/lib/minitest.rb#L138-L145

BlakeWilliams:
Closing this out in favor of https://github.com/primer/view_components/pull/789

camertron:
Hey @BlakeWilliams, how would you feel about keeping the `MT_CPU = 0` line for benchmark tests? Seems like potentially a bad idea to run our benchmarks in parallel, thoughts?

PR #789, opened 2021-09-03
Use the allocation_stats gem to count object allocations in our benchmarks

camertron:
@BlakeWilliams discovered that `GC.compact` likely invalidates internal Ruby caches, so I spent some time looking for a better way to count object allocations. This PR uses the [allocation_stats gem](https://github.com/srawlins/allocation_stats) which uses the internal `ObjectSpace.trace_object_allocations` method available since Ruby 2.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ The category for changes related to documentation, testing and tooling. Also, fo
 
     *Kate Higa*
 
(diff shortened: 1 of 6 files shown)


BlakeWilliams (approved):
Nice!

It might be good to re-run the tests a few times in CI, just in case but this looks great!

camertron:
Re-ran 'em a few times and all is well 🎉

PR #790, opened 2021-09-03
Make octicon cache key construction more performant

camertron:
We currently use an intermediate array and two strings (the dash and the resulting key string) to construct the key for the octicon cache. I tried using `Hash#hash` instead and got almost a 2.5x speedup and a ~16% memory reduction 😄


BlakeWilliams (approved):
Nice find!

Is it possible to share the output of the benchmarks?

camertron:
Sure! Here are the results of the speedup via benchmark-ips:

```
         get_key_old    62.735k i/100ms
             get_key   144.985k i/100ms
         get_key_old    601.260k (± 5.7%) i/s -      3.011M in   5.025841s
             get_key      1.438M (± 5.0%) i/s -      7.249M in   5.055706s

Comparison:
             get_key:  1437699.7 i/s
         get_key_old:   601259.7 i/s - 2.39x  (± 0.00) slower
```

I used allocation_stats to measure memory usage. Total allocated memory for 10 `Primer::OcticonComponent.new(icon: :alert)`s went from 15440 bytes to 11680 bytes. Here's the script I used:

```ruby
stats = AllocationStats.new(burn: 3).trace do
  10.times do
    Primer::OcticonComponent.new(icon: :alert)
  end
end

table = stats
  .allocations(alias_paths: true)
  .group_by(:sourcefile, :sourceline, :class)
  .to_text

puts table

total_bytes = stats.allocations.all.inject(0) do |memo, stat|
  memo + stat.memsize
end

puts "Total: #{total_bytes} bytes"
```

jonrohan (approved):
WOW! Nice improvments

camertron:
Thanks @manuelpuyol :)

PR #792, opened 2021-09-08
Autocorrect buttons with href, name, value and tabindex

manuelpuyol:
Those are simple attributes that are being used with `buttons` in dotcom and autocorrecting them should be safe


test/linters/button_component_migration_counter_test.rb
```diff
@@ -99,6 +99,22 @@ def test_autocorrects
     assert_equal expected, corrected_content
   end
 
+  def test_autocorrects_attributes
+    @file = <<~HTML
+      <button class="btn" href="href" value="value" name="name" tabindex="tabindex">
```
joelhawksley, line 104: Is a button with an `href` valid? I thought we tried to avoid that combination.

manuelpuyol: buttons can be `<a>`, so they have `hrefs`

camertron (approved):

PR #793, opened 2021-09-08
`Primer::Breadcrumbs` requires `href`s for all items

joelhawksley:
(... and no longer accepts `selected`)

As part of our work to move `Breadcrumbs` to `stable`, this PR
makes API changes suggested during our accessibility audits.


manuelpuyol (commented):
I think we need to update the story 😄

manuelpuyol (approved):

PR #800, opened 2021-09-10
Removing validation prefixes that are handled by the Utilities class

jonrohan:
This should have been done in https://github.com/primer/view_components/pull/759

In the validation class we're checking if a prefix is in the supplied classname (old way) OR if the Utilities class has the specific classname in it's data.

Now that the more accurate checking is occurring we shouldn't be throwing errors on all classnames with `color-` prefix.


lib/primer/classify/validation.rb
```diff
@@ -6,7 +6,7 @@ module Primer
   class Classify
     # :nodoc:
     class Validation
-      INVALID_CLASS_NAME_PREFIXES = /bg-|color-|text-|box-shadow-|text-|box_shadow-/.freeze
```
manuelpuyol, line 9: could this affect `color-bg-*`?

jonrohan: Fair point, I can put that back in

camertron (approved):

PR #801, opened 2021-09-10
Pulling in the bg color utilities from primer

jonrohan:
Follow up to https://github.com/primer/view_components/pull/759

This pr, refactors the `bg:` key to pull in the utility classes from primer and populate the utility file. This does include a breaking change, where we're no longer accepting `#000000` hex colors as arguments. https://github.com/github/github/pull/193583 will address the places it's being used.

Similar to the other PR, we will need to refactor the places still using the deprecated values. I'll try and send a pr before this time though.


lib/primer/classify.rb
```diff
@@ -167,12 +165,6 @@ def extract_value(memo, key, val, breakpoint)
           BOOLEAN_MAPPINGS[key][:mappings].each do |m|
             memo[:classes] << m[:css_class] if m[:value] == val && m[:css_class].present?
           end
-        elsif key == BG_KEY
-          if val.to_s.start_with?("#")
-            memo[:styles] << "background-color: #{val};"
-          else
```
jonrohan, line 173: I was going to still support this at the top, but I did a search on dotcom and this is used less than 20 times. I think I'd rather drop the support and convert `bg: "#000"` to `style: "background-color: #000;"`

jonrohan: https://github.com/github/github/pull/193583 will take care of this issue

joelhawksley (approved):

PR #806, opened 2021-09-15
Remove copy as this isn't supported anymore.

maximedegreve:
Before we did support setting an integer on the `w` and `h` props but this isn't supported anymore.


jonrohan (approved):
🙌🏻 THANKS!

🚀

camertron (approved):

manuelpuyol (approved):

PR #807, opened 2021-09-16
Improve performance of the Classify module

camertron:
The `Classify` module, specifically `Classify.call`, is one of the hottest codepaths in the framework. This PR introduces two classes of improvements: 1) dumb Ruby tricks, and 2) more aggressive caching.

### Dumb Ruby tricks
1. Avoid calling `Object#presence` and `Object.blank?`. While ergonomic, they incur method call overhead and are generally much slower than `!obj || obj.empty? ? nil : obj`.
1. In the same vein, `!obj.nil?` is faster than `!!obj`, which is faster than `obj.present?`.
1. Mutate strings with methods like `squeeze!` and `strip!` instead of allocating new strings with their non-bang counterparts.
1. Cache simple lookup operations with hashes with default procs.

### More aggressive caching

We currently preload a bunch of common Primer classes in `Primer::Classify::Cache`, but not all the possibilities are covered. Given that there aren't _that_ many class mappings, it seems like a good idea to cache the results of every result that passes through `Classify.call`.

### Miscellaneous

1. There's no need to call `VALID_KEYS.include?` in `extract_css_attrs` (nee `extract_hash`) because `classes_from` (nee `extract_value`) returns `nil` if the key isn't valid.

### Results

With a preloaded and warmed cache, this PR results in 11 fewer allocations per call to `Classify.call` and a ~2.4x speed improvement. I can't actually run the benchmarks anymore since I removed the old code, but there are the results of running the following benchmark on main vs this branch:

```ruby
args = { bg: :blue_5, color: :text_warning, vertical_align: :text_bottom, display: :block, mx: 4, mb: 4 }

Benchmark.ips do |x|
  x.report do
    Primer::Classify.call(args)
  end
end
```

|Branch|IPS|
|-|-|
|Main| 96.293k (± 5.3%) i/s -    482.950k in   5.030905s|
|This branch|230.319k (± 5.5%) i/s -      1.155M in   5.030947s|

230.319 / 96.293 = 2.392x speedup


lib/primer/classify.rb
```diff
@@ -94,19 +94,18 @@ class Classify
...
-          style
-        ].compact.join.presence
-
-        extracted_results
+        extract_css_attrs(args).tap do |extracted_results|
+          classes = +"#{validated_class_names(classes)} #{extracted_results[:class]}"
+          classes.strip!
+          # rubocop:disable Rails/Blank
+          # do this instead of using presence/blank?, which are a lot slower
+          extracted_results[:class] = !classes || classes.empty? ? nil : classes
```
manuelpuyol, line 102: for the sake of readability, I wonder if we could create our "presence" method instead of repeating the logic

```rb
def presence(obj)
  !obj || obj.empty? ? nil : classes
end
```

lib/primer/classify/cache.rb
```diff
@@ -9,107 +9,124 @@ module Primer
...
-      # rubocop:disable Style/MutableConstant
-      LOOKUP = {}
-      # rubocop:enable Style/MutableConstant
-
-      class <<self
-        def read(memo, key, val, breakpoint)
-          value = LOOKUP.dig(breakpoint, key, val)
-          memo[:classes] << value if value
-        end
+      include Singleton
```
manuelpuyol, line 12: I'm not really familiar with Singletons in Rails, since most apps run multiple instances, will each instance have a different cache?

lib/primer/classify.rb
```diff
@@ -94,19 +94,18 @@ class Classify
...
-
-        extracted_results[:style] = [
-          extracted_results.delete(:styles),
-          style
-        ].compact.join.presence
-
-        extracted_results
+        extract_css_attrs(args).tap do |extracted_results|
+          classes = +"#{validated_class_names(classes)} #{extracted_results[:class]}"
+          classes.strip!
```
manuelpuyol, line 99:
```suggestion
          classes = "#{validated_class_names(classes)} #{extracted_results[:class]}".strip
```

I _think_ that running strip here will remove the need of `+"`

lib/primer/classify.rb
```diff
@@ -133,80 +136,94 @@ def validated_class_names(classes)
...
         end
 
-        memo[:classes] = memo[:classes].join(" ")
+        {
+          class: classes.join(" "),
+          style: styles
+        }
+      end
+
+      def extract_css_attrs_into(classes, styles, key, val, breakpoint)
```
manuelpuyol, line 165: reading this method I asked myself "into what?"

manuelpuyol (commented):
nice work!

BlakeWilliams (approved):
Looks great!

Nice performance improvement!

lib/primer/classify.rb
camertron: Ah yeah I know what you mean. I was trying to convey that `classes` and `styles` would be mutated as a side-effect of calling the method. Is there a better name that conveys the same idea or should I just get rid of the "into"?

lib/primer/classify.rb
camertron: Yeah, I was trying to avoid the extra allocation.

lib/primer/classify/cache.rb
camertron: That's a good question. I believe we run Unicorn in production, which preloads the app and then forks into child processes. Any time the program writes to a shared place in memory (i.e. changes a variable), the memory is copied, then mutated (this is known as copy-on-write, or CoW). Since it's very likely the singleton instance will get mutated _after_ forking, each process will make a copy of the original `@lookup` hash. If my mental model is correct, that's the same as what happened when we used the `LOOKUP` constant.

camertron: We could make this explicit by implementing our own singleton logic that could store the instance per-thread (eg. per-process) in `Thread.current`. I actually tried that earlier this week and found that it negatively affected performance.

lib/primer/classify.rb
camertron: Ordinarily I'd agree with you, but I'd like to avoid any additional method call overhead. Lemme put it in tho and see how the numbers look.

lib/primer/classify.rb
manuelpuyol: I think removing the `into` would be fine

lib/primer/classify.rb
camertron: Looks like the additional method adds a negligible amount of overhead 🎉

PR #808, opened 2021-09-17
Add `gem "view_component"`

maximedegreve:
While setting up a brand new rails project and [template](https://github.com/primer/view_components-template) I've noticed that this gem wouldn't work without `view_component`. I'm not sure if this is a bug where the dependency should come with `primer_view_components` instead because I have no experience with Rails.

(I've accidentally committed this directly on the main branch initially but removed it again. 🥴)

### DIFF
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -21,6 +21,7 @@ Render Primer ViewComponents from templates:
 In `Gemfile`, add:
 
 ```ruby
+gem "view_component"
 gem "primer_view_components"
 ```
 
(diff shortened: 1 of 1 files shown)


manuelpuyol:
installing `primer_view_components` should install `view_component` automatically as a dependency for you 🤔 
maybe you just need to `require` it in your app?

maximedegreve:
Oh I will give that a try @manuelpuyol and change the docs accordingly!

maximedegreve:
@manuelpuyol This indeed fixed my issue 🎉 Also updated the docs to mention this now. Thanks for your help.

manuelpuyol (approved):

PR #812, opened 2021-09-20
Fix `ClipboardCopy` octicons not toggling correctly after first click

manuelpuyol:
Make sure to delete the timer after it runs. This was causing the toggle behavior to be stuck in a bad state.
Also refactors the code to be more clear


joelhawksley (approved):
Is there any way for us to write a test for this behavior?

jonrohan (approved):

PR #813, opened 2021-09-20
restrict usage of padding and font size on Breadcrumbs

joelhawksley:
This PR does two things:

1) It removes the extra `breadcrumb-item-selected` that was necessary before https://github.com/primer/css/pull/1590 landed.
2) Restricts the usage of `padding` and `font_size` on Breadcrumbs.
3) Adds a linter for the component.


app/components/primer/beta/breadcrumbs.rb
```diff
@@ -6,6 +6,9 @@ module Beta
     class Breadcrumbs < Primer::Component
       status :beta
 
+      PADDING_MESSAGE = "Padding system arguments are not allowed on Breadcrumbs. Consider using margins instead."
+      FONT_SIZE_MESSAGE = "Breadcrumbs does not support the font_size system argument."
```
camertron, line 10:
```suggestion
      FONT_SIZE_MESSAGE = "Breadcrumbs do not support the font_size system argument."
```

app/components/primer/beta/breadcrumbs.rb
```diff
@@ -22,6 +25,10 @@ def initialize(**system_arguments)
         @system_arguments = system_arguments
         @system_arguments[:tag] = :nav
         @system_arguments[:aria] = { label: "Breadcrumb" }
+        @system_arguments[:system_arguments_denylist] = {
```
camertron, line 28: Can this be moved into a constant? Same with the aria hash above.

camertron (approved):

PR #820, opened 2021-09-21
Use `copy` instead of `paste` icon in ClipboardCopy

colebemis:
## Problem
From https://github.com/primer/octicons/issues/606

> Looking at `copy` actions throughout GitHub, it uses a clipboard icon clippy. However, a clipboard is used for representing the paste action, not, not copying. I think Octicons should have a more common two-sheets-of-paper metaphor for copy actions.

## Solution

This PR swaps the `paste` icon for the `copy` icon in the ClipboardCopy component:

| Before | After | 
| --- | --- |
| CleanShot 2021-09-21 at 09 44 56@2x | CleanShot 2021-09-21 at 09 42 43@2x |

---
cc @ashygee


joelhawksley:
@colebemis I'm going to assume that we'd want to switch all clipboard copy's to use this new icon and not just places that use the ViewComponent, right?

colebemis:
@joelhawksley Yup. I'm working on a PR to replace the `paste` icon in dotcom as well 👍

app/components/primer/clipboard_copy.html.erb
```diff
@@ -2,7 +2,7 @@
   <% if content.present? %>
     <%= content %>
```
joelhawksley, line 3: We'll need to check for cases where we pass in content that uses the paste icon.

joelhawksley (approved):

app/components/primer/clipboard_copy.html.erb
colebemis: Would you like me to do that in this PR?

joelhawksley: Sorry, I meant when we go to add this change to dotcom. I just wanted to call our that there may be uses of this component that pass in the `paste` icon as content.

PR #823, opened 2021-09-24
Adding octicon test with title attribute

jonrohan:
I was noticing that the linter wasn't replacing octicon calls with a title attribute. ie.

```
octicon(:icon, title: "icon")
```

Adding a test in this PR and going to attempt to fix it.


camertron (approved):
🎉

PR #833, opened 2021-10-05
Add Stats class

manuelpuyol:
I'm proposing adding a `Stats` class that can provide users with some metrics related to PVC. The first metric I'm implementing here is the amount of components with the `@accessibility` tag on their docs.
If y'all think it doesn't make sense to have those numbers in PVC, I can reimplement something similar in dotcom, since we'll need to report that number to keep track of it

### DIFF
--- a/lib/primer/view_components/stats.rb
+++ b/lib/primer/view_components/stats.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module Primer
+  # :nodoc:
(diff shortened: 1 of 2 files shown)


jonrohan:
I like this idea, and having it in the gem, will make it easier to collect stats in the report action https://github.com/github/github/blob/master/.github/workflows/report_code_stats.rb

camertron (approved):

PR #843, opened 2021-10-12
Aggressively cache classes

camertron:
`Primer::Classify.call` is still one of the hottest codepaths (from this project) on github.com. We currently preload a bunch of common classes into the classify cache, but we're still going through `extract_css_attrs` and `validated_class_names` on cache misses (although the latter is disabled in production). Even with the caching added in #807, we're still doing too much unnecessary work IMHO.

I got to thinking a few days ago that classify is probably most often called with a bounded set of arguments, since components are most often initialized with a static set of classes. There's not much point performing the exact same calculations over and over again, so why not cache entire sets of classes?

This PR implements such a caching strategy. Obviously we don't want the cache to grow unbounded, so I thought we should probably use an LRU (Least Recently Used) cache that kicks out old entries if the cache gets too big. It's not clear what size limit we should use for github.com, but I'm in the process of semi-scientifically coming up with that number.

### Results

```ruby
args = {
  bg: :success, color: :text_warning, vertical_align: :text_bottom, display: :block, mx: 4, mb: 4
}

Benchmark.ips do |x|
  x.report do
    Primer::Classify.call(**args)
  end
end
```

|Branch|IPS|
|-|-|
|Main|215.139k (± 3.6%) i/s -      1.081M in   5.029640s|
|This branch|539.161k (± 4.5%) i/s -      2.743M in   5.097681s|

539.161k / 215.139 = 2.5x speedup

Combined with #807 that's a ~5x speed increase from the original implementation.


BlakeWilliams:
This looks good and makes sense. I'm curious how we plan to measure the results in production since it depends even more on cache hits/misses now (especially since misses are now more expensive). We had spiked out something similar to this at one point but we weren't certain about the hit/cache ratio and decided to go the existing cache route. Maybe we could add some logging or emit some metric for cache hits/misses?

Given that, I'm also wondering if it would make sense to keep the old caching logic in place too. I don't think it would do much for the benchmark, but it could still prove to be beneficial for cache misses.

camertron:
@BlakeWilliams yeah, I'm also a bit concerned about hits/misses. Is there a good mechanism we already have for emitting metrics?

Yeah I'm happy to add the old caching logic back in.

lib/primer/classify/attr_cache.rb
```diff
@@ -0,0 +1,115 @@
...
+      def fetch(breakpoint, key, val)
+        found = @lookup.dig(breakpoint, key, val)
+        return found if found
+
+        yield.tap do |result|
+          set(result, breakpoint, key, val) if @cache_enabled
+        end
+      end
+
+      def disable
```
BlakeWilliams, line 27: Less of an immediate issue and more of something to be aware of, but this isn't thread-safe. Should we document that somewhere, maybe as part of some basic module docs?

test/benchmarks/bench_classify.rb
```diff
@@ -39,31 +39,25 @@ def setup
...
 
-  def bench_allocations_with_cache_preloaded
-    Primer::Classify::Cache.instance.clear!
-    Primer::Classify::Cache.instance.preload!
-
-    assert_allocations "3.0" => 42, "2.7" => 35, "2.6" => 38, "2.5" => 39 do
-      Primer::Classify.call(**@values)
-    end
-  end
+  def bench_allocations_with_cache_warmed
```
BlakeWilliams, line 53: Since we know have two forms of caching (preloading + an LRU cache) should there be two separate tests?

test/lib/classify/cache_test.rb
```diff
@@ -5,9 +5,60 @@
...
+
+    cache.fetch(:boo) { :boo }
+
+    assert_equal cache.size, 3
+    assert_includes lookup, [:bar].hash
+    assert_includes lookup, [:baz].hash
+    assert_includes lookup, [:boo].hash
+    refute_includes lookup, [:foo].hash
+
+    cache.max_size = Rails.application.config.primer_view_components.max_classify_cache_size
```
BlakeWilliams, line 38: Thoughts on extracting this into a setup+teardown?

```ruby
def setup
  @original_max_size = Rails.application.config.primer_view_components.max_classify_cache_size
end

def teardown
  Rails.application.config.primer_view_components.max_classify_cache_size = @original_max_size 
end
```

test/lib/classify/cache_test.rb
```diff
@@ -5,9 +5,60 @@
...
 
-    assert_empty Primer::Classify::Cache.instance.instance_variable_get(:@lookup)
+    assert Primer::Classify::Cache.instance.empty?
+    assert Primer::Classify::AttrCache.instance.empty?
   ensure
-    Primer::Classify::Cache.instance.preload!
+    Primer::Classify::AttrCache.instance.preload!
+  end
+
+  def test_evicts_entries
```
BlakeWilliams, line 16:
```suggestion
  def test_evicts_lru_entries
```

I saw what I believed to be some LRU behavior in the code above, what are your thoughts on naming the test to reflect that we expect the cache to behave that way?

test/lib/classify/cache_test.rb
```diff
@@ -5,9 +5,60 @@
...
+    end
+
+    Primer::Classify::Cache.instance.clear!
+    Primer::Classify.call(classes: "m-1")
+    Primer::Classify.call(classes: "m-1")
+
+    assert_equal hits, 1
+    assert_equal misses, 1
+  ensure
+    ActiveSupport::Notifications.unsubscribe(hit_sub) if hit_sub
```
BlakeWilliams, line 61: This is totally fine, but we could also use [temporary subscriptions](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html#module-ActiveSupport::Notifications-label-Temporary+Subscriptions) if we wanted.

BlakeWilliams (approved):
I had a few small questions, but overall this is looking great!

lib/primer/classify/attr_cache.rb
camertron: Hmm... we could, although it's supposed to only be used in tests.

test/benchmarks/bench_classify.rb
camertron: Hmm that's a good call, I can put in another one.

test/lib/classify/cache_test.rb
camertron: I like it!

test/lib/classify/cache_test.rb
camertron: Whoa I had no idea that was a thing, very cool! Definitely a clearer way to go for these tests :)

test/lib/classify/cache_test.rb
camertron: Yeah, nice!

camertron:
Slack conversation between @jhawthorn and myself, reproduced here for posterity.

@jhawthorn:

LRU logic looks good, though I'm surprised this is faster than not caching TBH.

deleting and inserting into hashes isn't cheap, and actually hashing with non-basic objects (like hashes and arrays) can be quite slow even for lookups

The `ActiveSupport::Notifications.instrument` is also probably a lot slower if anything subscribes to it (and the subscriber would add a lot)

so it would be good to check if that's all in the end going to eat all the gains

so it would be good to check if that's all in the end going to eat all the gains

(hash modifications and access are not thread safe)

@camertron:

> deleting and inserting into hashes isn't cheap, and actually hashing with non-basic objects (like hashes and arrays) can be quite slow even for lookups

Hmm that's interesting, is that true even if the hash's internal capacity isn't changing? It seems to be pretty performant, although that may just be because `classify` is so slow 😆 

Although `#fetch` accepts `*args` I think we can get away with only hashing a single string since that's the only argument passed in. I'll update the PR.

> also I imagine thread safety is an issue if the instance is global

Based on my (rudimentary) understanding of Unicorn and singletons, I think each process will have its own cache instance. Is that not right?

@jhawthorn

> Hmm that's interesting, is that true even if the hash's internal capacity isn't changing? It seems to be pretty performant, although that may just be because classify is so slow

Yes, the hash value needs to be calculated either way. Of course performance is relative but I would resort to caching as a last resort vs making uncached version fast (since the hash lookup can only ever be so fast).

@camertron 

> Yes, the hash value needs to be calculated either way.

Right, ok so the expensive part is calculating the hash key, not the actual insert and delete operations themselves.

> I would resort to caching as a last resort vs making uncached version fast

I'm no Ruby performance expert, but I haven't found any other ways to squeeze performance out of classify besides what I've already done in this and the previous performance PR. That's why I'm looking to add this additional layer of caching.

camertron:
Based on @jhawthorn's comments, I'm going to close this PR in favor of this one: https://github.com/primer/view_components/pull/898

Issue #824, opened 2021-09-24
Regression: border added underneath tab in `Alpha::TabNavComponent`


paulcsmith:
It looks like the transition from `TabNavComponent` to `Alpha::TabNav` introduced a small regression.

## Problem

Border appears under selected tab. See https://primer.style/view-components/components/alpha/tabnav

<img width="218" alt="Screen Shot 2021-09-24 at 2 26 24 PM" src="https://user-images.githubusercontent.com/22394/134723003-61975bd7-876f-460c-aaa2-78162460eaf2.png">

## Expected

No border under selected tab. See https://primer.style/css/components/navigation#tabnav

<img width="181" alt="Screen Shot 2021-09-24 at 2 25 30 PM" src="https://user-images.githubusercontent.com/22394/134722889-f980b656-ac3b-4665-b6b6-44d8ada6fd1a.png">

## Cause of the problem

An extra `tabnav` class was accidentally added to the generated `ul`

```
<!-- from Primer docs -->
<ul data-view-component="true" class="tabnav-tabs tabnav">
```

It should be just `tabnav-tabs`

```
<ul data-view-component="true" class="tabnav-tabs">
```

---

I tested this with the inspector in the TabNav component docs and it removes the border. I looked at the PR but I did not see where the extra class is coming from 🤔I'm not super familiar with how Primer works so I'm probably missing something obvious. Happy to take a look at fixing this if someone can point me in the right direction

PR #854, opened 2021-10-15
Fix tab nav active border

theinterned:
fixes https://github.com/primer/view_components/issues/824

It looks like the transition from `TabNavComponent` to `Alpha::TabNav` introduced a small regression: the active tab renders a bottom border when it _should_ have no bottom border. See this comment https://github.com/github/github/pull/192828#issuecomment-926818751

## Cause

As @paulsmith identified, the border is caused by an extra `.tabnav` class on the `ul`. 

Where did that extra class come from? It turns out that `Primer:: Alpha::TabNav#intialize` was assigning the passed in `system_arguments[:classes]`  to the `@body_arguments[:classes]` instance variable.

## Solution

I just changed the `#initialize` to correctly assign `body_arguments[:classes]` to `@body_arguments[:classes]`, rather than assigning `system_arguments[:classes]`  to the `@body_arguments[:classes]`.  https://github.com/primer/view_components/pull/854/commits/da397cbc542d22c48f24a8a8126b6fe8aa57a5ce

```diff
- @body_arguments[:classes] = tab_nav_body_classes(system_arguments[:classes])
+ @body_arguments[:classes] = tab_nav_body_classes(body_arguments[:classes])
```

#### Before

Screen Shot 2021-10-15 at 5 31 44 PM

There is a a border under the active tab that should not be there.

#### After

Screen Shot 2021-10-15 at 5 30 11 PM

No more border!

## Also

As an aside, I noticed the html of the `tav_nav.stories.rb` was malformed: the `.extra` html had an opening `button` tag but a closing `a` tag. This was causing the story to render all kinds of weird. 

### Solution 

I just changed the closing tag to a matching `button`!

 See https://github.com/primer/view_components/pull/854/commits/fd9c86471d39644adb36362807a96023a61abab9

#### Before

Screen Shot 2021-10-15 at 5 45 46 PM

#### After
Screen Shot 2021-10-15 at 5 45 15 PM

## Question re mutating passed arguments

I also noticed that we are mutating the passed `system_arguments` and `body_arguments` function rather than cloning them. This resulted in the subsequent assignments to `@system_arguments` having the side-effect of mutating the `system_arguments` argument which is later passed. Usually mutating function args is considered a code smell. Should we be `.clone`ing these hash arguments to avoid mutating them?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ The category for changes related to documentation, testing and tooling. Also, fo
 * Make `Spinner` more accessible by adding `sr-only` loading text.
 
     *Manuel Puyol*
+
 ### Bug Fixes
 
 * Removes unwanted bottom border from active tab of `Alpha::TabNav`.


theinterned:
Hello @manuelpuyol (or other reviewers!) I'm wondering if you have tips on how I could add a test for this? Do you think it's even necessary to add coverage for this? Do we snapshot our stories?

PR #856, opened 2021-10-18
Blankslate slot refactor

manuelpuyol:
Related to https://github.com/github/primer/issues/373 and https://github.com/github/primer/issues/372.

With all the a11y and API recommendations, it looks like a good idea to prepare the braking changes with a new `Primer::Beta::Blankslate`. Also, since the component was super outdated, upgrading it to use slots will enable us faster and better API updates!

This PR also introduces some a11y changes:

1. Increase the button size for better touch area
2. Increase Link padding fo better touch area
3. a11y guidelines in the doc
4. Make `title#tag` required, so users must check their context to define an appropriate heading level


app/components/primer/beta/blankslate.rb
```diff
@@ -17,7 +17,11 @@ class Blankslate < Primer::Component
 
       # Optional graphic visual.
       #
-      # @param type [Symbol]  <%= one_of(Primer::Beta::Blankslate::GRAPHIC_OPTIONS) %>
+      # Use:
+      # - `graphic_icon` for an <%= link_to_component(Primer::OcticonComponent) %>.
```
joelhawksley, line 22: I think this is fine for now 👍🏻 We might want to look into something more automated in the future 😄

app/components/primer/beta/blankslate.rb
```diff
@@ -0,0 +1,241 @@
...
+        title.present?
+      end
+
+      def wrapper
+        unless @border
+          yield
+          return # returning `yield` caused a double render
+        end
+
+        content_tag(:div, class: "Box") do
```
joelhawksley, line 235: Wouldn't we want to use BorderBox here?

app/components/primer/beta/blankslate.rb
```diff
@@ -0,0 +1,241 @@
...
+          @system_arguments[:classes],
+          "blankslate",
+          "blankslate-narrow": narrow,
+          "blankslate-large": large,
+          "blankslate-spacious": spacious
+        )
+      end
+
+      def render?
+        title.present?
```
joelhawksley, line 226: This feels like a feature for the framework: required slots!

app/components/primer/beta/blankslate.rb
manuelpuyol: `BorderBox` doesn't allow us to put stuff directly into `<div class="Box">` :/

app/components/primer/beta/blankslate.rb
```diff
@@ -0,0 +1,241 @@
...
+      #     <% c.title(tag: :h2).with_content("Title") %>
+      #     <% c.description { "Description"} %>
+      #   <% end %>
+      #
+      # @example Icon
+      #   @description
+      #     Add an `icon` to give additional context. Refer to the [Octicons](https://primer.style/octicons/) documentation to choose an icon.
+      #   @code
+      #     <%= render Primer::Beta::Blankslate.new(icon: :globe) do |c| %>
+      #       <% c.graphic_icon(icon: :globe) %>
```
joelhawksley, line 113: Is this example correct? I'd expect us to pass `icon` to the slot, not the component itself.

joelhawksley (approved):

PR #865, opened 2021-10-21
Use .text-small instead of .btn-sm for small-variant link ButtonComponent

cheshire137:
This modifies the class that gets applied when you have a `variant: small, scheme: :link` ButtonComponent. Instead of applying `.btn-sm`, which doesn't affect the appearance of a link button, it applies `.text-small`. I hit this as a point of confusion; I felt like passing `variant: :small` would work regardless of my ButtonComponent's `scheme`.

I wasn't sure of a good class to apply for `variant: :large, scheme: :link`, so I omitted that from being a valid `variant` option for scheme=link. Suggestions welcome!


jonrohan:
Closing as stale

PR #870, opened 2021-10-26
Remove `hx_Subhead--responsive`

simurai:
This is part of https://github.com/github/primer/issues/322 and removes the `hx_Subhead--responsive` class.

`Subhead` is now [responsive by default](https://github.com/primer/css/pull/1690) and this class is not needed anymore.


simurai:
Ok, with https://github.com/github/github/pull/198706 merged on dotcom, the `hx_Subhead--responsive` class does not exist anymore. So I guess this PR could already be merged too since having `hx_Subhead--responsive` doesn't really do anything.

camertron (approved):

PR #875, opened 2021-10-28
Add cacheable module

BlakeWilliams:
This introduces a new module, `Primer::Cacheable` which is extracted
from the monolith. This module provides the ability for an individual
component to opt-in to caching behavior.

When enabled, this module:
* Caches component instances based on the args, kwargs, and block it
  receives.
* Caches the content generated by each component instance, so renders
  are free after the first call.
* Introduces a new method, `.with` which must be used instead of `.new`
  to opt-in to the caching behavior.

Each component that includes this module must implement a method called
`cache_container` that returns a mutable hash. It's recommended that
this method return an attribute defined on an instance of
`ActiveSupport::CurrentAttributes` which allows the cache to be cleared
after each request/response.

The largest con to this approach is that component calls must be updated
to use `.with` instead of `.new`. This is intentional since it could be
difficult to track down this functionality if we kept it behind the
standard `.new.


jonrohan:
Closing as stale

PR #877, opened 2021-11-03
Add host for Storybook JS

manuelpuyol:
In production, Storybook is trying to load assets from `primer.style/assets/asset-name`, which doesn't exist.
I'm proposing a new Vercel redirect specific for rails routes (such as assets and test controllers) https://github.com/primer/primer.style/pull/276

This should fix stories that depend on JS in our Storybook

cc @inkblotty


inkblotty (approved):
🙌

jonrohan (approved):

PR #880, opened 2021-11-08
More blankslate updates

manuelpuyol:
Following review by @ashygee, I'm updating Blankslate:

1. Changed the `title` slot to `heading`.
2. Rolled back the button to use variant `:medium`.
3. Added `px` information about `spacious` and `narrow` variants.
4. Removed `large` variant.
5. Renamed `graphic` slot to `visual`


alliethu:
@ashygee @manuelpuyol 

> Is the default value for heading not set to :h3? In the API audit I noted that h3 is the default. @alliethu what do you think from an a11y perspective?


Unless I misunderstood, [@khiga8's comment](https://github.com/github/primer/issues/372#issuecomment-942619071) suggested we don't pass in a default heading, because heading type will be contextual per view. Instead we should require them to pass it in themselves, with the guidance to follow the correct hierarchical sequence of the page.

ashygee:
> Instead we should require them to pass it in themselves, with the guidance to follow the correct hierarchical sequence of the page.

Thank you for clarifying @alliethu! So in this case then we are, in a way, forcing the developer to make a decision upon implementation?

manuelpuyol:
> > Instead we should require them to pass it in themselves, with the guidance to follow the correct hierarchical sequence of the page.
> 
> Thank you for clarifying @alliethu! So in this case then we are, in a way, forcing the developer to make a decision upon implementation?

Yes!

manuelpuyol:
> Should we clarify if the Primary_action is either a button or a link?

The primary_action is **always** a link that **looks like a button** 🤦 
@ashygee is there a good way of making that clear?

ashygee:
> @ashygee is there a good way of making that clear?

@manuelpuyol in the description for the slot we can write that information. Once we formalize a plan for a `ButtonLink` type of component then we can update the `Blankslate` to follow the new pattern but for now we could just state something like:

> The `Primary_action` is applied as an anchor link which is visually styled as a button to provide more emphasis to the primary action. An optional `Secondary_action` link can also be added to the Blankslate to link the user to additional information (e.g. Help documentation). This link will always be styled as a normal text link.

manuelpuyol:
@ashygee this is ready for a final review

ashygee (approved):
🚀  Looks good! Ship it!

joelhawksley (approved):

PR #887, opened 2021-11-10
Add support for responsive borders

jklina:
Primer CSS has support for [responsive borders](https://primer.style/css/utilities/borders#responsive-borders), however this was not ported to PVC. This is an attempt to bring support for those over to PVC by adding responsive options to the border system arguments. The API resembles other responsive attributes:

```ruby
assert_generated_class("border-bottom-0 border-sm-bottom border-lg-bottom-0", { border_bottom: [0, true, nil, 0, nil] })
assert_generated_class("border-0 border-sm-left border-lg-right border-xl", { border: [0, :left, nil, :right, true] })
```

With the new responsive layout component in the pipeline, responsive borders will come in handy when paired with the shuffling of the layout during breakpoints.


camertron (changes requested):
Could you also preload the classify cache with these guys? https://github.com/primer/view_components/blob/main/lib/primer/classify/cache.rb#L37

jklina:
@camertron done! https://github.com/primer/view_components/pull/887/commits/a7e1055e9ad295d5ab7922b5d7a67eda58b250b7

camertron (approved):

PR #892, opened 2021-11-12
If there's no content, use a self-closing tag

owenniblock:
Based off [this conversation](https://github.slack.com/archives/CGJTTJ738/p1636719084231700) and the comment below I've added a condition so that only valid tags self-close. I've also made it so that these tags always self-close - I figure if these are the list of valid void elements then we don't need to check that `content` exists.

TODO:
- [x] Add some tests


koddsson:
The HTML spec only allows ["void elements"](https://html.spec.whatwg.org/multipage/syntax.html#void-elements) to be self-closing. These are: 

> area, base, br, col, embed, hr, img, input, link, meta, param, source, track, wbr

Self-closing any other HTML elements is considered invalid HTML.

>This error occurs if the parser encounters a start tag for an element that is not in the list of void elements or is not a part of foreign content (i.e., not an SVG or MathML element) that has a U+002F (/) code point right before the closing U+003E (>) code point. The parser behaves as if the U+002F (/) is not present.

https://html.spec.whatwg.org/multipage/parsing.html#parse-error-non-void-html-element-start-tag-with-trailing-solidus

camertron (approved):
LGTM :)

tobiasahlin:
This is great 🙏 it's maybe worth noting that self-closing with a `/` is an old remnant from XHTML. It's compatible with HTML5, but the character is ignored either way, so we may as well generate `<br>`, ``, and not `<br />` or ``: https://dev.w3.org/html5/html-author/#void

owenniblock:
@tobiasahlin that's a good point. I'd be happy to change this to exclude the `/` if enough users wanted it. Or maybe add a switch so users who have a preference can choose.

I guess I went for self-closing with the `/` because I'm old 😂

PR #897, opened 2021-11-15
Update button spacing

manuelpuyol:
Related to https://github.com/github/primer/issues/253#issuecomment-949215306

I'm changing the `Button` ERB structure so we can avoid unwanted spaces and newlines, which were interfering with spacing. See https://github.com/github/view_component/pull/1121

As stated in the issue, I'm increasing the spacing from icon and counter to 8px and I'm also updating `Dropdown` to use the `Button` caret instead of creating a custom one

Before/After:

image
image

The change is really subtle, but it does have a slight difference. Before we used margins of 4px but also had a whitespace at the beginning of the text, which made the actual distance of around 6px


app/components/primer/button_component.rb
```diff
@@ -111,5 +119,10 @@ def initialize(
     def link?
       @scheme == LINK_SCHEME
     end
+
+    def trimmed_content
+      # strip unsets `html_safe`, so we have to set it back again to guarantee that HTML blocks won't break
+      content.strip.html_safe # rubocop:disable Rails/OutputSafety
```
manuelpuyol, line 125: opinions here? we treat `content` as safe, so I don't think this should be a problem

app/components/primer/button_component.rb
```diff
@@ -28,12 +28,20 @@ class ButtonComponent < Primer::Component
     # Icon to be rendered in the button.
     #
     # @param system_arguments [Hash] Same arguments as <%= link_to_component(Primer::OcticonComponent) %>.
-    renders_one :icon, Primer::OcticonComponent
+    renders_one :icon, lambda { |**system_arguments|
```
langermank, line 31: I have some concerns with using utility classes to address the button spacing issues, though I'm happy to see it being addressed! I would prefer that we make layout changes in Primer CSS so there's consistency between PVC and Primer CSS. 

The React team is also drafting a new Button API that I've been reviewing/pairing with @pksjce on. I've started drafting an API change for Primer CSS, which would address this spacing issue but also introduce a breaking change. I don't think we should wait for those changes necessarily, so maybe its worth making this change in CSS and cutting a minor release in the meantime. What do you think @manuelpuyol ?

camertron: Yeah, I'm a little concerned about this as well. Shouldn't spacing be handled by the caller?

app/components/primer/button_component.rb
joelhawksley: Can we add a check to make sure `content` is safe?

app/components/primer/button_component.rb
joelhawksley: I _think_ it's OK to have a wrapping component handle spacing amongst its children. This prevents us from having to write something like `.btn .octicon { margin-right...`.

manuelpuyol: I also think it's ok for the parent component to set the spacing of its children. The user won't be able to customize those spaces and we won't need CSS changes to achieve what we want

CHANGELOG.md
```diff
@@ -38,13 +38,17 @@ The category for changes related to documentation, testing and tooling. Also, fo
 
 ### Updates
 
-* Bump Storybook version to include Skip to Content links for keyboard auditors
+* Bump Storybook version to include Skip to Content links for keyboard auditors.
 
     *Katie Foster @inkblotty*
 
+* Update `Button` spacing.
```
joelhawksley, line 45: Can we be more specific about the spacing we're now overriding? Should we provide any feedback when someone tries to set the spacing themselves?

app/components/primer/button_component.rb
camertron: Ah sorry I didn't realize the spacing was on a child component, carry on!

langermank: @joelhawksley I think we _should_ handle it with CSS, and `btn` should have specific styles for handling `icon` or any kind of visual. Not by targeting `.octicon` but by providing a specific class like `.btn-icon` or even more generic `.leadingVisual` as we've introduced in ActionList. I'm not a fan of using utility classes inside components, even if its not visible to the end user of PVC.

`button` should be addressed on a larger scale and the CSS should address all child element scenarios. That said, when we do work on these changes we can simply remove this utility assuming its not required anymore.

vdepizzol: I agree with @langermank. Placing the inner component bits together in CSS like we did with `actionList` makes it easier for us to build, test, and maintain these components long-term. 

Not all responsive adaptations we're using now can be done by utilities anyway, and it feels it'd be harder to keep only _some_ properties in a PVC file.

🙇

manuelpuyol: We can aim to have these changes in PrimerCSS, but I think we can keep the utilities here while they aren't upstreamed :)

camertron (commented):
Where is the actual 8px change in this diff?

manuelpuyol:
> Where is the actual 8px change in this diff?

I updated the description with a screenshot.
Now, the button looks like:

`icon` 8px `text` 8px `counter`

manuelpuyol:
could I get 👀  @joelhawksley @ashygee ?

app/components/primer/button_component.rb
```diff
@@ -111,5 +119,16 @@ def initialize(
     def link?
       @scheme == LINK_SCHEME
     end
+
+    def trimmed_content
```
joelhawksley, line 123: I forget, did you file a ViewComponent issue for this? It feels like a framework feature 😄

joelhawksley (approved):

PR #898, opened 2021-11-16
Rewrite Classify.call

camertron:
`Primer::Classify.call` is still one of the hottest codepaths (from this project) on github.com. This PR adds all remaining Primer classes to utilities.yml and replaces the existing cache with it. Now all lookups are made on the data contained within this file, which skips a bunch of computation.

### Results

```ruby
args = {
  bg: :success, color: :text_warning, vertical_align: :text_bottom, display: :block, mx: 4, mb: 4
}

Benchmark.ips do |x|
  x.report do
    Primer::Classify.call(**args)
  end
end
```

|Branch|IPS|
|-|-|
|Main|196.214k (± 7.6%) i/s -    975.732k in   5.002691s|
|This branch|382.536k (± 7.0%) i/s -      1.905M in   5.006189s|

196.214 / 382.536 ~ 2x speedup

Combined with #807 that's a ~4.4x speed increase from the original implementation.


jonrohan (approved):
Nice I like your custom_utilities angle, I was stuck on migrating the rest of the utilities out of classify because the naming in primer didn't match the naming in view_components which made everything complicated.

joelhawksley (approved):

Issue #656, opened 2021-06-22
Raise warning or error when user sets `tag:` attribute to components that have fixed tag


khiga8:
Currently, we ignore the users input which isn't great. 

Maybe we can have something similar to https://github.com/primer/view_components/pull/634 but at the component level rather than base component level.

PR #900, opened 2021-11-16
Raise a warning on tag parameter when tag is fixed

owenniblock:
# Motivation

Attempt to fix this https://github.com/primer/view_components/issues/656 Issue. I've refactored the `denylist` code out to the `Component` as this is shared between the BaseComponent and higher level components.

I've also added a couple of helper methods:

1. So in simple cases we don't need to add the denylist to the arguments array.
2. For the case where we just want to check that the user hasn't specified the `:tag`

I need to add a couple of tests for some codepaths that we're not hitting and get some feedback on this approach before applying it to other components.


app/components/primer/base_component.rb
```diff
@@ -151,36 +151,10 @@ class BaseComponent < Primer::Component
...
 
-            raise(ArgumentError, message)
-          end
-        end
-
-        # Remove :system_arguments_denylist key and any denied keys from system arguments
-        @system_arguments.except!(:system_arguments_denylist)
-        @system_arguments.except!(*denylist.keys.flatten)
-      end
+      raise ArgumentError, "`class` is an invalid argument. Use `classes` instead." if system_arguments.key?(:class) && !Rails.env.production?
```
manuelpuyol, line 157: why not also move this to `validate_arguments`?

app/components/primer/component.rb
```diff
@@ -42,5 +42,47 @@ def validate_aria_label
     def silence_deprecations?
       Rails.application.config.primer_view_components.silence_deprecations
     end
+
+    def check_no_user_tag(**arguments)
+      check_denylist({ [:tag] => "This component has a fixed tag." }, **arguments)
```
manuelpuyol, line 47: since this is super straightforward, I wonder if we should do something like we do for `class`

```rb
raise ArgumentError, "This component has a fixed tag." if arguments.key?(:tag) && raise_on_invalid_options? && !ENV["PRIMER_WARNINGS_DISABLED"]
```

manuelpuyol (commented):
a few comments, but I'm interested to see this approach in prod :)

app/components/primer/component.rb
owenniblock: 👍 Yeah, that's much simpler!

app/components/primer/base_component.rb
owenniblock: Good call, I've also changed it to use the same switch as the other alerts here.

owenniblock:
Once we're happy with the approach, shall I apply this to all components that have a fixed tag in this PR?

app/components/primer/image.rb
```diff
@@ -27,7 +27,7 @@ class Image < Primer::Component
     # @param lazy [Boolean] Whether or not to lazily load the image.
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     def initialize(src:, alt:, lazy: false, **system_arguments)
-      @system_arguments = system_arguments
+      @system_arguments = deny_single_argument(:tag, "This component has a fixed tag.", **system_arguments)
```
manuelpuyol, line 30: I think we could still have a `deny_tag_argument` helper:

```rb
def deny_tag_argument(**arguments)
  deny_single_argument(:tag, "This component has a fixed tag.", **system_arguments)
end
```

since this will be used in many components

manuelpuyol (commented):
just one more suggestion, but we can expand it to all components!

owenniblock:
OK - marking this as Ready for Review.

Some notes:

I have removed the check from `app/components/primer/octicon_component.rb` because it made it fail the benchmark test - I figure we want this component to be as fast and efficient as possible. If we want to do that to any other components, let me know.

It also got me thinking about if there are any ways we could squeeze some more performance out of the code that's there. I guess we could remove `arguments.except!(key)` here:

```
    def deny_single_argument(key, help_text, **arguments)
      raise ArgumentError, "`#{key}` is an invalid argument. #{help_text}" \
        if arguments.key?(key) && raise_on_invalid_options? && !ENV["PRIMER_WARNINGS_DISABLED"]

      arguments.except!(key)
    end
```

We don't really need to remove the argument because we always override it later.

That would change:

```
@system_arguments = deny_tag_argument(**system_arguments)
```

To:

```
deny_tag_argument(**system_arguments)
```

manuelpuyol (approved):

PR #906, opened 2021-11-17
Use handles instead of names in CHANGELOG

manuelpuyol:
I'm proposing using handles since this is how you can reach most of the people that contribute in this project. It also has a nice integration with releases, showing all contributors in a release


joelhawksley (commented):
🤔 I see the argument for using handles, but I do like using full names as it puts a more human face on our release documentation. It's also the style Rails uses: https://github.com/rails/rails/blob/main/actionview/CHANGELOG.md

PR #916, opened 2021-11-25
Lock Node and Yarn version

pouretrebelle:
I couldn't get Storybook to work out of the box because I was using
the latest version of node instead of the long-term support version.
The LTS version (currently 16.3.0) works fine, and this locks each
package.json to it and the latest version of Yarn.


pouretrebelle:
Giving up on this one - [Vercel currently only supports Node v14](https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version) but we wouldn't want to lock the bundles served with view_components to v14, so it's sensible that it's left murky. I'll keep using v16 locally and keep my fingers crossed.

PR #917, opened 2021-11-25
Fix link to JS installation instructions

pouretrebelle:
I've been trying to figure out why the 'Please refer to the Installation section to set it up' link on primer.style doesn't work, and through that discovered how the documentation site fits together all the separate mini doc sites!

## Tests

- [x] Locally - Link to JS instructions on Autocomplete page goes to `/#installation`
- [x] Vercel preview - Link to JS instructions on Autocomplete page goes to `/view-components/#installation`

PR #918, opened 2021-11-29
Component status page

pouretrebelle:
## Summary

Counterpart to https://github.com/primer/react/pull/1656, part of https://github.com/github/primer/issues/114

- Component descriptions are generated from first line of docs pages via gatsby's excerpt for the mdx node

## [Preview](https://view-components-4d1640e2c-primer.vercel.app/view-components/status)

Screenshot 2021-11-29 at 11 08 58


docs/src/component-statuses.js
```diff
@@ -0,0 +1,67 @@
...
+      }
+    }
+  `)
+
+  const components = data.allMdx.nodes.map((node) => {
+    // get first sentence of the excerpt
+    // search for a period space and capital letter to avoid splitting at e.g.
+    const match = node.excerpt
+      .replaceAll('\n', ' ')
+      .match(/(^.+?\.)(\s+[A-Z])/, '.')
```
pouretrebelle, line 31: This is a bit sketchy. Would we want to have descriptions defined separately? I thought about updating the `docs.rake` script to add a frontmatter description to each of the md files but that would have taken me a lot longer than doing it this way 😅 and have similar sanitisation issues as this... thoughts welcome

docs/src/component-statuses.js
```diff
@@ -0,0 +1,67 @@
...
+      }
+    }
+  `)
+
+  const components = data.allMdx.nodes.map((node) => {
+    // get first sentence of the excerpt
+    // search for a period space and capital letter to avoid splitting at e.g.
+    const match = node.excerpt
+      .replace(/\n/g, ' ')
+      .match(/(^.+?\.)(\s+[A-Z])/, '.')
```
pouretrebelle, line 31: This is a bit sketchy. Would we want to have descriptions defined separately? I thought about updating the `docs.rake` script to add a frontmatter description to each of the md files but that would have taken me a lot longer than doing it this way 😅 and have similar sanitisation issues as this... thoughts welcome

pouretrebelle:
Currently any mdx node with a status value in the frontmatter is pulled into the table, but this doesn't line up with the sidebar navigation which is pulled in [from a yml file](https://github.com/primer/view_components/blob/v0.0.61/docs/src/@primer/gatsby-theme-doctocat/nav.yml). This confused me initially because this file wasn't mentioned in the contrib doc, and it seems quite easy to get out of sync; I wonder whether we could have that auto-generated as part of the docs build?

It seems like there a quite a few internal components (e.g. BorderBoxHeader used in BorderBox, BaseButton used in button components) that aren't in the sidebar and aren't linked to from other docs page - do they need a status or is it logical that whatever the 'parent' component is would reflect the status of the child? Is there a better way to call out these internal components? (preferably in a way that could exclude them from this table 🙃)

pouretrebelle:
N.B. failing lint is from main, I've fixed it in another PR https://github.com/primer/view_components/pull/917/commits/2fa2529ce402cbcee7418ca3b6d3761700bb72d5

docs/src/component-statuses.js
pouretrebelle: Should this PR also include copy changes so the descriptions have the same format? (Use X for Y)

manuelpuyol:
I think that using a query like that is more complicated than it needs to be 🤔 
We currently export static files with all the info we need https://github.com/primer/view_components/blob/main/static/statuses.json

Is it possible to read the data from that file in your React component? If not maybe we could add a new step to `docs.rake`, which can generate the necessary MDX based on the content of the file

joelhawksley:
> Is it possible to read the data from that file in your React component? If not maybe we could add a new step to docs.rake, which can generate the necessary MDX based on the content of the file

I would vote for that approach ☝🏻

docs/src/component-statuses.js
colebemis: Here's a list of how we could implement component descriptions (roughly increasing in "correctness"):

1. Parse the first sentence of page content (this is what your PR currently does)
2. Update the docs setup to use the `description` frontmatter variable. This will render the description slightly larger on the page and set the `meta` description tag (so better for SEO).
3. Move component descriptions into the [component-metadata](https://github.com/primer/component-metadata) repo and reference that metadata with the [componentId](https://primer.style/doctocat/usage/front-matter#component-metadata) frontmatter variable

I'll leave it up to you to decide what feels right to implement in this PR. We can take an incrementally correct approach here ✨

colebemis:
> I think that using a query like that is more complicated than it needs to be 🤔
We currently export static files with all the info we need https://github.com/primer/view_components/blob/main/static/statuses.json
> 
> Is it possible to read the data from that file in your React component? If not maybe we could add a new step to docs.rake, which can generate the necessary MDX based on the content of the file

I think trying to use `statuses.json` here would be more complicated to implement (open to being proven wrong though). Using Gatsby's GraphQL API makes it easier to get page links and component descriptions in addition to component statuses.

The `statuses.json` file and the `status` frontmatter variable get generated from the same data, right? So there shouldn't be any risk of the two being out of sync.

pouretrebelle:
I think the graphql method is the best:
- Getting simplified names and slugs from the data in `statuses.json` would be messy and we already have that info in graphql.
- Adding a step to `docs.rake` to generate the table would also be scrappy because we'd have to use HTML syntax rather than markdown (bc we'd need to use Gatsby's Link component so that it respects publicPath), and it also adds another auto-generated committed file that could be accidentally edited.

docs/src/component-statuses.js
pouretrebelle: Now you've explained to me the intentions of the component-metadata repo that seems like easily the best choice here, but I think this is a good interim solution until all the component data is in that repo because otherwise the descriptions are quite sparse.

Since here and #1656 are doing almost the same thing maybe we could build this page into doctocat based on pages that have `componentId` defined?

colebemis: Sounds good. Let's ship this as-is then collaborate on improving the component-metadata package.

colebemis (approved):

manuelpuyol (approved):

PR #920, opened 2021-11-29
Add Spinner system arguments to outermost element

### LINKED ISSUE #919: Regression: Spinner no longer toggleable with CSS
https://github.com/primer/view_components/pull/845 wrapped the spinner svg in a `span` to add a11y features, but this has caused an issue in .com where a sibling selector was used to toggle the spinner on/off [[see this Slack thread](https://github.slack.com/archives/CMZ4DC9BL/p1638179917092500)].

Passing a class into Spinner renders the class on the child svg element instead of the wrapping span element, so although toggling it with `display: none/block` would show/hide the spinner, the a11y advantages are lost because it's not hiding/showing the loading message itself - in fact it's worse than when there was no loading label because it'll announce loading as soon as the element is rendered.

Ideally the whole element would be added and removed from the DOM when needed, but since a spinner is almost always a side effect of something else it's probably much more common to toggle them using CSS, as in the issue linked above.

pouretrebelle:
Move the system arguments to Spinner's outmost element to fix a11y and nesting issues. This should result in no visual changes.

Closes #919

Note for reviewers - I'm a Rails n00b! Please teach me your conventions and best practices 🙏


pouretrebelle:
FYI CI lint error fixed in https://github.com/primer/view_components/pull/917/commits/2fa2529ce402cbcee7418ca3b6d3761700bb72d5

app/components/primer/spinner_component.html.erb
```diff
@@ -1,7 +1,15 @@
...
+<%= render Primer::BaseComponent.new(**@system_arguments) do %>
   <span class="sr-only">Loading</span>
-  <%= render Primer::BaseComponent.new(**@system_arguments) do %>
+  <%= render Primer::BaseComponent.new(
+    tag: :svg,
+    viewBox: '0 0 16 16',
+    fill: 'none',
+    animation: :rotate,
+    width: @size,
+    height: @size,
```
pouretrebelle, line 9: I was worried that there'd be instances where `style` was passed to override the spinner width/height but scanning through .com `style` is only used in one place and it's to set a `top` value

app/components/primer/spinner_component.html.erb
```diff
@@ -1,7 +1,15 @@
...
+    vertical_align: :bottom
```
pouretrebelle, line 10: Added this because imgs and svgs act weirdly inline without it, and there's no case where we wouldn't want a spinner to only take up its box. Could be too big of a change for this PR?

Screenshot 2021-11-29 at 17 14 17
Screenshot 2021-11-29 at 17 13 49

app/components/primer/spinner_component.html.erb
```diff
@@ -1,7 +1,15 @@
-<span role="status">
+<%= render Primer::BaseComponent.new(**@system_arguments) do %>
   <span class="sr-only">Loading</span>
-  <%= render Primer::BaseComponent.new(**@system_arguments) do %>
+  <%= render Primer::BaseComponent.new(
+    tag: :svg,
+    viewBox: '0 0 16 16',
+    fill: 'none',
```
manuelpuyol, line 6:
```suggestion
    fill: :none,
```

manuelpuyol (commented):
Adding system arguments to the wrapper has some strange effects, like not adding margins properly:

| Before | After |
|--------|--------|
| image | image |

pouretrebelle:
@manuelpuyol That layout bug should be fixed, I added `display: inline-block` to the span so it'll respect box styling properties like svg does. I'm not sure whether this counts as a bug fix or breaking change for the Changelog? Also am I meant to do anything about the failing benchmark test? I don't think it's to do with any of these changes.

app/components/primer/spinner_component.rb
```diff
@@ -31,13 +31,12 @@ class SpinnerComponent < Primer::Component
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     def initialize(size: DEFAULT_SIZE, style: DEFAULT_STYLE, **system_arguments)
       @system_arguments = deny_tag_argument(**system_arguments)
-      @system_arguments[:tag] = :svg
+      @system_arguments[:tag] = :span
+      @system_arguments[:role] = :status
+      @system_arguments[:display] = :inline_block
```
manuelpuyol, line 36: shouldn't a `span` always be `inline-block`?

pouretrebelle: No they're `inline` by default, that's why the margin wasn't being respected

manuelpuyol: hah TIL

manuelpuyol:
@pouretrebelle looks like `test/components/image_crop_test` is failing

PR #922, opened 2021-11-29
Add basic Content component to stop using slots hack

manuelpuyol:
In some cases, we want to change the wrapper/parent arguments according to what is given to a slot and only render the content it received. In those cases, we rely on a "hack", calling `view_context.capture { block&.call }`, which can confuse developers reading our code since we are dealing with ViewComponent internals.
We decide to use that hack to avoid creating an empty `Content` component, but I think that `Primer::Content` can make it easier to understand. Also, this hack is blocking changes in the framework itself https://github.com/github/view_component/pull/938


jonrohan (approved):
Seems ok to me 👍🏻

PR #923, opened 2021-11-29
Prevent aria-label from being used with invalid tags

manuelpuyol:
Follow-up on the Avatar a11y audit.

It is fairly common to add `aria-label` to elements that do not support it. I'm adding a general way of preventing it of being added to `div`, `span` and `p` without a proper `role`.

cc @khiga8 @owenniblock 

### References

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute#support
https://www.tpgi.com/short-note-on-aria-label-aria-labelledby-and-aria-describedby/


docs/content/components/tooltip.md
```diff
@@ -27,7 +27,7 @@ import Example from '../../src/@primer/gatsby-theme-doctocat/components/example'
 
 ### Default
 
-<Example src="<div class='pt-5'>  <span aria-label='Even bolder' data-view-component='true' class='tooltipped tooltipped-n'>Default Bold Text</span></div>" />
+<Example src="<div class='pt-5'>  <span data-view-component='true' class='tooltipped tooltipped-n'>Default Bold Text</span></div>" />
```
khiga8, line 30: Given `aria-label` is used for the current tooltip implementation, I think we need to keep it here and update examples to use `button` or some other supported element.

app/components/primer/component.rb
```diff
@@ -80,18 +82,35 @@ def validate_arguments(denylist_name: :system_arguments_denylist, **arguments)
...
 
     def deny_single_argument(key, help_text, **arguments)
       raise ArgumentError, "`#{key}` is an invalid argument. #{help_text}" \
-        if raise_on_invalid_options? && !ENV["PRIMER_WARNINGS_DISABLED"] && arguments.key?(key)
+        if should_raise_error? && arguments.key?(key)
 
       arguments.except!(key)
     end
 
+    def deny_aria_label(tag:, arguments:)
```
khiga8, line 97: We will also want to deny `aria-labelledby`

khiga8:
Love this! I am working on a linter which should complement this :) 

https://github.com/github/erblint-github/pull/5

app/components/primer/component.rb
manuelpuyol: I wanted to start with `aria-label` first to get a feel for it, we can expand it later :)

docs/content/components/tooltip.md
manuelpuyol: hmm I wonder if we should add an exception for Tooltip until we figure out the new version, wdyt?

khiga8: yeah, I think we should just deprecate `Tooltip`. 😅

manuelpuyol: I'm gonna skip the check for Tooltip, when you propose your new version here we can deprecate the old one :)

joelhawksley (approved):

app/components/primer/component.rb
```diff
@@ -44,7 +46,7 @@ def silence_deprecations?
     end
 
     def check_denylist(denylist = [], **arguments)
-      if raise_on_invalid_options? && !ENV["PRIMER_WARNINGS_DISABLED"]
+      if should_raise_error?
```
owenniblock, line 49: ❤️

owenniblock (approved):
Looks great!

PR #924, opened 2021-11-30
Button API updates

manuelpuyol:
Follow-up on the Button API audit recommendations.

### Changes:

1. Rename `caret` argument to `dropdown` in `ButtonComponent`.
1. Remove `:large` variant from `ButtonComponent`.
    - This variant is barely used and should be easy to remove.
    - For cases that currently use it, we can add `classes: "btn-large"` for now until @ashygee figures out what to do with them.
1. Deprecate `icon` and `counter` slots in `ButtonComponent` in favor of `leading_visual` and `trailing_visual`.
    - I'm using `alias` to keep the old slot names since migrating them is going to take some time.


app/components/primer/button_component.rb
```diff
@@ -20,28 +20,29 @@ class ButtonComponent < Primer::Component
...
       Primer::OcticonComponent.new(**system_arguments)
     }
+    alias icon leading_visual # remove alias when all buttons are migrated to new slot names
 
-    # Counter to be rendered in the button.
+    # Trailing visuals appear to the right of the button text and are set to <%= link_to_component(Primer::CounterComponent) %>.
     #
     # @param system_arguments [Hash] Same arguments as <%= link_to_component(Primer::CounterComponent) %>.
-    renders_one :counter, lambda { |**system_arguments|
+    renders_one :trailing_visual, lambda { |**system_arguments|
```
joelhawksley, line 40: What's the reasoning for this rename? Without any context, this API seems less clear. I could see this making sense as a polymorphic slot so we would have `trailing_visual_counter`, for example.

manuelpuyol: the idea is to make it more consistent with other components that also follow the leading/trailing pattern
I wanted to start without the polymorphic at first since we only accept counter, but I guess `trailing_visual_counter` could be ok to do so we can avoid yet another breaking change in the future

app/components/primer/button_component.rb
```diff
@@ -20,28 +20,33 @@ class ButtonComponent < Primer::Component
...
     VARIANT_MAPPINGS = {
       :small => "btn-sm",
-      DEFAULT_VARIANT => "",
-      :large => "btn-large"
+      DEFAULT_VARIANT => ""
     }.freeze
     VARIANT_OPTIONS = VARIANT_MAPPINGS.keys
 
-    # Icon to be rendered in the button.
+    # Leading visuals appear to the left of the button text and are set to <%= link_to_component(Primer::OcticonComponent) %>.
```
joelhawksley, line 27:
```suggestion
    # Leading visuals appear to the left of the button text and are passed to <%= link_to_component(Primer::OcticonComponent) %>.
```

app/components/primer/button_component.rb
```diff
@@ -20,28 +20,33 @@ class ButtonComponent < Primer::Component
...
+        system_arguments[:mr] = 2
 
-      Primer::OcticonComponent.new(**system_arguments)
+        Primer::OcticonComponent.new(**system_arguments)
+      }
     }
+    alias icon leading_visual_icon # remove alias when all buttons are migrated to new slot names
 
-    # Counter to be rendered in the button.
+    # Trailing visuals appear to the right of the button text and are set to <%= link_to_component(Primer::CounterComponent) %>.
```
joelhawksley, line 39:
```suggestion
    # Trailing visuals appear to the right of the button text and are passed to <%= link_to_component(Primer::CounterComponent) %>.
```

joelhawksley (approved):

joelhawksley (approved):

PR #932, opened 2021-12-03
Deprecate button `variant` in favor of `size`.

manuelpuyol:
`size` is **not** an HTML attribute for `<button>` so it should be fine to have a `size` argument in the component.
Here, I'm deprecating `variant` instead of removing it so we can have more time to migrate buttons in dotcom. I'm also adding a new linter to migrate deprecated arguments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ The category for changes related to documentation, testing and tooling. Also, fo
 
 ### Deprecations
 
-* Deprecate `variant` in favor of `size` in `ButtonComponent`.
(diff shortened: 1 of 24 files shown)


langermank (approved):
This is exciting! Will this allow us to re-introduce `variant` in the future to represent `scheme`? Or is the goal to remove `variant` permanently?

manuelpuyol:
> This is exciting! Will this allow us to re-introduce `variant` in the future to represent `scheme`? Or is the goal to remove `variant` permanently?

the goal is to remove `variant` since that is actually a concept from the `view_component` framework (and also Rails).

PR #940, opened 2021-12-09
Add `NavigationList` as alpha component

leequarella:
This promotes the new `NavigationList` component from `Experimental` to `Alpha`.


app/components/primer/alpha/navigation_list.html.erb
```diff
@@ -0,0 +1,18 @@
...
+      <% items.each do |item| %>
+        <%= item %>
+      <% end %>
+      <% unless sections.empty? %>
+        <li class="ActionList-sectionDivider"></li>
+      <% end %>
+      <% sections.each_with_index do |section, idx| %>
+        <% if idx > 0 %>
+          <li class="ActionList-sectionDivider"></li>
+        <% end %>
```
manuelpuyol, line 13: would this be the same as

```erb
<% sections.each_with_index do |section, idx| %>
  <li class="ActionList-sectionDivider"></li>
```

app/components/primer/alpha/navigation_list.rb
```diff
@@ -0,0 +1,131 @@
...
+      # @param list_arguments [Hash] <%= link_to_system_arguments_docs %>
+      def initialize(selected_item_id: nil, item_classes: "", **list_arguments)
+        @list_arguments = list_arguments
+        @list_arguments[:classes] = class_names(
+          "ActionList",
+          @list_arguments[:classes]
+        )
+
+        aria_label = aria(:label, list_arguments)
+        raise ArgumentError, "an aria-label is required" if aria_label.nil?
```
manuelpuyol, line 122:
```suggestion
        raise ArgumentError, "an aria-label is required" if aria_label.nil? && !Rails.env.production?
```

manuelpuyol: we could also update

https://github.com/primer/view_components/blob/1b6359e871749c30f62f3d55daed49ea75842fc8/app/components/primer/component.rb#L44-L47

Right now, it looks at `@system_arguments`, but we could pass `@list_arguments` as an argument

app/components/primer/alpha/navigation_list/heading.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= render(Primer::BaseComponent.new(**@system_arguments)) do %>
+  <%= content %>
+<% end %>
```
manuelpuyol, line 3: let's move this to a `call` method in the `.rb` file

```rb
def call
end
```

app/components/primer/alpha/navigation_list/item.rb
```diff
@@ -0,0 +1,121 @@
...
+        # @param selected_by_ids [String, Array] The unique identifier or identifiers used to determine if the item is selected.
+        # @param href [String] The URL to link to.
+        # @param selected_item_id [String] The id of the currently selected item in the whole list. Can refer to an item or subitem.
+        # @param content_classes [String] Additional classes to add to the item's content, either an <a> or <span> tag.
+        # @param submenu_classes [String] Additional classes to add to each of the item's subitems, i.e. each <li> tag.
+        # @param disabled [Boolean] Whether or not the item is disabled. Disabled items are rendered as <span> tags instead of links.
+        # @param expanded [Boolean] Whether or not the list of subitems is expanded or collapsed.
+        # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+        def initialize(selected_by_ids: [], href: nil, selected_item_id: nil, content_classes: "", submenu_classes: "", disabled: false, expanded: false, **system_arguments)
+          @selected_by_ids = Array(selected_by_ids)
```
manuelpuyol, line 49: why do we need `Array()`?

app/components/primer/alpha/navigation_list/item.html.erb
```diff
@@ -0,0 +1,37 @@
+<%= render(Primer::BaseComponent.new(tag: :li, **@system_arguments)) do %>
```
manuelpuyol, line 1: since we always use `tag: :li` we can make it part of the `system_arguments` with `@system_arguments[:tag] = : li` and use `deny_tag_argument`

app/components/primer/alpha/navigation_list/item.rb
```diff
@@ -0,0 +1,121 @@
...
+          @system_arguments = system_arguments
+          @system_arguments[:"data-item-id"] = @selected_by_ids.join(" ")
+          @system_arguments[:classes] = class_names(
+            "ActionList-item",
+            @system_arguments[:classes]
+          )
+
+          @content_arguments = {
+            tag: !@href || disabled ? :span : :a,
+            href: @href,
```
manuelpuyol, line 63: what happens when `@href` is `nil`? will the HTML render something like `<span href>` or does it remove it?

app/components/primer/alpha/navigation_list/item.rb
```diff
@@ -0,0 +1,121 @@
...
+        end
+
+        def before_render
+          raise "Cannot render a trailing visual for an item with subitems" if subitems.present? && trailing_visual.present?
+
+          @system_arguments[:"aria-expanded"] = expanded?.to_s if subitems.present?
+          @system_arguments[:"aria-haspopup"] = "true" if subitems.present?
+          @system_arguments[:classes] = class_names(
+            selected? ? "ActionList-item--navActive" : "",
+            subitems.present? ? "ActionList-item--hasSubItem" : "",
```
manuelpuyol, line 105:
```suggestion
            "ActionList-item--navActive"  => selected?,
            "ActionList-item--hasSubItem"  => subitems.present?,
```

app/components/primer/alpha/navigation_list/section.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if heading %>
+  <%= heading %>
+<% end %>
```
manuelpuyol, line 3:
```suggestion
<%= heading %>
```

if there is nothing on the `heading` slot this doesn't render anything

app/components/primer/alpha/navigation_list/section.rb
```diff
@@ -0,0 +1,59 @@
...
+          )
+
+          component_klass.new(selected_item_id: @selected_item_id, **system_arguments)
+        }
+
+        # @param selected_item_id [Symbol] The id of the selected item. Should correspond to one of the item ids in the list.
+        # @param item_classes [Array<String>] Additional classes to add to the list's items.
+        # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+        def initialize(selected_item_id: nil, item_classes: "", **system_arguments)
+          @id = "nav-list-section-#{SecureRandom.uuid}"
```
manuelpuyol, line 37: not a problem with the component, but this will generate a different id every time we generate docs. Is there any way we can make the output always the same for docs?
Maybe `@id = "nav-list-section-#{SecureRandom.uuid}" || system_arguments[:id]` and we force the id on the examples?

app/components/primer/alpha/navigation_list/section.rb
```diff
@@ -0,0 +1,59 @@
...
+            "ActionList",
+            "ActionList--subGroup",
+            @system_arguments[:classes]
+          )
+
+          @selected_item_id = selected_item_id
+          @item_classes = item_classes
+
+          aria_label = aria(:label, system_arguments)
+          raise ArgumentError, "an aria-label is required" if aria_label.nil?
```
manuelpuyol, line 50:
```suggestion
          validate_aria_label
```

app/components/primer/alpha/navigation_list/item.rb
leequarella: It's not strictly necessary, but it is a nice convenience to allow folks to pass single arguments: `selected_by_ids: :an_id` or `selected_by_ids: [:an_id, :a_second_id]`. I'm open to removing that if you feel strongly about it. Let me know 😄

app/components/primer/alpha/navigation_list/item.rb
leequarella: When `@href` is `nil` it gets rendered as a `span` instead of `a`

app/components/primer/alpha/navigation_list.html.erb
leequarella: This is present to make sure the first section doesn't have a leading divider.

app/components/primer/alpha/navigation_list.rb
```diff
@@ -0,0 +1,131 @@
...
+    # `NavigationList` contains a list of navigation items (links) and corresponding leading
+    # and trailing visuals such as icons. Items can themselves have sub-items.
+    # Finally, `NavigationList` supports sections, which are logical groups of items with
+    # a section header.
+    class NavigationList < Primer::Component
+      # Top-level nav items shown above all sections.
+      #
+      # @param component_klass [Class] A custom component class to use instead of the default Item class.
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+      renders_many :items, lambda { |component_klass: Item, **system_arguments|
```
joelhawksley, line 16: Do we need this `component_klass:` option? I don't see it used anywhere and I think we had debated leaving it out at one point.

app/components/primer/alpha/navigation_list/item.rb
```diff
@@ -0,0 +1,127 @@
...
+      # Items are rendered as styled links. They can optionally include leading and/or trailing visuals,
+      # such as icons, avatars, and counters. Items are selected based on their item_id, which is a
+      # required parameter. Items can also contain lists of subitems, which are rendered hidden but can
+      # be expanded on click.
+      class Item < Primer::Component
+        include ViewComponent::PolymorphicSlots
+
+        # The leading visual rendered before the link.
+        #
+        # @param kwargs [Hash] The arguments accepted by <%= link_to_component(Primer::Beta::Avatar) %> or <%= link_to_component(Primer::OcticonComponent) %>
```
joelhawksley, line 15: What about the `svg` behavior? (Perhaps this means that we should find a way to document each polymorphic slot individually?

manuelpuyol:
Closing it because of the infinite Docs loop

PR #941, opened 2021-12-10
Update Primer::Image to use image_path

manuelpuyol:
This will make the API easier to use, and users can pass relatives paths safely


joelhawksley (approved):

app/components/primer/image.rb
```diff
@@ -29,8 +29,8 @@ class Image < Primer::Component
     def initialize(src:, alt:, lazy: false, **system_arguments)
```
joelhawksley, line 29: Might we add docs for this change to the `src` argument?

PR #947, opened 2021-12-15
Update blankslate accessibility docs

alliethu:
This updates the Accessibility usage guidelines for the blankslate component.


pouretrebelle:
@alliethu your changes were cruelly overwritten by the docs action because they're generated from the [YARD](https://yardoc.org/features.html) comments in the component ruby files 😞  I copied your changes to the right place and now they should be all good. (fwiw I think it's a very confusing system for the markdown files to be committed because it makes them look editable when they're fully generated, I will change that in the not distant future).

My only comment on the changes is the casing of the component name; it looks like in other component docs we use the the imported name and styling (`Blankslate`) but you've used blankslate here. Is that an intentional choice? [There was a discussion about it recently](https://github.com/github/primer/discussions/487) but no conclusion 🤔

manuelpuyol (approved):

alliethu:
👋🏼 @pouretrebelle thank you for doing that heavy lifting! I would have been completely lost trying to sort things out. 

> My only comment on the changes is the casing of the component name; it looks like in other component docs we use the the imported name and styling (Blankslate) but you've used blankslate here. Is that an intentional choice? There was a discussion about it recently but no conclusion 

It was an intentional choice, as that is how I've been seeing it done elsewhere. I did leave a comment in the discussion you shared to hopefully nudge a decision being made!

app/components/primer/beta/blankslate.rb
```diff
@@ -5,9 +5,11 @@ module Beta
     # Use `Blankslate` when there is a lack of content within a page or section. Use as placeholder to tell users why something isn't there.
     #
     # @accessibility
-    #   - Set the `heading` level based on what is appropriate for your page hierarchy. <%= link_to_heading_practices %>
-    #   - `secondary_action` can be set to provide more information that is relevant in the context of the `Blankslate`.
-    #   - `secondary_action` text should be meaningful out of context and clearly describe the destination. Avoid using vague text like, "Learn more" or "Click here".
+    #   - The blankslate uses a semantic heading that must be set at the appropriate level based on the hierarchy of the page.
+    #   - All blankslate visuals have been programmed as decorative images, using `aria-hidden=”true”` and `img alt=””`,  which will hide the visual from screen reader users.
+    #   - The blankslate supports a primary and secondary action. Both actions have been built as semantic links with primary and secondary styling.
+    #   - All links within the blankslate must receive visual and programmatic focus, and be accessible through mouse, keyboard and touch device.
```
khiga8, line 11: > All links within the blankslate must receive visual and programmatic focus 

Wondering if this is something the consumer needs to do. Does this happen by default if someone is setting a link using the slots: `c.secondary_action(href:`?

app/components/primer/beta/blankslate.rb
```diff
@@ -5,9 +5,11 @@ module Beta
     # Use `Blankslate` when there is a lack of content within a page or section. Use as placeholder to tell users why something isn't there.
     #
     # @accessibility
-    #   - Set the `heading` level based on what is appropriate for your page hierarchy. <%= link_to_heading_practices %>
-    #   - `secondary_action` can be set to provide more information that is relevant in the context of the `Blankslate`.
-    #   - `secondary_action` text should be meaningful out of context and clearly describe the destination. Avoid using vague text like, "Learn more" or "Click here".
```
khiga8, line 10: I think this might be helpful to keep. What do you think?

app/components/primer/beta/blankslate.rb
alliethu: @khiga8 per our convo, I will go ahead and remove this, as it's already taken care of in the component itself.

khiga8 (approved):
Looks good!

PR #949, opened 2021-12-16
DRAFT: Handover thoughts for new dialog component

owenniblock:
- https://github.com/github/accessibility/issues/451

Just been tinkering, this is not finished yet (we have to settle on the design, mark-up and CSS) but I wanted to add my thoughts for how we'd implement this before I get pulled on to something else next week.

TODO List:
- [x] Fix template and examples so it creates expected output
- [x] Discuss plans with Primer team
- [ ] See if we can use anything on the Overlay work?
- [ ] Agree design / HTML
- [x] Should we hide the `<footer>` when no buttons added?
- [ ] Add mark-up and CSS to primer/css
- [ ] Adjust code here to match decisions above
- [ ] Add stories for testing
- [ ] Add further examples to docs
- [ ] Get ready for Alpha release


owenniblock:
Some questions:

* Should we split the dialog in to sub components (thinking something like an Overlay Header, and an Overlay Footer)?
* Do any of the modals in discussions require the `...` button in the header or should we ignore that for now?

owenniblock:
Decided we should hide the `<footer>` when there are no contents.

owenniblock:
I've been looking through Primer today to get a clearer view in my mind as to what this will look like once we're done.

I think the next step could be to start building the JS in a separate project like here: https://github.com/github/details-menu-element - this seems to match the patterns used in other Primer components. I guess that any shared behaviors (e.g. Overlay) should be moved to primer/behaviors and the new JS component should inherit from that.

owenniblock:
@bolonio I added some work in progress code for the dialog. It's very incomplete and rough and ready but I wanted to test that the process would work. Turned out that you need to build the TyeScript using `npm run prepare` if you make changes to the TypeScript. I'll add that as a gotcha to the friction document.

PR #950, opened 2021-12-16
Refactors selectors to args logic

jklina:
Now that all selectors are stored in utilities.yml, we can build a hash with selectors as keys and the arguments and positions as values. This will allow us to remove some specialized mapping rules and clean things up overall.

This also has the nice advantage of being about 25x faster:

```ruby
Benchmark.ips do |x|
  x.report do
    Primer::Classify::Utilities.classes_to_args("mr-1 mr-sm-2 baz bin")
  end
end
```

```
Before:

                         1.266k i/100ms
                         12.570k (± 9.2%) i/s -     63.300k in   5.085757s

After:

                        30.697k i/100ms
                        305.832k (± 5.1%) i/s -      1.535M in   5.033715s
```

I did a memory estimate using `ObjectSpace.memsize_of` and its reporting this hash uses `57440 bytes`, though I don't believe this is used in production.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The category for changes related to documentation, testing and tooling. Also, fo
 ### Updates
 
 * Optimize logic for converting class names into args
+
     *Josh Klina*
 
 ### Deprecations
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ The category for changes related to documentation, testing and tooling. Also, fo
 
 ## main
 
+### Updates
+
+* Optimize logic for converting class names into args
+    *Josh Klina*
+
 ### Deprecations
 
 * Deprecate `variant` in favor of `size` in `ButtonComponent` and `ButtonGroup`.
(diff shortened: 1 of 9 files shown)


test/linters/argument_mappers/base_test.rb
```diff
@@ -30,12 +30,12 @@ def test_returns_known_system_arguments_as_string
   end
 
   def test_raises_if_a_class_cannot_be_converted
-    @file = '<div class="mr-1 p-3 d-none d-md-block anim-fade-in text-center"></div>'
+    @file = '<div class="mr-1 p-3 d-none d-md-block anim-fade-in text-fuzzy-waffle"></div>'
```
jklina, line 33: This is an actual class but for some reason was used as an example of a class that could not be converted. I'm not sure why `text-center` was not convertible, but with these changes it does become convertible. If that was a deliberate choice, I can revert.

CHANGELOG.md
```diff
@@ -30,6 +30,11 @@ The category for changes related to documentation, testing and tooling. Also, fo
 
 ## main
 
+### Updates
+
+* Optimize logic for converting class names into args
+    *Josh Klina*
```
joelhawksley, line 36:
```suggestion

    *Josh Klina*
```

joelhawksley (approved):

PR #951, opened 2021-12-16
Fix broken stories

manuelpuyol:
Updates stories with incorrect/missing attributes:

1. TabPanels
2. Truncate
3. BorderBox
4. Heading
5. Menu
6. Subhead

### DIFF
--- a/demo/Gemfile.lock
+++ b/demo/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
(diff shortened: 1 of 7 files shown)


joelhawksley (approved):
Is there a way we could have caught this issue in CI?

manuelpuyol:
> Is there a way we could have caught this issue in CI?

I'm not sure :(

PR #959, opened 2021-12-20
DRAFT: Rename the AutoComplete component to Autocomplete

hectahertz:


pouretrebelle:
A note for implementing the name change in consumers; `Primer::Beta::AutoComplete` is used in [8 files in .com](https://cs.github.com/?scopeName=github&scope=org%3Agithub&q=Primer%3A%3ABeta%3A%3AAutoComplete) and the old `Primer::AutoComplete` is used in [github/advisory-db](https://github.com/github/advisory-db/blob/8593f6909d63b8cc24b63dd208b3b0c632853300/app/components/cwes/autocomplete_component.html.erb#L5), although I don't think we have to worry about that.

PR #964, opened 2022-01-06
Fixes component-generator output when js packages added

owenniblock:
# Motivation

While working on a draft to add a dialog component, I discovered some flaws in the `component_generator.thor` script.

I validated these fixed in my branch which you can see on [my draft PR](https://github.com/primer/view_components/pull/949) but I wanted to pull those changes out since there is other separate component work planned by accessibility in the coming weeks.

# Notes

It may be useful for us to separate the logic triggered by `--js some-npm-package` - I can see further instances where we would want to take an existing component and add a JavaScript package to it (for example stripping out behaviours and moving them to https://github.com/primer/behaviors.

The biggest change is to the `test/application_system_test_case.rb` file - this test was not accessing the `status` directories that were being created. Another approach to this could be to avoid using `alpha` or `beta` entirely in this process but this felt like the approach that was intended by the code that was already there.

I have left in the changes:
https://github.com/primer/view_components/compare/fix-component-generator?expand=1#diff-c3ad7f6264098b2f742d9566b61dbe9da505daf112c9e318989dc723e4e02e02
and
https://github.com/primer/view_components/compare/fix-component-generator?expand=1#diff-0395af05f34bf8d74de71be36172e81ffb9bd74cbcaddf502dd49fe91035944d

Although they don't have bearing on this PR these changes occurred while running tests locally. Please let me know if you'd like me to revert those changes on this PR.

cc @pouretrebelle as I noticed that she already fixed one of these issues yesterday: https://github.com/primer/view_components/pull/963

### DIFF
--- a/demo/Gemfile.lock
+++ b/demo/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
-    primer_view_components (0.0.64)
+    primer_view_components (0.0.66)
       actionview (>= 5.0.0)
       activesupport (>= 5.0.0)
       octicons (~> 15)
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -21,9 +21,13 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   AXE_WITHIN_SELECTOR = "body"
 
   def with_preview(preview_name)
-    component_uri = self.class.name.gsub("Test", "").gsub("Integration", "").underscore
+    component_name = self.class.name.gsub("Test", "").gsub("Integration", "")
+    status = ""
+    status = "beta/" if component_name.include?("Beta")
+    status = "alpha/" if component_name.include?("Alpha")
+    component_uri = component_name.gsub("Alpha", "").gsub("Beta", "").underscore
 
-    visit("/rails/view_components/primer/#{component_uri}/#{preview_name}")
+    visit("/rails/view_components/primer/#{status}#{component_uri}/#{preview_name}")
 
     assert_accessible(page)
   end
--- a/component_generator.thor
+++ b/component_generator.thor
@@ -66,7 +66,7 @@ class ComponentGenerator < Thor::Group
   end
 
   def import_in_primer_ts
-    append_to_file("app/components/primer/primer.ts", "import './#{status_path}#{underscore_name}'") if js_package_name
+    append_to_file("app/components/primer/primer.ts", "import './#{status_path}#{underscore_name}'\n") if js_package_name
   end
 
   def install_js_package
--- a/static/constants.json
+++ b/static/constants.json
@@ -354,8 +354,7 @@
     }
   },
   "Primer::Dropdown": {
-    "Menu": "Primer::Dropdown::Menu",
-    "MenuTest": "Primer::Dropdown::MenuTest"
+    "Menu": "Primer::Dropdown::Menu"
   },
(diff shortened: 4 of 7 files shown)


test/application_system_test_case.rb
```diff
@@ -21,9 +21,13 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
...
   def with_preview(preview_name)
-    component_uri = self.class.name.gsub("Test", "").gsub("Integration", "").underscore
+    component_name = self.class.name.gsub("Test", "").gsub("Integration", "")
+    status = ""
+    status = "beta/" if component_name.include?("Beta")
+    status = "alpha/" if component_name.include?("Alpha")
+    component_uri = component_name.gsub("Alpha", "").gsub("Beta", "").underscore
 
-    visit("/rails/view_components/primer/#{component_uri}/#{preview_name}")
+    visit("/rails/view_components/primer/#{status}#{component_uri}/#{preview_name}")
```
pouretrebelle, line 30: This seems like a fragile way of deriving the preview path 😬 out of the scope of this PR to refactor but I'll make a note of it

demo/Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
-    primer_view_components (0.0.64)
+    primer_view_components (0.0.66)
```
pouretrebelle, line 4: Yeah this is annoying, I'll try and make this self-update in the release script

pouretrebelle (approved):

test/application_system_test_case.rb
owenniblock: Yeah, I thought the same - let's do it in this PR and then it's done! Initially I was just trying to get my stuff working.

Maybe I could change this to [starts_with](https://apidock.com/ruby/String/start_with%3F) and then at least we wouldn't be hit with a problem if someone creates a `BetaAlphabetSelector` although it would still fail if someone created `AlphabetSelector`.

If we want to go further, I could use some regex to match `Alpha` and `Beta` and a follow-up capital letter?

pouretrebelle: I was thinking more that getting the data from the class is a bit fragile, but is that pretty normal in Rails? Either way no harm in refactoring this with a bit of regex. Would something like this work:

```rb
match = /^(Alpha|Beta)([A-Z])/.match(component_name)
status = match[1]
statusPath = status ? "#{status.downcase}/" : ""
component_uri = component_name.gsub(status, "").underscore
```

owenniblock: > I was thinking more that getting the data from the class is a bit fragile, but is that pretty normal in Rails?

No, I wouldn't say normal. There's another option (some sort of look-up based on https://github.com/primer/view_components/blob/2ce820192a5fa027a9bc1bb806bf63475758369d/static/statuses.json#L1), but it would break when things do not match our expectations around naming conventions (which some components seem not to). I think the regex solution is better than my original solution and anything wider would need some conversations to work out the correct approach (and some remediation work to move around some of the existing preview files etc).

I'm happy to resolve this if you're happy with the code? I made a few changes from your suggestion above.

pouretrebelle (approved):

demo/Gemfile.lock
pouretrebelle: Fixed in https://github.com/primer/view_components/pull/966

manuelpuyol (approved):

PR #968, opened 2022-01-07
Fix Changelog required check

pouretrebelle:
@owenniblock alerted me that he was getting emails about invalid workflows. The error message can be found [in this run](https://github.com/primer/view_components/actions/runs/1667154685):

> The workflow is not valid. .github/workflows/changelog.yml (Line: 27, Col: 12): Job 'check_for_changes' depends on unknown job 'changedfiles'.

This is because of a typo introduced in https://github.com/primer/view_components/pull/766. The feature was first added in https://github.com/primer/view_components/pull/565.

The original intention of the lint was to require Changelog entries for every PR but there are lots of changes (e.g. this one) that don't effect the library itself where we'd want to skip this check. In the vein of https://github.com/primer/react and their `skip changeset` PR label, this uses a `skip changelog` PR label to bypass the check.

## Testing

- [x] No changelog changes and no labels = fail
- [x] No changelog changes and `skip changelog` label = skipped
- [x] Changelog change and no labels = pass
- [x] Changelog change and `skip changelog` label = skip

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,6 +1,15 @@
 name: CHANGELOG
 
-on: [pull_request]
+on:
+  pull_request:
+    branches:
+      - main
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+      - unlabeled
 
 jobs:
   changed_files:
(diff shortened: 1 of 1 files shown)


pouretrebelle:
Oh wow I didn't realise it would fail CI on changes that don't effect library output. I'll see if I can find an easy way of checking whether the changes do effect the lib output, or add an escape hatch like a `skip-changelog` PR label.

owenniblock (approved):
❤️ love this! Looking forward to seeing it merged 👍

manuelpuyol (approved):

PR #969, opened 2022-01-07
Remove blankslate heading styling

pouretrebelle:
`Primer::Beta::Blankslate` forces `.h2` styling on the heading slot even if a non-h2 tag is passed in.

[This was flagged by @rewinfrey on Slack](https://github.slack.com/archives/CGJTTJ738/p1641414573490700), a lot of confusion was caused by the `h4` tag having a `.h2` class, and since `.h2` uses `!important` declarations it couldn't be overridden.

[The migration script to beta defaults to a h2 tag](https://github.com/primer/view_components/blob/main/lib/primer/view_components/linters/blankslate_api_migration.rb#L51), so [there are only a handful of examples that specify a different tag](https://cs.github.com/?scopeName=github%2Fgithub&scope=repo%3Agithub%2Fgithub&q=%2Fc%5C.heading%5C%28tag%3A+%3Ah%5B1345%5D%2F); in these instances the font size would be reduced with this change.


pouretrebelle:
Question for @ashygee  - do we want to maintain the h2 sizing even if a different tag is passed in? I'm guessing that was the original intention but we'll have to add a `.blankslate-heading` class to Primer CSS to do that instead of using the utility (we'll have to do that as part of https://github.com/github/primer/issues/585 anyway to remove the margin utility).

jonrohan (approved):

ashygee:
> Question for @ashygee - do we want to maintain the h2 sizing even if a different tag is passed in? I'm guessing that was the original intention but we'll have to add a `.blankslate-heading` class to Primer CSS to do that instead of using the utility (we'll have to do that as part of [github/primer#585](https://github.com/github/primer/issues/585) anyway to remove the margin utility).

@pouretrebelle no I don't think we should. At the time of the [API audit ](https://github.com/github/primer/issues/373#issuecomment-943927091)I had made a recommendation that the default would be an `h3` but that the heading size could be changed depending on the hierarchical placement of the blankslate. The original intention of the blankslate, in my mind, was to be a placeholder in larger, more spacious UI. Thus a larger `h3` or `h2` heading size felt appropriate. But with the discovery of the usage in more condensed UI and the fact that we do not have an [alternative solution](https://github.com/github/primer/issues/624) at this time, I would say that we can and should be flexible with the heading sizing.

manuelpuyol (approved):

pouretrebelle:
Superseded by #1000

PR #988, opened 2022-01-24
Wrap icon, if provided, in a label for beta search autocomplete to provide accessible label

andrialexandrou:
Closes github/github#194695

Feedback most heartily invited, I'm new to working in both Ruby and Rails so anything you see is something I'd love to learn.

As mentioned in the issue, the icon being wrapped in a label covers the scenario of `the lack of a separate label reduces the hit area/target size of the field`, which an aria-label on the input wouldn't sufficiently cover.


app/components/primer/beta/auto_complete.rb
```diff
@@ -95,8 +104,7 @@ class AutoComplete < Primer::Component
       #   <% end %>
       #
       # @example With Icon
-      #   <%= render(Primer::Beta::AutoComplete.new(src: "/auto_complete", list_id: "fruits-popup-4", input_id: "fruits-input-4", position: :relative)) do |c| %>
-      #     <% c.label(classes:"").with_content("Fruits") %>
+      #   <%= render(Primer::Beta::AutoComplete.new("aria-label": "Fruits", src: "/auto_complete", list_id: "fruits-popup-4", input_id: "fruits-input-4", position: :relative)) do |c| %>
```
andrialexandrou, line 107: note API change
adding `aria-label="Your a11y description"`

app/components/primer/beta/auto_complete.rb
```diff
@@ -42,7 +42,16 @@ class AutoComplete < Primer::Component
 
       # Optional icon to be rendered before the input. Has the same arguments as <%= link_to_component(Primer::OcticonComponent) %>.
       #
-      renders_one :icon, Primer::OcticonComponent
+      renders_one :icon, lambda { |**system_arguments|
+        aria_label = aria("label", system_arguments) || @aria_label
+        if aria_label.present?
+          system_arguments[:"aria-label"] = aria_label
```
andrialexandrou, line 48: Note the addition here, *without* a subtraction of the `input` field's `aria-label`. It sounds okay to me with voiceover, not sure how important it is to have just one aria-label.

docs/content/contributing.md
```diff
@@ -42,6 +42,18 @@ system_arguments[:tag] = :h1
 system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, DEFAULT_TAG)
 ```
 
+## Running the demo app
```
inkblotty, line 45: 🙇  Thank you

app/components/primer/beta/auto_complete.rb
inkblotty: I'm interested to hear feedback from @jscholes on labelling this the same as the input.

inkblotty (commented):
💪

app/components/primer/beta/auto_complete.rb
andrialexandrou: correction -- api is the same as before, and as we discussed with James this is likely not its final state but fine for now.

app/components/primer/beta/auto_complete.rb
andrialexandrou: We spoke and he recommends changing the overall API to only allow for
- `label`
- `icon`
- `isLabelVisible`

in the interest of not allowing for a conflict between the `label` and if someone has provided an `aria-label`. But we'll be forking that work to be done elsewhere. I'll create an issue as follow-up.

docs/content/contributing.md
manuelpuyol: this should be covered by the `### Storybook / Documentation` section 🤔 
maybe we should move it up? `script/dev` should run everything you need

app/components/primer/beta/auto_complete/auto_complete.html.erb
```diff
@@ -1,6 +1,13 @@
 <%= label %>
 <%= render Primer::BaseComponent.new(**@system_arguments) do %>
+  <% if icon.present? %>
+    <label for="<%= @input_id %>">
```
manuelpuyol, line 4: what happens if we have both `label` and `icon`? will it have two `<label>`?

andrialexandrou: As it is currently written, yes it will have two labels. We have an incoming change to the `auto-complete-element` that will change this design.

manuelpuyol (approved):

docs/content/contributing.md
andrialexandrou: change uploaded 👍

Issue #982, opened 2022-01-19
Git-ignore generated content


pouretrebelle:
We currently keep a lot of ancillary files that are generated by scripts in git history. This means that for relatively small changes we can end up with large diffs, it causes confusion (e.g. [when](https://github.com/primer/view_components/pull/947#issuecomment-995857311) [people](https://github.com/primer/view_components/pull/426#issuecomment-812684660) update the component docs in the markdown files), and bugs that add unnecessary workarounds to the library ([e.g. infinite CI docs build loops because of randomised a11y ids](https://github.slack.com/archives/CGJTTJ738/p1642591536075900)). [We've had to add a script to our CI to generate the docs](https://github.com/primer/view_components/blob/v0.0.66/.github/workflows/docs-build.yml) to make sure they're up to date, increasing workflow times and adding extra, messy commits when they're out of sync.

I think these problems can be solved by us git-ignoring all the generated files, but continuing to build them in development and during deployment. I'm going to do a spike into how this could work and uncover any roadblocks, will report back in this thread.

## Generated content

`bundle exec utilities:build`:
- [ ] `lib/primer/classify/utilities.yml`

`bundle exec rake docs:build`:
- [x] Docs site markdown (components & adrs) https://github.com/primer/view_components/pull/996
- [ ] `static/arguments.yml`
- [ ] `static/classes.yml`

`bundle exec static:dump` (relies on docs build)
- [ ] `static/audited_at.json` (introduced in #860 but not used)
- [ ] `static/constants.json`
- [ ] `static/statuses.json`

N.B. `static/assets/` is not generated

pouretrebelle:
I started by tidying up the existing git-ignore rules in https://github.com/primer/view_components/pull/984. I then realised quite quickly (and I feel a little silly that I didn't realise this earlier) that the Vercel integration we use to deploy the docs site won't work if we ignore the files as we need Rails to run and build the docs before deployment. HOWEVER. [CSS uses a workflow for Vercel deployment](https://github.com/primer/css/blob/v19.2.0/.github/workflows/deploy_preview.yml) which would enable us to pre-build the docs in the workflow. This has two other advantages:
- The deployment alert is a timeline item rather than a comment on the PR, I find the comments can be annoying for notification management / to see whether anyone's reviewed your PR
- CSS also uses this for Storybook PR deployment which is something we don't have at all, and investigating this will enable that
I'm going to try this out and see how far I get 🔍

PR #989, opened 2022-01-25
Use GitHub actions for Vercel deployments

pouretrebelle:
## Description

This is a stepping stone to #982.

To be able to git-ignore generated content we need a way to build this content for deployment. Currently Vercel builds the Gatsby site but it expects all the YARD-generated markdown to already be in the file system.

By using a GitHub Action we can set up Ruby to run the static file generation and pass that content to Vercel for the build. In the spirit of doing all the building on our end, I also moved the Gatsby build from Vercel to the action, so Vercel just has to deploy the static files. I managed to get Vercel to stop running the build on their side (8b077a8ae12383a6d460af7fecddb50604be2c7b) but couldn't get it to stop installing the dependencies, so the total build time on Vercel is ~30secs. I think [we can follow these instructions](https://vercel.com/docs/concepts/deployments/build-step#build-command) to remove the installation completely, but this can only be done from the UI settings so should only be done once this is merged and all PRs are up to date (to avoid breaking existing PR preview apps). Since this is all static now we could use a different service for deployment (GH pages?) but that's a conversation for another day.

> Frustration sidenote - [Vercel's move to zero config](https://vercel.com/blog/zero-config) has discouraged people from using `vercel.json` files and doing all the config in the UI which makes it really difficult to make these sort of setup changes, and they've also removed all the docs related to `vercel.json` which is a big kick in the teeth when they're not even deprecated. Also there's no documentation on how the ['root directory'](https://vercel.com/docs/concepts/deployments/build-step#root-directory) changes things when you're using `.vercelignore` (without this we were uploading thousands of files for the deployment which would crash); by trial and error I found that `.vercelignore` _must_ be in the actual root but `vercel.json` _must_ be in the 'root directory' 🙃
>> Ideally we'd set the ['working directory'](https://github.com/amondnet/vercel-action#inputs) of the action to docs (then the `.vercelignore` can be in docs and not have to do the allowlist/blocklist thing it's currently doing) and the 'root directory' to `.` but we can't because ofc that's controlled in the Vercel settings UI and I can't figure out how to override it in `vercel.json` 🙃🙃 so we're stuck with this unless we make a separate Vercel project for the shift.

## Visible changes

| Before | After |
|--------|--------|
|  |  | 
| Vercel comments on your PR with the Preview link | GitHub Actions adds a timeline item with the Preview link |

The new way takes ~3m40 from action initiation to complete deployment, as I said above we should be able to shave off 30s from that after merge. The old way took ~2m20s to deploy on Vercel so we are losing a bit of time, but I'm unsure whether the Vercel integration gets triggered as quickly as the GitHub action does so the time difference could be negligible.

## Validation

I checked in #994 (branched off after ba407fe) that the workflow was actually building and deploying from source by turning off the docs-build action and changing the source. The path to #982 is clear!

## Possible issues

I have pretty high confidence in the deploy-production workflow because [it does the same thing as @primer/css](https://github.com/primer/css/blob/main/.github/workflows/deploy_production.yml) but if it doesn't deploy correctly when this is merged it either won't deploy at all (which we can fix) or if deploys a broken one we can promote a previous deployment to production in the Vercel UI (which happens instantly) and then fix it.

## Follow-up

In a few weeks to make sure all working PRs are up to date with this:

- [ ] Remove build command from [settings](https://vercel.com/primer/view-components/settings)
- [ ] Remove `now-build` script hack from `docs/package.json`

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -15,6 +15,33 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          token: ${{ secrets.GPR_AUTH_TOKEN_SHARED }}
+      - name: Setup Ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.7.x
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - run: yarn
+      - uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: gems-deploy-preview-${{ hashFiles('**/Gemfile.lock') }}
+
+      - name: Generate static files
+        run: |
+          gem install bundler:2.2.9
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+          bundle exec rake utilities:build
+          bundle exec rake docs:build
+          bundle exec rake static:dump
+        env:
+          SKIP_STORYBOOK_PRELOAD: 1
+
       - uses: chrnorm/deployment-action@v1.2.0
         name: Create GitHub deployment
         id: deployment
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -1,0 +1,50 @@
+name: Deploy
+on:
+  push:
+    branches-ignore:
+      - 'main'
+    paths:
+      - 'app/**'
+      - 'docs/**'
+      - '.github/workflows/deploy*.yml'
+      - 'package.json'
+jobs:
+  deploy-preview:
+    if: ${{ github.repository == 'primer/view_components' }}
+    name: Preview
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: chrnorm/deployment-action@v1.2.0
+        name: Create GitHub deployment
+        id: deployment
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          environment: Preview
+
+      - name: Vercel Action
+        uses: amondnet/vercel-action@v20
+        id: vercel-action
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          vercel-token: ${{ secrets.VERCEL_TOKEN_SHARED }}
+          github-comment: false
+          vercel-org-id: ${{ secrets.VERCEL_ORG_ID_SHARED }}
+          vercel-project-id: ${{ secrets.VERCEL_PROJECT_ID }}
+
+      - name: Update deployment status (success)
+        if: success()
+        uses: chrnorm/deployment-status@v1.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          environment_url: ${{ steps.vercel-action.outputs.preview-url }}
+          state: "success"
+          deployment_id: ${{ steps.deployment.outputs.deployment_id }}
+
+      - name: Update deployment status (failure)
+        if: failure()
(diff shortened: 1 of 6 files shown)


pouretrebelle:
This has been very frustrating. Notes for the morning:
- There's an enduring issue with the vercel-action being triggered twice which was introduced in 25109bf, I need to do some elimination to figure out why
- It would be nice to build the docs site in the action and deploy it statically, to save doing yarn installs on each side (also I think our actions are faster than Vercel)
  - The build is triggered on Vercel by a UI setting that is used for the whole project, so we can't fuck with it
  - Although it's [potentially overridden](https://vercel.link/unused-build-settings) by some deprecated setting we're already using?
- Same prob with the working directory, if we keep it in the config/UI it has to download all the files in the cwd (~11k), it would be better to specify the working directory in the action config

docs/vercel.json
```diff
@@ -1,13 +1,14 @@
 {
-  "version": 2,
-  "name": "primer-viewcomponents",
   "builds": [
     {
       "src": "package.json",
-      "use": "@now/static-build",
+      "use": "@vercel/static-build",
       "config": {"distDir": "public"}
```
pouretrebelle, line 6: I didn't change this but FYI there's no way for us to deploy  _just_ the public folder because [Vercel requires all static content to be in a `public` folder](https://vercel.com/guides/upgrade-to-zero-configuration#frameworks-with-zero-configuration); which is why we're deploying the whole docs folder but ignoring everything but the public folder (e531b7e1cab0d22a326cf337c1ce4c26c354b2a0)

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,85 @@
...
+on:
+  push:
+    branches-ignore:
+      - 'main'
+    paths:
+      - 'app/**'
+      - 'docs/**'
+      - '.github/workflows/deploy*.yml'
+      - 'package.json'
+      - '.vercelignore'
```
pouretrebelle, line 11: Are there other root paths we'd want to include here which could change the docs output?

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,85 @@
...
+        id: deployment
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          environment: Preview
+
+      - uses: actions/checkout@v2
+      # - name: Setup Ruby
+      #   uses: actions/setup-ruby@v1
+      #   with:
+      #     ruby-version: 2.7.x
```
manuelpuyol, line 28: should we delete comments in the `yml` files?

docs/vercel.json
```diff
@@ -1,13 +1,14 @@
...
     }
   ],
   "rewrites": [{"source": "/view-components(/.*)?", "destination": "/$1"}],
-  "redirects": [{"source": "/", "destination": "/view-components"}]
+  "redirects": [{"source": "/", "destination": "/view-components"}],
+  "github": {
+    "enabled": false
+  }
```
manuelpuyol, line 13: what is this?

manuelpuyol (approved):
looks good, but we have to make sure it doesn't break the `primer.style` redirects

.github/workflows/deploy-preview.yml
pouretrebelle: I'm uncommenting them in https://github.com/primer/view_components/pull/996, left them here as a sign they will be used imminently, and to leave more of a paper trail 🔍

docs/vercel.json
pouretrebelle: It stops the Vercel bot from deploying (the thing that makes the comments on PRs) 6ee245630b5b0138997ee9b21e2cf1a132301fe3

PR #996, opened 2022-01-27
Remove generated docs content from repo

pouretrebelle:
Removing the generated markdown docs from the repo. Since #989 they're generated as part of the deployment process so we don't need them in git history.

I've left in [the docs-build workflow](https://github.com/primer/view_components/blob/v0.0.67/.github/workflows/docs-build.yml) because it updates other files (those in /static and css utilities) which are still added to the repo. See https://github.com/primer/view_components/issues/982 for progress on the rest of these generated files; the intention is to remove all of them and then make the docs-build workflow redundant.

## Checks

- [x] ~~Running `script/dev` without the components+adr folder works as expected~~
    This doesn't work because the .md files aren't present when the doctocat script is initiated. I added static file generation to `script/setup` to get around this, but open to other ideas
- [x] Fresh install, cd this branch, running `script/setup` and then `script/dev` works as expected

### Before merge

- [x] Remove temp commit to force deployment

### Follow-up

- [ ] I think this change may break [this test](https://github.com/github/view_component/blob/main/.github/workflows/ci.yml#L96-L107) in the view_component library, we'll need to add static file generation to that as well


jonrohan (approved):
Nice

PR #1000, opened 2022-02-02
Remove CSS utilities from Blankslate

hectahertz:

This is part of the effort to remove all usages of utility classes to improve consistency and prevent specificity bugs. More context on https://github.com/github/primer/issues/585. Also includes changes the HTML structure and style tweaks to better match Primer CSS. See the [corresponding `primer/css` pr](https://github.com/primer/css/pull/1919). 

#### Before

before

#### After

after

I decided to go for a minimally invasive set of changes so we can make these changes in all components with the least amount of friction possible. No breaking changes.

I found a couple inconsistencies that I'd like to mention.

#### Heading font size

As per [the discussions on this issue](https://github.com/primer/view_components/pull/969) it seems that we don't want to default to any size and keep the default heading styling. @ashygee 

`Blankslate` adds a `h2` utility class by default but, because we want to stop this behavior, I didn't add it on `primer/css`, and removed the utility class. This will make all headings render with their default level styles.

> **_NOTE_**  This PR is redundant with https://github.com/primer/view_components/pull/969 cc @pouretrebelle

### Are additional changes needed?

- [X] No, this PR should be ok to ship as is. 🚢


manuelpuyol (approved):
amazing!

PR #1001, opened 2022-02-03
Remove preview deployments for Dependabot-generated PRs

pouretrebelle:
Since #989 we've been using a GitHub action for deployment of preview
apps and production. The preview action fails on Dependabot PRs because
Dependabot-run actions have read-only GitHub tokens and don't have
access to secrets, so they can't initiate the Deployment action step.

This turns off the deploy-preview workflow for actions initiated by
Dependabot so the whole CI doesn't fail (the same solution as #980).

We _should_ have preview deployments for dependency PRs but it's almost
impossible with this actions workflow. In this repo all PRs have to be
up to date with the base branch before merging though, so reviewers are
often forced to click 'update branch' manually before merging, which
will trigger a deploy because it triggers a merge by the user who
clicked the button and therefore that user triggers the actions. This
workflow can be used by reviewers who need to check the docs site for
changes. It's not ideal but preferable to having checks fail on all
dependency PRs.

For more info on the secrets issue see:
- https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544
- https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-797125425
- https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/


pouretrebelle:
>  In this repo all PRs have to be up to date with the base branch before merging though, so reviewers are often forced to click 'update branch' manually before merging, which will trigger a deploy... this workflow can be used by reviewers who need to check the docs site for changes.

I feel like this should be documented somewhere but I don't know where would be a good spot for it 🤔

hectahertz (approved):

hectahertz:
> > In this repo all PRs have to be up to date with the base branch before merging though, so reviewers are often forced to click 'update branch' manually before merging, which will trigger a deploy... this workflow can be used by reviewers who need to check the docs site for changes.
> 
> I feel like this should be documented somewhere but I don't know where would be a good spot for it 🤔

We could add a new section to [the contributors readme](https://github.com/primer/view_components/blob/main/contributor-docs/README.md). Likely not the best place for it, but I can't think of any others.

pouretrebelle:
I realised this is also an issue for primer/react and primer/css, so I've added a note to the DI docs https://github.com/github/design-infrastructure/pull/2267

PR #1006, opened 2022-02-09
avatar link arguments

jamieshark:
This PR adds an option for passing arguments directly to the rendered `a` tag in the Avatar component.

**Situation**:
This came up as an issue when I tried to place a margin on the Avatar component, but it distorted the image. When trying to use the `"mr-2"` class, the linter will not allow it, asking to prefer system arguments instead.

**Solution**:
Create a `link_arguments` param that will allow direct placement of link arguments to the rendered `LinkComponent`


camertron (approved):

jamieshark:
Hey @camertron - not sure if there's something else I need to do to merge these changes, but seems like as a first-time contributor there are some checks that a maintainer has to do first?

test/components/beta/avatar_test.rb
```diff
@@ -66,55 +66,51 @@ def test_squared_avatar
   def test_renders_link_wrapper
     render_inline(Primer::Beta::Avatar.new(src: "https://github.com/github.png", alt: "github", href: "#given-href"))
 
-    assert_selector("a.avatar") do |(a)|
+    assert_selector("a") do |(a)|
       assert_equal("#given-href", a["href"])
-      assert_selector("img")
-      refute_selector("img.avatar")
+      assert_selector("img.avatar")
```
camertron, line 71: Is it just me or did this change from `refute_selector` to `assert_selector`, and if so, why? I don't see any HTML changes?

app/components/primer/beta/avatar.rb
```diff
@@ -65,15 +70,15 @@ def initialize(src:, alt:, size: DEFAULT_SIZE, shape: DEFAULT_SHAPE, href: nil,
...
-          "lh-0" => href # Addresses an overflow issue with linked avatars
+          "circle" => shape == DEFAULT_SHAPE
         )
       end
 
       def call
         if @href
-          render(Primer::LinkComponent.new(href: @href, classes: @system_arguments[:classes])) do
-            render(Primer::BaseComponent.new(**@system_arguments.except(:classes))) { content }
+          (@link_arguments[:classes] ||= []) << "lh-0" # Addresses an overflow issue with linked avatars
```
camertron, line 79: I believe we usually do default classes like this in the constructor.

app/components/primer/beta/avatar.rb
```diff
@@ -65,15 +70,15 @@ def initialize(src:, alt:, size: DEFAULT_SIZE, shape: DEFAULT_SHAPE, href: nil,
...
+          render(Primer::LinkComponent.new(href: @href, **@link_arguments)) do
+            render(Primer::BaseComponent.new(**@system_arguments)) { content }
```
camertron, line 81: Do we want to keep `.except(:classes)` here to maintain backwards compatibility?

test/components/beta/avatar_test.rb
```diff
@@ -66,55 +66,51 @@ def test_squared_avatar
...
   end
 
   def test_defaults_circle_link_wrapper
     render_inline(Primer::Beta::Avatar.new(src: "https://github.com/github.png", alt: "github", href: "#"))
 
-    assert_selector("a.avatar.circle") do
-      assert_selector("img")
-      refute_selector("img.circle")
+    assert_selector("a") do
+      assert_selector("img.avatar.circle")
```
camertron, line 79: Also wondering why this changed from `refute_selector` to `assert_selector` without accompanying HTML changes.

camertron (changes requested):
Hey @jamieshark, looks like a number of other changes were made here, so I've left a few additional comments 😄

jamieshark:
I think I'm going to close this PR with changes for now since we have found a workaround on our end, and I am nervous to make these API changes to this component, but not others that render nested components. I can re-open in the future if this is a more widely adopted pattern!

PR #1008, opened 2022-02-11
feat: updates autocomplete api to prevent accessibility violations

lindseywild:
Closes https://github.com/github/accessibility/issues/596.

## Description
Updates the AutoComplete API for an easier developer experience and to prevent accessibility violations.

## Changes
1. Removes the need to pass in an `input` and `label` slot - these will now be programmatically added based on the arguments that a user passes in. A user can still add an `input` slot if desired, for use cases such as adding a custom class. The `label` slot has been removed to prevent misuse, but can be added back if we find the need for custom classes or another need.
2. Adds `is_label_visible` parameter - allows a user to visually hide a label, but keeps it for screen readers
3. Removes `aria-label` and `aria-labelledby` options - these should not be needed if the component is used as stated in the documentation.

## Validation Steps
1. Navigate to https://view-components-n0olq29t2-primer.vercel.app/view-components/components/beta/autocomplete
2. Open each example and test for keyboard functionality, inspect the markup and ensure it's accurate, test on a screen reader to ensure proper announcements
3. Open Storybook
4. Verify all stories and controls related to AutoComplete are functional

## Notes
- I am new to Ruby, Rails and contributing to Primer, so please let me know if there are better ways of doing what I have ❤️


andrialexandrou:
Did a behavioral check, looking good! Can you add an example for when we have icon *only* and no visible text label?

lindseywild:
@andrialexandrou Updated! https://view-components-n0olq29t2-primer.vercel.app/view-components/components/beta/autocomplete

I also fixed the failing tests 🙌

test/components/auto_complete_test.rb
```diff
@@ -5,99 +5,120 @@
...
   end
 
-  def test_renders_results_without_explicitly_calling_it
-    render_inline Primer::Beta::AutoComplete.new(src: "/url", input_id: "test-input", list_id: "my-list-id") do |component|
-      component.label(classes: "")
-      component.input(id: "test-input", name: "input")
+  def test_renders_label_without_explicitly_calling_it
+    render_inline Primer::Beta::AutoComplete.new(label_text: "Cool Fruits", src: "/url", input_id: "test-input", list_id: "my-list-id")
+
+    assert_selector("auto-complete[for=\"my-list-id\"][src=\"/url\"]") do
```
camertron, line 19: I think it's permissible to wrap these strings in single quotes because otherwise you have to escape all the double quotes.

```suggestion
    assert_selector('auto-complete[for="my-list-id"][src="/url"]') do
```

camertron (approved):
NIce, this looks great! 🎉

jscholes:
This looks great!  One comment:

> A user can still add an input slot if desired, for use cases such as adding a custom class.

This seems like it could significantly undermine this work, because the `input` is one of the key pieces of the puzzle.  Could the potential use cases be enumerated, to see if there could just be params for those too?  E.g. adding custom classes sounds like all itneeds is a string.

lindseywild:
> This looks great! One comment:
> 
> > A user can still add an input slot if desired, for use cases such as adding a custom class.
> 
> This seems like it could significantly undermine this work, because the `input` is one of the key pieces of the puzzle. Could the potential use cases be enumerated, to see if there could just be params for those too? E.g. adding custom classes sounds like all itneeds is a string.

Good call, James! I'm thinking we just completely remove the `input` slot, and if a user needs to add a custom class then we can add that parameter later. Right now it's just an assumption which may be incorrect. I'd like to err on the side of including less and then adding it in if it's truly needed.

jscholes:
@lindseywild I agree.

lindseywild:
Thanks for the reviews, @jscholes and @camertron! Based on feedback, I made a few updates to this PR that could use another review (sorry about that) -- this is the most notable commit https://github.com/primer/view_components/pull/1008/commits/35c65765741eafc5206775a5aaa62273e65b10e7 - the others are formatting and a changelog update.

jscholes:
@lindseywild To what extent should I review this at the current stage?  For example, looking at https://view-components-5oe1huxh2-primer.vercel.app/view-components/components/beta/autocomplete:

* The inputs don't have the required combobox mark-up, that I assume is the intent here for creating a single-select component with autocomplete.
* When suggestions are available, they can't be reached from the keyboard, and the mark-up is also not correct.

Are these things already on the radar, and will be addressed in follow-up PRs?  If not, I wouldn't want to suggest that the component is currently accessible, because it isn't.

The changes in this PR look good to me.  I note that there are two parameters, `input_id` and `list_id`, which I assume are there to allow developers to set the IDs of those generated elements?  Rather than to render those elements themselves, and pass in the IDs of the mark-up they've created.  If that is the case, the docs might be clearer as:

* `input_id`: ID to apply to the generated input.
* `list_id`: ID to apply to the generated list of suggestions.

jscholes:
Oh, and one more question: what overlap, if any, does this have with other components?  For example, there is currently a lot of work going on relating to single-selects, inside and outside forms.  Those themselves could have autocomplete, but don't need to.

lindseywild:
@jscholes Sorry about that, I should've been more explicit. This is just to update what is currently here, the fixes from @andrialexandrou in the auto-complete-element aren't published to this repo yet. This was to update the API before those go into place. This does not have any overlap with the other components currently (that I am aware of). Thanks!

jscholes:
@lindseywild Got it, thanks for the clarification!

andrialexandrou (approved):

PR #1011, opened 2022-02-12
Migrate Rails storybook server from Heroku to Azure

camertron:
This PR introduces a new workflow that uses [Kuby](https://getkuby.io) to build the storybook app in demo/ into a Docker image and deploy it to Azure. We now have our own Azure subscription for Primer, which includes an instance of the Azure Container Registry (ACR) and a Kubernetes cluster managed by the Azure Kubernetes Service (AKS). Any commit to main should kick off a build and deploy.

There are obviously quite a few details I'm skipping over here. I will be writing up some documentation next week for internal GitHub employees.


jonrohan (approved):
Not sure if this is verify-able, so 👍🏻

GrantBirki (approved):
:star: Looks good! I have one suggestion for your Actions workflow just because I have been bitten by this in the past. 

_Pinning 3rd party dependencies in your workflow_:

I would suggest using [this tool](https://github.com/mheap/pin-github-action) to pin all your 3rd party workflows to prevent your CI from breaking if they release a breaking change. Same concept to using a pinned Docker image rather than `:latest`.

:smile_cat:

PR #1017, opened 2022-02-16
Add missing alt attribute to README

khiga8:
### Context

As I was browsing around with [github-a11y chrome extension](https://github.com/khiga8/github-a11y,) I noticed we're missing `alt` attribute for this README image. I believe this is a decorative image so I'm setting the `alt` to empty.

Related:

- [erblint-github/blob/main/docs/rules/accessibility/image-has-alt.md](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/image-has-alt.md)
- https://webaim.org/techniques/alttext/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,10 @@ The category for changes related to documentation, testing and tooling. Also, fo
 
 ### Misc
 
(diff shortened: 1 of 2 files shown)


camertron (approved):
Thank you!

PR #1018, opened 2022-02-17
Fix link in system-arguments.md

lumaxis:


camertron (changes requested):
Hey @lumaxis, thanks for fixing this! Unfortunately the file you've modified is auto-generated. You'll want to edit [this one](https://github.com/primer/view_components/blob/f5930a3a1c8b9d5202d5cc2f12835fbd6437ac6f/app/components/primer/base_component.rb#L8) instead :)

lumaxis:
@camertron Whoops! Sorry about that 😅 I just tried to follow the "Edit this page on GitHub" button (that also actually points to a non-functional link) but I guess that's not reliable for these auto-generated pages.

I updated the original file too and ran `yarn build`, does this look right now?

camertron:
@lumaxis oops, sounds like something we should look at!

camertron (approved):

camertron:
@lumaxis would you mind adding an entry to the CHANGELOG so CI passes?

PR #1024, opened 2022-02-23
Fix sequence of content in Subhead

hectahertz:

When a heading, description, and action are rendered in the Subhead component, the content now follows logical order on screen readers.

- Fixes https://github.com/github/primer/issues/644
- Originally reported on [the subhead accessibility audit](https://github.com/github/primer/issues/358#issuecomment-996908226).

#### Before
Before
Accessibility tree
Before, accessibility tree

#### After

After
Accessibility tree
After, accessibility tree

I decided to reorder the elements on the DOM and use the [flexbox order property](https://www.w3.org/TR/css-flexbox-1/#order-accessibility) so elements are still rendered in the same way.

See [the primer/css pull request](https://github.com/primer/css/pull/1950) for more details.

### Are additional changes needed?

- [ ] Depends on https://github.com/primer/css/pull/1950

### DIFF
--- a/app/components/primer/subhead_component.html.erb
+++ b/app/components/primer/subhead_component.html.erb
@@ -1,5 +1,5 @@
 <%= render Primer::BaseComponent.new(**@system_arguments) do %>
   <%= heading %>
-  <%= actions %>
   <%= description %>
+  <%= actions %>
 <% end %>


camertron (approved):
🎉

PR #1032, opened 2022-02-27
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/view_components/commit/e07cbf22f622f56d881fda50532a7d9a2e83913a#commitcomment-67562045

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/view_components/commit/c1b56964586dc152fe17823b7b53c825acc6cf93

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

### DIFF
--- a/app/components/primer/markdown.rb
+++ b/app/components/primer/markdown.rb
@@ -93,7 +93,7 @@ module Primer
     #     <ol>
     #       <li>Michael Jackson</li>
     #       <li>Michael Bolton</li>
-    #       <li>Michael Buble</li>
+    #       <li>Michael Bublé</li>
     #     </ol>
     #
     #     <p>And an unordered task list:</p>
--- a/script/update-statuses-project.rb
+++ b/script/update-statuses-project.rb
@@ -11,7 +11,7 @@ statuses_json = JSON.parse(statuses)
 class QueryExecutionError < StandardError; end
 NOTE_SEPARATOR = " --- "
 
-module Github
+module GitHub
   GITHUB_ACCESS_TOKEN = ENV.fetch("GITHUB_TOKEN")
   URL = "https://api.github.com/graphql"
   HttpAdapter = GraphQL::Client::HTTP.new(URL) do
@@ -28,7 +28,7 @@ end
 
 # Project is a GraphQL wrapper for interacting with GitHub projects
 class Project
-  ProjectQuery = Github::Client.parse <<-'GRAPHQL'
+  ProjectQuery = GitHub::Client.parse <<-'GRAPHQL'
     query {
       repository(owner: "primer", name: "view_components") {
         project(number: 3) {
@@ -54,7 +54,7 @@ class Project
     }
   GRAPHQL
 
-  CreateCard = Github::Client.parse <<-'GRAPHQL'
+  CreateCard = GitHub::Client.parse <<-'GRAPHQL'
     mutation($note: String!, $projectColumnId: ID!) {
       addProjectCard(input:{note: $note, projectColumnId: $projectColumnId, clientMutationId: "pvc-actions"}) {
         __typename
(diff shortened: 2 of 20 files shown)


app/components/primer/markdown.rb
```diff
@@ -93,7 +93,7 @@ class Markdown < Primer::Component
     #     <ol>
     #       <li>Michael Jackson</li>
     #       <li>Michael Bolton</li>
-    #       <li>Michael Buble</li>
+    #       <li>Michael Bublé</li>
```
jsoref, line 96: I didn't see any specific reason to use the unaccented form, so this makes the entry consistent w/ another instance in this repository.

script/update-statuses-project.rb
```diff
@@ -11,7 +11,7 @@
 class QueryExecutionError < StandardError; end
 NOTE_SEPARATOR = " --- "
 
-module Github
+module GitHub
```
jsoref, line 14: Often projects don't like renaming modules. So I expect I'll drop this commit (I didn't include it for another repository).

test/components/beta/avatar_stack_test.rb
```diff
@@ -50,7 +50,7 @@ def test_renders_default_body
     end
 
     assert_selector(".AvatarStack") do
-      refute_selector(".toltipped")
+      refute_selector(".tooltipped")
```
jsoref, line 53: This is scary

test/components/border_box_component_test.rb
```diff
@@ -95,7 +95,7 @@ def test_strips_denied_system_arguments
   def test_renders_row_with_schemes
     { neutral: "gray", info: "blue", warning: "yellow" }.each_pair do |scheme, color|
       render_inline(Primer::BorderBoxComponent.new) do |component|
-        component.row(scheme: scheme) { "Row row row your boat" }
+        component.row(scheme: scheme) { "Row, row, row your boat" }
```
jsoref, line 98: Happy to drop, but technically this is the standard form for the song.

A recent addition to my tooling objects to `row row` as it's often an error. Quite happy to drop this or any other change.

jsoref (commented):
Most changes were automatically suggested by Google Sheets.

The `GitHub` and `row,` changes were very manual.

All fault is mine.

app/components/primer/markdown.rb
camertron: I'm pretty sure the accented form is more correct anyway, since that's the way his name is actually spelled.

script/update-statuses-project.rb
camertron: Actually I think renaming in this case is a good idea, since it is indeed spelled GitHub with the two capitals. We just need to make sure the code still works 😅

test/components/border_box_component_test.rb
camertron: 🤣  this is pretty funny, but acceptable.

camertron (approved):

test/components/beta/avatar_stack_test.rb
camertron: Hmm yeah would you mind reverting this commit? I think we'll have to handle it separately.

camertron:
Thanks for this @jsoref! Would you mind adding an entry to the CHANGELOG?

script/update-statuses-project.rb
jsoref: Ok. At some point I'll have to figure out which was the other repo w/ this and also offer it there...

camertron: We maintain a number of projects, but maybe you're thinking of github/view_component?

jsoref: https://github.com/primer/react/blob/cfbafe63e00a863d1940f8b3671bf5a21ef2691b/script/component-status-project/deploy.rb#L15

I didn't include that change in primer/react#1905

camertron: Hmm ok, interesting. Obviously this code isn't shared, but it's good to be consistent. Can I ask that you update primer/react to use the properly cased `GitHub`?

jsoref:
@camertron: done, I think

camertron:
I don't know what's going on with the "Build Docs" check, so I'm going to merge this PR and look into it separately.

PR #1034, opened 2022-02-28
Introduce `Primer::Alpha::Tooltip` component

khiga8:
Related to: https://github.com/github/accessibility/issues/670

## Context

This PR introduces a more accessible tooltip component, upstreamed after proven usage in dotcom. This also introduces a new "pattern" of having the related custom element code living in PVC rather than a separate repo. This is a decision we came to at  UI engineering code review. If we change our mind later, we can always move it.

## Upstream decisions

- I decided to house the tooltip CSS in the shadow DOM upon discussion with @koddsson. I also referenced the [encapsulation example shown in viewcomponent.org](https://viewcomponent.org/guide/javascript_and_css.html#encapsulating-assets). 
- This introduces a new "pattern" of having the related custom element code living in PVC rather than a separate custo element repo. This is a decision we came to at  UI engineering code review. If we change our mind later, we can always move it.

## Thoughts

We should probably not expose this `Tooltip` in component docs, because I fear misuse and tooltip abuse. Instead, we should update API of components we think should support tooltip. Related: https://github.com/github/accessibility/issues/802 and https://github.com/github/primer/issues/438.


docs/content/components/beta/autocomplete.md
```diff
@@ -38,17 +38,69 @@ be used unless there is compelling reason not to. A placeholder is not a label.
 
 ## Slots
 
-### `Label`
+### `Icon`
```
khiga8, line 41: I have no idea why this file got updated.

khiga8:
`Tooltip` looks good based off it's rendering in local dev docs.

Though, it looks like vercel preview uses the JS pulled in from npm package. How can we have vercel preview deploy our new JS code?

khiga8:
The test failure I'm getting is for `blankslate` which is a totally unrelated component. Why?

docs/content/components/beta/autocomplete.md
khiga8: Addressing with https://github.com/primer/view_components/pull/1036

package-lock.json
```diff
@@ -0,0 +1,6019 @@
+{
```
owenniblock, line 1: I think this file is untracked on my codespace instance? Do we need this here.

owenniblock (commented):
This looks great! I think you still need to copy over your `.js` test code from the experimental component. I guess this may be non-trivial (are there any JS or TS tests in PVC?)

Previously we'd been advised to move the web component to `primer/behaviors` (so they can be shared with React at some point) so maybe that's an option but it would increase the difficulty on releasing and fixing issues 🤷 So I can understand why we're putting the `.ts` in here.

package-lock.json
khiga8: Good call!

khiga8:
@owenniblock Thanks for the review! We do have browser tests to test behaviors for JS components, like this [test/system/tooltip_test.rb](test/system/tooltip_test.rb)!

test/components/alpha/tooltip_test.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class PrimerAlphaTabPanelsTest < Minitest::Test
```
khiga8, line 5: Todo: change name

demo/Gemfile.lock
```diff
@@ -108,13 +108,6 @@ GEM
       signet (~> 0.15)
     helm-cli (0.3.0)
       helm-rb (~> 0.1)
-    helm-rb (0.2.1-aarch64-linux)
-    helm-rb (0.2.1-arm-linux)
-    helm-rb (0.2.1-arm64-darwin)
-    helm-rb (0.2.1-arm64-linux)
-    helm-rb (0.2.1-ppc64le-linux)
-    helm-rb (0.2.1-s390x-linux)
-    helm-rb (0.2.1-x86-linux)
```
hectahertz, line 117: I think this should not be here; when I rebuild the PR locally these changes get undone.

test/components/alpha/tooltip_test.rb
```diff
@@ -0,0 +1,32 @@
...
+    assert_selector("tool-tip[for='someButton'][data-type='label']", text: text, visible: false)
+  end
+
+  def test_renders_type_description
+    text = "This describes the button"
+    render_inline(Primer::Alpha::Tooltip.new(type: :description, for_id: "someButton", direction: :n, text: text))
+    assert_selector("tool-tip[for='someButton'][data-type='description']", text: text, visible: false)
+  end
+
+  def test_raises_if_text_is_not_string
```
hectahertz, line 26: Thanks for including this one, good that we have this covered.

hectahertz:
The docs are wonderful, great job!

I feel comfortable with it as it is, I see it [meets all the criteria to be in Alpha](https://primer.style/contribute/component-lifecycle#alpha) and it looks good overall.

After reading the docs carefully I'm convinced we should list it on the component index.

I understand the hesitation as we could start seeing unintended usages, but I'd rather lint against those than remove it from the docs site index. The Tooltip will be searchable in the codebase anyway, so I'd rather have an explicit linter against this.

khiga8:
Thank you so much @hectahertz for thorough review :) 

> I'd rather lint against those than remove it from the docs site index. 

I totally agree something like an ERB lint rule would be most ideal to block bad tooltip usage! I investigated this earlier in the tooltip work and found there are limitations with using a static code linter against something like: _Don't set tooltips on static elements._

As an example, there might be ERB code like:

```.erb
<%= render_inline(Issues::ProgressComponent.new(size: :small, id: "progress-1")) %>
<%= render_inline(Primer::Alpha::Tooltip.new(for_id: "progress-1", text: "This labels the progress component", type: :description)) %>
```

With an ERB linter, it's impossible to know what the HTML output of this `Issues::ProgressComponent` is, since we need the Rails app running to know what the final markup is. 

Though we don't have an ERB lint rule, we do have a [couple custom axe rules](https://github.com/github/github/blob/master/app/assets/modules/github/axe-custom-rules.ts#L217 ) for this tooltip. The downside is axe is non-blocking and it depends on developer to run the app locally.

I'd be up for discussing other options for how we can flag bad tooltip usage without ERB! One idea that I explored in the past was having accessibility checks in [existing Rails controller tests](https://github.com/github/github/pull/204623). I thought this could be too wild so I abandoned it 😛.

If we can have a reliable way to block bad tooltip usage, I'd be up for exposing this to consumers!

khiga8:
@koddsson unrelated to the changes, might u know what this message is (when I run `npm run prepare`)?:

```
app/components/primer/primer.js → app/assets/javascripts/primer_view_components.js...
(!) `this` has been rewritten to `undefined`
https://rollupjs.org/guide/en/#error-this-is-undefined
app/components/primer/alpha/tooltip.js
1: var __classPrivateFieldSet = (this && this.__classPrivateFieldSet) || function (receiver, state, value, kind, f) {
                                 ^
2:     if (kind === "m") throw new TypeError("Private method is not writable");
3:     if (kind === "a" && !f) throw new TypeError("Private accessor was defined without a setter");
``` 

js seems to be working fine so I'm not sure if I need to worry about it

koddsson:
> @koddsson unrelated to the changes, might u know what this message is (when I run `npm run prepare`)?:

Yeah, it's a warning that we ignore in other places where we use Rollup so I added the same ignore to this PR.

Context: https://github.com/rollup/rollup/issues/3788

hectahertz (approved):

PR #1050, opened 2022-03-07
chore: bumps autocomplete element package to 3.1.0

lindseywild:
Bumps autocomplete element package to [3.1.0](https://github.com/github/auto-complete-element/packages/15211).


khiga8:
You may want to rebase and rerun `npm run prepare`, which will regenerate the `primer_view_components.js` files.

jonrohan (approved):

app/components/primer/beta/auto_complete.rb
```diff
@@ -85,6 +93,35 @@ def initialize(label_text:, src:, list_id:, input_id:, is_label_visible: true, i
...
       def before_render
         results(classes: "") unless results
+        input unless input
+      end
+    end
+
+    # This component is part of `Primer::Beta::AutoCompleteComponent` and should not be
+    # used as a standalone component.
+    class Input < Primer::Component
+      # TODO: Do we fix this to always use `text`?
```
lindseywild: Yes - let's make it always `text`!

app/components/primer/beta/auto_complete.rb
```diff
@@ -85,6 +93,35 @@ def initialize(label_text:, src:, list_id:, input_id:, is_label_visible: true, i
...
+      DEFAULT_TYPE = :text
+      TYPE_OPTIONS = [DEFAULT_TYPE, :search].freeze
+
+      # TODO: Decide if we want to allow aria-labels here?
+
+      # @param type [Symbol] <%= one_of(Primer::Beta::AutoComplete::Input::TYPE_OPTIONS) %>
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+      def initialize(type: DEFAULT_TYPE, **system_arguments)
+        @system_arguments = deny_tag_argument(**system_arguments)
+        @system_arguments = deny_single_argument(:autofocus, "autofocus is not allowed for accessibility reasons.", **system_arguments)
```
lindseywild: Should we be more explicit about `accessibility reasons`? Maybe point to a resource if one exists?

lindseywild:
@owenniblock Thanks for making the API updates! Just a note so we remember to update the Storybook stories and tests once we're happy with the changes 😄

app/components/primer/beta/auto_complete.rb
```diff
@@ -32,14 +32,41 @@ class AutoComplete < Primer::Component
...
       # @example Default
       #   <%= render(Primer::Beta::AutoComplete.new(label_text: "Fruits", src: "/auto_complete", input_id: "fruits-input-1", list_id: "fruits-popup-1", position: :relative)) %>
       #
       # @example With Non-Visible Label
       #   <%= render(Primer::Beta::AutoComplete.new(label_text: "Fruits", src: "/auto_complete", input_id: "fruits-input-2", list_id: "fruits-popup-2", is_label_visible: false, position: :relative)) %>
       #
-      # @example With Custom Classes for the Results
+      # @example With Custom Classes for the Input
       #   <%= render(Primer::Beta::AutoComplete.new(label_text: "Fruits", src: "/auto_complete", input_id: "fruits-input-3", list_id: "fruits-popup-3", position: :relative)) do |c| %>
+      #     <% c.input(classes: "special-form", autofocus: true) %>
```
khiga8, line 65: should we remove `autofocus: true` in this example given the guidance we provide?

> "autofocus is not allowed for accessibility reasons"

app/components/primer/beta/auto_complete.rb
```diff
@@ -32,14 +32,41 @@ class AutoComplete < Primer::Component
...
       }
 
+      # Customizable input used to search for results.
+      # It is preferred to use this slot sparingly - it will be created by default if not explicity added.
+      #
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+      renders_one :input, lambda { |**system_arguments|
+        deny_tag_argument(**system_arguments)
+        deny_single_argument(system_arguments[:autofocus], "autofocus is not allowed for accessibility reasons. See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus#accessibility_considerations for more information.", **system_arguments)
+        deny_single_argument("aria-label", "instead of `aria-label`, include `label_text` and set `is_label_visible` to `false`.", **system_arguments)
```
khiga8, line 42:
```suggestion
        deny_single_argument("aria-label", "instead of `aria-label`, include `label_text` and set `is_label_visible` to `false` on the component initializer.", **system_arguments)
```

app/components/primer/beta/auto_complete.rb
```diff
@@ -32,14 +32,41 @@ class AutoComplete < Primer::Component
         Primer::BaseComponent.new(**system_arguments)
       }
 
+      # Customizable input used to search for results.
+      # It is preferred to use this slot sparingly - it will be created by default if not explicity added.
```
khiga8, line 36: In what scenarios could it make sense to use this `input` slot? Would it be when someone wants to set custom `system_arguments` on `input`?

khiga8 (commented):
I added a couple comments as someone very new to the autocomplete work! Thanks for starting this work 💪

I agree with @lindseywild suggestion of explaining why autofocus is bad, maybe in the `@accessibility` section. In addition, we might want to explain the purpose of the optional clear button.

app/components/primer/beta/auto_complete.rb
owenniblock: Yes, we have examples where this is necessary in dotcom (for example when custom classes are applied).

yarn.lock
```diff
@@ -39,10 +39,10 @@
     minimatch "^3.0.4"
     strip-json-comments "^3.1.1"
 
-"@github/auto-complete-element@^3.0.2":
-  version "3.0.2"
-  resolved "https://registry.npmjs.org/@github/auto-complete-element/-/auto-complete-element-3.0.2.tgz"
-  integrity sha512-2N6SP/WOvPPnMm5E0uq21AyCxsJ0d6cBxBZ0yVNpvgeaNgPNI8gR1h54c2Ao4iQtvFce6eWdIezBXLDPMPbDag==
+"@github/auto-complete-element@^3.1.0":
+  version "3.1.0"
+  resolved "https://registry.yarnpkg.com/@github/auto-complete-element/-/auto-complete-element-3.1.0.tgz#afd5ee64a17b4507c42136cfd790103217009b50"
```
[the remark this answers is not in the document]
owenniblock: Someone probably used `npm-cli` to add it originally but the docs say to use `yarn` - I don't think it matters but it's probably better using `yarn` while that's the documented tooling.

owenniblock:
We need to make some changes. Currently the `Invite a member to actions` dialog looks like this:

Dialog containing AutoComplete control, icon sits above the input and right side of input has unexpected border-radius

To fix the icon issue (icon sits above input) - we can move the icon below the input in the markup (it's `aria-hidden` so I think that should be fine).

The `border-radius` needs some CSS changes:

```
.input-group .form-control:first-child, .input-group-button:first-child .btn {
border-top-right-radius: 0;
border-bottom-right-radius: 0;
}
```

This element is no longer the `first-child` because we've added a label.

mperrotti:
Screen Shot 2022-03-11 at 11 06 02 AM
Screen Shot 2022-03-11 at 11 06 15 AM

This diverges from how we're handling input icons and actions in [primer/react](https://github.com/primer/react). We're putting icons and actions inside the text input border:

image

There's a new PR for text input trailing actions: https://github.com/primer/react/pull/1947
We already support icons inside of text inputs, which you can see in the TextInput docs: https://primer.style/react/TextInput

app/components/primer/beta/auto_complete.rb
lindseywild: Yes! Sorry I added this to test if the `deny_single_argument` tags were working (looks like this is outdated now anyway!)

app/components/primer/beta/auto_complete.rb
```diff
@@ -32,14 +32,45 @@ class AutoComplete < Primer::Component
...
+      renders_one :input, lambda { |**system_arguments|
+        sanitized_args = deny_tag_argument(**system_arguments)
+        sanitized_args = deny_single_argument(:autofocus, "autofocus is not allowed for accessibility reasons. See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus#accessibility_considerations for more information.", **sanitized_args)
+        deny_aria_key(
+          :label,
+          "instead of `aria-label`, include `label_text` and set `is_label_visible` to `false` on the component initializer.",
+          **sanitized_args
+        )
+        sanitized_args[:id] = @input_id
+        sanitized_args[:name] = @input_id
```
khiga8, line 48: Should we also explicitly deny `id` and `name` given that it is ignored and overridden by `@input_id`?

lindseywild: Ooh that is a good idea - I say we go with your suggestion and revisit if needed!

khiga8: Ok, I added an explicit deny for `id` and `name` in https://github.com/primer/view_components/pull/1050/commits/7fd497d24fc6623f2384e586fc2c5d73ca8a9e19 and tests in https://github.com/primer/view_components/pull/1050/commits/d2e7cfcad7a16d4a1096c3645f58eff104aaeeef. 

On second thought, I feel like we should allow `name` to be overridden since `name` and `id` have different functions. Down to revisit this later though!

owenniblock: I've added the `input_name` option back in 🙂

lindseywild:
Additional action items were added to the description of this PR now that the CSS PR has been approved.

CC @khiga8 for visibility!

khiga8:
I removed all the `position: relative` that were being set in the examples because I don't know what it's for: https://github.com/primer/view_components/pull/1050/commits/cf8cfd508d5d8f855993cd032637e8532bafd4de. 

If it's actually necessary, feel free to add it back in, but maybe it should be a default part of the API.

lindseywild:
@khiga8 @owenniblock @smockle I believe all of the work is now complete in this PR - If you all would be so kind as to check it over, review the docs and that they make sense, any other nitpicks... Please let me know!

khiga8:
I'd encourage everyone to double check swaps now that the API has been updated with new CSS classes.

We'll need to make a couple minor adjustments. 

For my `sponsorship` swap I noticed:
- Width is overflowing outside constraining box when stacked.
- When there's a [form-group](https://primer.style/css/components/forms#form-groups) parent, any nested `form-control` turns grey by default but white on focus. However, autocomplete has an embedded icon area outside the `input` needs which also change to white. Maybe we need to define something like:

```
.form-group {
  .autocomplete-body&:focus-within {
    background-color: var(--color-canvas-default);
  }
}
```

screenshot of swapped autocomplete with odd styles

lindseywild:
I added a [to do list](https://github.com/github/github/pull/214414) at the top of the draft PR in dotcom of what is left and to make sure we verify all instances are accurate!

smockle:
From @khiga8 in https://github.com/primer/view_components/pull/1050/#issuecomment-1079309276:
> I'd encourage everyone to double check swaps now that the API has been updated with new CSS classes.

Super advice! I found numerous issues. I pushed https://github.com/github/github/commit/78e62821b19f42fdd3c56a02cd1a1f51269208cb to address a red-screen-of-death caused by a required/missing `with_icon` update, then I opened https://github.com/github/github/pull/214426 with style & polish updates.

khiga8:
Hello @primer/rails-reviewers! This is ready for review.

We have a branch in dotcom linked to these changes and have confirmed all instances look good: https://github.com/github/github/pull/214414.  Through our swaps, we discovered additional CSS changes need to be made so we'll temporarily house those in dotcom while we work on getting it in Primer CSS.

camertron (approved):
LGTM!

Issue #1049, opened 2022-03-07
Docs site error: 'Can't resolve '@primer/components''


hectahertz:
# Problem

When running `script/dev` on a fresh install, Webpack fails to generate the bundle as it's missing the `@primer/components` dependency

PR #1051, opened 2022-03-07
Fix missing `@primer/components` dependency

hectahertz:
Replaced with `@primer/react` as `@primer/components` has been renamed.

Fixes https://github.com/primer/view_components/issues/1049


jonrohan (approved):
nice 👍🏻

PR #1058, opened 2022-03-11
Add `Tooltip` support to `Button`

hectahertz:

The API of `Button` should be updated so `Tooltip` can be rendered via the component arguments.

- Closes https://github.com/github/primer/issues/776

We decided to use a slot to render the `Tooltip` as it allows for explicit control of the defaults, which we want to ensure proper integration and an accessible outcome. We will follow the same pattern for the rest of the components that integrate `Tooltip`

For example, we set `type: :description` as the default, as most buttons will have a label already.

### Docs
Screen Shot 2022-03-15 at 13 03 10

Screen Shot 2022-03-17 at 14 33 00


app/components/primer/button_component.rb
```diff
@@ -56,6 +56,21 @@ class ButtonComponent < Primer::Component
     }
     alias counter trailing_visual_counter # remove alias when all buttons are migrated to new slot names
 
+    # `Tooltip` that appears on mouse hover or keyboard focus over the button. Use tooltips sparingly and as a last resort.
+    # Consult the <%= link_to_component(Primer::Alpha::Tooltip) %> documentation for more information.
+    #
+    # @param system_arguments [Hash] Same arguments as <%= link_to_component(Primer::Alpha::Tooltip) %>.
```
khiga8, line 62:
```suggestion
    # @param type [Symbol] (:description) <%= one_of(Primer::Alpha::Tooltip::TYPE_OPTIONS) %>
    # @param system_arguments [Hash] Same arguments as <%= link_to_component(Primer::Alpha::Tooltip) %>.
```

This adds documentation for the default `type`

app/components/primer/button_component.rb
```diff
@@ -56,6 +56,21 @@ class ButtonComponent < Primer::Component
...
+    renders_one :tooltip, lambda { |**system_arguments|
```
khiga8, line 63:
```suggestion
    renders_one :tooltip, lambda { |text:, **system_arguments|
```

We may want to explicitly surface the required `text:` argument here. If we do this, we'll want to add `@param` to document this.

app/components/primer/button_component.rb
```diff
@@ -56,6 +56,21 @@ class ButtonComponent < Primer::Component
     }
     alias counter trailing_visual_counter # remove alias when all buttons are migrated to new slot names
 
+    # `Tooltip` that appears on mouse hover or keyboard focus over the button. Use tooltips sparingly and as a last resort.
+    # Consult the <%= link_to_component(Primer::Alpha::Tooltip) %> documentation for more information.
```
khiga8, line 60: I think we may want to call extra ttention to the `type:` here. Maybe something like:

```suggestion
    # **Important:** This tooltip defaults to `type: :description`. In a few scenarios, `type: :label` may be more appropriate.
    # Consult the <%= link_to_component(Primer::Alpha::Tooltip) %> documentation for more information.
```

app/components/primer/button_component.rb
```diff
@@ -96,6 +111,12 @@ class ButtonComponent < Primer::Component
     #     Button
     #   <% end %>
     #
+    # @example With tooltip
+    #   <%= render(Primer::ButtonComponent.new(id: "button-with-tooltip")) do |c| %>
+    #     <% c.tooltip(text: "Tooltip text") %>
+    #     Button
+    #   <% end %>
```
khiga8, line 118: Maybe this is overkill, but should we link to `Tooltip` documentation here too using `@code` and `@description`?

khiga8 (commented):
I added a few comments.
Overall this is looking good!!

app/components/primer/button_component.rb
hectahertz: Not overkill at all, good idea!

app/components/primer/button_component.rb
hectahertz: I thought about this, but then I saw `type` and `for_id` are required too.

I decided keeping the definition of which parameters are required on `Tooltip` will be easier to maintain. I'll leave it as is for now if that's ok.

khiga8: That makes sense 👍

khiga8 (approved):
Nice work 💪

jonrohan (approved):

PR #1062, opened 2022-03-17
Add `Tooltip` support to `IconButton`

hectahertz:

The API of `IconButton` should be updated so `Tooltip` can be rendered via the component arguments.

- Closes https://github.com/github/primer/issues/780

~~We decided to use a slot to render the `Tooltip` as it allows for explicit control of the defaults, which we want to ensure proper integration and an accessible outcome. We will follow the same pattern for the rest of the components that integrate `Tooltip`~~

~~For example, we set `type: :label` as the default, as for `IconButton` there's usually no other label.~~

After [some discussion in the PR comments](https://github.com/primer/view_components/pull/1062#issuecomment-1071065360), we decided to always include a `Tooltip` on `IconButton. The `aria-label` property is always required, and this will improve the experience for sighted users.

Hopefully, this approach will simplify the API and will keep the instances of `IconButton` consistent and accessible.

#### With `aria-label`
Screen Shot 2022-03-18 at 17 49 47

#### With `aria-label` and `aria-description`
Screen Shot 2022-03-18 at 17 49 54

Screen Shot 2022-03-18 at 17 58 45

Screen Shot 2022-03-18 at 17 58 58

Screen Shot 2022-03-18 at 17 59 03


app/components/primer/icon_button.rb
```diff
@@ -18,6 +18,24 @@ class IconButton < Primer::Component
       :danger => "btn-octicon-danger"
     }.freeze
     SCHEME_OPTIONS = SCHEME_MAPPINGS.keys
+
+    # `Tooltip` that appears on mouse hover or keyboard focus over the button. Use tooltips sparingly and as a last resort.
+    # **Important:** This tooltip defaults to `type: :label`. In a few scenarios, `type: :description` may be more appropriate.
+    # Consult the <%= link_to_component(Primer::Alpha::Tooltip) %> documentation for more information.
+    #
+    # @param type [Symbol] (:description) <%= one_of(Primer::Alpha::Tooltip::TYPE_OPTIONS) %>
```
khiga8, line 26:
```suggestion
    # @param type [Symbol] (:label) <%= one_of(Primer::Alpha::Tooltip::TYPE_OPTIONS) %>
```

app/components/primer/icon_button.rb
```diff
@@ -55,11 +84,5 @@ def initialize(icon:, scheme: DEFAULT_SCHEME, box: false, **system_arguments)
 
       validate_aria_label
```
khiga8, line 85: It looks like `Primer::IconButton` requires an `aria-label` , but if tooltip slot is set with `type: :label`, an `aria-labelledby` association is created so we no longer need an `aria-label`. We should update this validation logic. Then we may want to update the comments in `@accessibility`.

app/components/primer/icon_button.rb
khiga8: Actually maybe we should set a `DEFAULT_TOOLTIP_TYPE = :label` at top then set this to 
```suggestion
    # @param type [Symbol] (Primer::IconButton::DEFAULT_TOOLTIP_TYPE) <%= one_of(Primer::Alpha::Tooltip::TYPE_OPTIONS) %>
```
so we don't forget to update this if it changes in the future.

khiga8 (changes requested):

khiga8:
This may be a design-level decision but one thing that came up in a convo with @inkblotty was whether tooltips (whether it be `type: :label` or `type: :description`) should always be set on icon-only buttons so that sighted users have more context. [Here's a related audit issue](https://github.com/github/accessibility-audits/issues/2429).

What do you think? (I don't want to block this PR so I'd be ok following up on this in a separate PR too)

hectahertz:
> What do you think? (I don't want to block this PR so I'd be ok following up on this in a separate PR too) cc: @alliethu

Oh, I 100% agree.

I can see two approaches here;
- Making it required now and fixing those instances as we port them to this implementation.
- Keeping it optional for now, porting all instances to use it, and then introducing the restriction and fixing the missing labels.

I'd like to go for the second approach, as it will allow us to reduce heterogeneity in implementations sooner, but both can work fine.

inkblotty:
Re: @hectahertz in [comment]():
> - Making it required now and fixing those instances as we port them to this implementation.
> - Keeping it optional for now, porting all instances to use it, and then introducing the restriction and fixing the missing labels.

Is the `aria-label` or similar prop already required? If so, I don't think we would need to add a new prop for it. We could reuse the existing accessible name. If it isn't required, that's a larger discussion.

hectahertz:
> Re: @hectahertz in comment:
> Is the `aria-label` or similar prop already required? If so, I don't think we would need to add a new prop for it. We could reuse the existing accessible name. If it isn't required, that's a larger discussion.

You are 100% right, it is! This definitely makes things easier; we'll still be missing the label on the instances where `IconButton` is not being used yet, but it should not be such a big issue.

Now I'm thinking that the current API might not be the best; maybe a better solution would be to ask for an accessible label only once and decide at the framework level how we want to use it.

hectahertz:
Re: @inkblotty  in comment:
> Is the `aria-label` or similar prop already required? If so, I don't think we would need to add a new prop for it. We could reuse the existing accessible name. If it isn't required, that's a larger discussion.

You are 100% right, it is! This definitely makes things easier; we'll still be missing the label on the instances where `IconButton` is not being used yet, but it should not be such a big issue.

The proposed API  here might not be the best then; maybe a better solution would be to ask for an accessible label only once and control at the framework level how we want to use it (showing it on a tooltip or not, etc.). cc: @alliethu

Allowing users granular control of showing a tooltip or not, with custom content different to the aria-label might lead us to unpredictable outcomes. We might not need that level of customization.

app/components/primer/icon_button.rb
```diff
@@ -8,38 +8,22 @@ module Primer
   #   The `aria-label` should describe the action to be invoked rather than the icon itself. For instance,
   #   if your `IconButton` renders a magnifying glass icon and invokes a search action, the `aria-label` should be
   #   `"Search"` instead of `"Magnifying glass"`.
+  #   Internally the `aria-label` will be used for the `Tooltip` text. The `button` element is associated via `aria-labelledby`
+  #   to the `Tooltip` so it will serve as the `aria-label` for the `button`.
```
khiga8, line 12: We should also add support for tooltip with `type: :description` to support cases like the icon buttons we have in the markdown toolbar. We will probably want to add a note/example of that too.

I'm not sure what a good API for this. Maybe we will need the tooltip slot for that?

hectahertz: Added support for `aria-description`, let me know what you think!

app/components/primer/icon_button.rb
```diff
@@ -36,16 +38,35 @@ class IconButton < Primer::Component
     #     <% end %>
     #   <% end %>
     #
+    # @example With an `aria-description`
+    #   @description
+    #     If you need to have a longer description for the icon button, use the both the `aria-label` and `aria-description`
```
khiga8, line 43:
```suggestion
    #     If you need to have a longer description for the icon button, use both the `aria-label` and `aria-description`
```

khiga8 (approved):
Great work!! 🎉

khiga8:
What do you think of adding a note at top of [tooltip documentation](https://primer.style/view-components/components/alpha/tooltip) that points user to `IconButton`, `Button` (and whatever other component we add tooltip support to) and discourages direct use? Either in this PR or follow up :)

app/components/primer/icon_button.erb
```diff
@@ -0,0 +1,4 @@
+<%= render Primer::BaseButton.new(**@system_arguments) do -%>
+  <%= render Primer::OcticonComponent.new(icon: @icon) %>
+  <%= render Primer::Alpha::Tooltip.new(for_id: @system_arguments[:id], type: @tooltip_type, text: @tooltip_text, direction: @system_arguments[:tooltip_direction]) %>
```
manuelpuyol, line 3:
```suggestion
  <%= render Primer::Alpha::Tooltip.new(**@tooltip_arguments) %>
```

app/components/primer/icon_button.rb
```diff
@@ -36,16 +38,35 @@ class IconButton < Primer::Component
...
     # @param scheme [Symbol] <%= one_of(Primer::IconButton::SCHEME_OPTIONS) %>
     # @param icon [String] Name of <%= link_to_octicons %> to use.
     # @param tag [Symbol] <%= one_of(Primer::BaseButton::TAG_OPTIONS) %>
     # @param type [Symbol] <%= one_of(Primer::BaseButton::TYPE_OPTIONS) %>
     # @param box [Boolean] Whether the button is in a <%= link_to_component(Primer::BorderBoxComponent) %>. If `true`, the button will have the `Box-btn-octicon` class.
+    # @param aria-label [String] String that can be read by assistive technology. A label should be short and concise. See the accessibility section for more information.
+    # @param aria-description [String] String that can be read by assistive technology. A description can be longer as it is intended to provide more context and information. See the accessibility section for more information.
+    # @param tooltip_direction [Symbol] (Primer::Alpha::Tooltip::DIRECTION_DEFAULT) <%= one_of(Primer::Alpha::Tooltip::DIRECTION_OPTIONS) %>
     # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
     def initialize(icon:, scheme: DEFAULT_SCHEME, box: false, **system_arguments)
```
manuelpuyol, line 62:
```suggestion
    def initialize(icon:, scheme: DEFAULT_SCHEME, box: false, tooltip_direction: Primer::Alpha::Tooltip::DIRECTION_DEFAULT, **system_arguments)
```

app/components/primer/icon_button.rb
```diff
@@ -36,16 +38,35 @@ class IconButton < Primer::Component
...
       @icon = icon
 
       @system_arguments = system_arguments
+
+      @system_arguments[:id] ||= "icon-button-#{SecureRandom.hex(4)}"
+      @system_arguments[:tooltip_direction] ||= Primer::Alpha::Tooltip::DIRECTION_DEFAULT
```
manuelpuyol, line 68:
```suggestion
```

app/components/primer/icon_button.rb
```diff
@@ -54,11 +75,22 @@ def initialize(icon:, scheme: DEFAULT_SCHEME, box: false, **system_arguments)
...
+
+      if @tooltip_type == :description
+        @system_arguments.delete(:"aria-description")
+        @system_arguments[:aria].delete(:description) if @system_arguments.include?(:aria)
+        @tooltip_text = @aria_description
+      else
+        @system_arguments.delete(:"aria-label")
+        @system_arguments[:aria].delete(:label) if @system_arguments.include?(:aria)
+        @tooltip_text = @aria_label
       end
```
manuelpuyol, line 94:
```suggestion
      @tooltip_arguments = {
        for_id: @system_arguments[:id],
        direction: tooltip_direction
      }

      # If we have both an `aria-label` and a `aria-description`, we create a `Tooltip` with the description type and keep the `aria-label` in the button.
      # Otherwise, the `aria-label` is used as the tooltip text, which is the `aria-labelled-by` of the button, so we don't set it in the button.
      if @aria_label.present? && @aria_description.present?
        @system_arguments.delete(:"aria-description")
        @system_arguments[:aria].delete(:description) if @system_arguments.include?(:aria)
        @tooltip_arguments[:text] = @aria_description
        @tooltip_arguments[:type] = :description
      else
        @system_arguments.delete(:"aria-label")
        @system_arguments[:aria].delete(:label) if @system_arguments.include?(:aria)
        @tooltip_arguments[:text] = @aria_label
        @tooltip_arguments[:type] = :label
      end
```

khiga8:
@hectahertz is there anything blocking this?

jonrohan:
👋🏻 I can take over here, I'd like to update this with a lookbook preview

jonrohan (approved):

app/components/primer/icon_button.erb
```diff
@@ -0,0 +1,4 @@
+<%= render Primer::Beta::BaseButton.new(**@system_arguments) do -%>
+  <%= render Primer::OcticonComponent.new(icon: @icon) %>
+  <%= render Primer::Alpha::Tooltip.new(**@tooltip_arguments) %>
```
khiga8, line 3: @jonrohan

Sorry I just caught that this tooltip element is nested inside the `Button`. This should be adjacent to the `Button`. Similar with `Button` component and `Link` component.

This current nesting will result in the button (or link) being triggered when one tries to copy the tooltip text.

I think that in the case of tooltip, there will need to be a wrapper `div` or `span` element over the component and tooltip.

PR #1069, opened 2022-03-22
Add `Tooltip` support to `Link`

hectahertz:

The API of `Link` should be updated so `Tooltip` can be rendered via the component arguments.

- Closes https://github.com/github/primer/issues/781

We decided to use a slot to render the `Tooltip` as it allows for explicit control of the defaults, which we want to ensure proper integration and an accessible outcome. We will follow the same pattern for the rest of the components that integrate `Tooltip`.

For example, we set `type: :description` as the default, as all links will have a label already.

Very similar to:
- https://github.com/primer/view_components/pull/1058

### Docs

Screen Shot 2022-03-22 at 18 12 07

Screen Shot 2022-03-22 at 18 10 26

Screen Shot 2022-03-22 at 18 11 33

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ The category for changes related to documentation, testing and tooling. Also, fo
 
     _Hector Garcia_
 
+- Add `Tooltip` support to `Link`.
+
+    _Hector Garcia_
+
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


manuelpuyol (approved):

khiga8 (approved):
Looks great!

PR #1078, opened 2022-03-25
Graduate NavigationList from github.com to primer/view_components

camertron:
We've been using `NavigationList` for settings pages on github.com for a while now and are ready to graduate the component into primer/view_components.


camertron:
Oh shoot I forgot about the javascript! Coming soon.

jonrohan (approved):
Minus the catalyst/not-catalyst question, this looks good 👍🏻

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2017",
+    "experimentalDecorators": true,
```
keithamus, line 4: This shouldn't be needed I think.

hectahertz (approved):
Looking good! 👍

app/components/primer/alpha/navigation_list_element.ts
```diff
@@ -0,0 +1,39 @@
...
+        item.setAttribute('aria-expanded', 'true')
+      }
+    }
+
+    e.stopPropagation()
+  }
+}
+
+export {}
+
```
keithamus, line 29: Is this line necessary? It might be better to instead export the class on L1

PR #1079, opened 2022-03-29
Add responsive utitlity classes for text_align

lumaxis:
It's currently not possible to use responsive values for `text_align` even though the corresponding Primer CSS classes do exist: https://primer.style/css/utilities/typography#responsive-text-alignment

Screenshot 2022-03-29 at 14 24 36@2x


camertron (approved):
Awesome, thank you 🎉

PR #1082, opened 2022-03-29
[Temp] `ActionMenu` component markup

khiga8:
Temp development in PVC to take advantage of docs page.


khiga8:
cc: @jonrohan 

we added these temp commits to get vercel preview running with our local JS changes:
- https://github.com/primer/view_components/pull/1082/commits/b58003eba1f2831c8ce67b72b9e3c93b10da2cda
- https://github.com/primer/view_components/pull/1082/commits/21ab0379e6facae8b842dcda640a4c5e67da7733

That created this preview link: https://view-components-azqpabr6d-primer.vercel.app/view-components/. When we navigate to the `ActionMenu` page, the JS is working fine as demonstrated by pressing the `Menu` button. But when going directly to the URL or refreshing the `ActionMenu` page the JS doesn't run. Any ideas? 

Ultimately we want to move these commit changes to: https://github.com/primer/view_components/pull/1113 or some other separate PR.

lindseywild:
@khiga8 It seems moving the private variables into `connectedCallback` didn't work 🤔 I say we punt on that issue for now until we can get some of the other work further along, since this will be living in dotcom initially.

lindseywild:
@khiga8 I updated the examples to remove the ones without interactive elements. I don't see a situation where we'd have a non-interactive element... I also think we need to add the following which I can look at today:
- Example using `button`
- A way to pass the action for a `button` element to the item
- Support for `menuitemcheckbox` and `menuitemradio`

khiga8:
Sorry @lindseywild I totally missed your comment!

> I don't see a situation where we'd have a non-interactive element.

Could it be possible that this is the common markup in dotcom:

```.html.erb
<li role="none">
   <button type="button" role="menuitem" class="js-some-behavior"> Action 1 </button>
</li>
```

when the JS behavior doesn't need to be on a button element and could be like the following?

```.html.erb
<li role="menuitem" class="js-some-behavior">
   Action 1
</li>
```

Could it make sense to keep support for `li` with non-interactive elements to avoid needing to add an unnecessary interactive element when setting element-agnostic js class behaviors?

lindseywild:
@khiga8 Ah, good point - maybe we should keep it then! Once James provides feedback, I can look into adding this!

app/components/primer/alpha/action-menu-element.ts
```diff
@@ -0,0 +1,382 @@
...
+          position: absolute;
+          width: 160px;
+          background-color: var(--color-canvas-overlay);
+          z-index: 1000000;
+          border: $border-width $border-style var(--color-border-default);
+          border-radius: $border-radius;
+          box-shadow: var(--color-shadow-large);
+        }
+    `
+    document.querySelector('body')?.appendChild(style)
```
khiga8, line 61: Just temporarily keeping this here since `Overlay` isn't available yet in the previews. 
Can you confirm that all of this can be replaced by `Overlay`? @langermank

langermank: Yes, exactly! Overlay will replace any inline style

PR #1088, opened 2022-03-30
Contributor docs updates

pouretrebelle:
Back on the Docs train! Just some smallish edits before I add more pages.

Part of https://github.com/github/primer/issues/534


hectahertz (approved):
Looks good 👍

PR #1089, opened 2022-03-30
Configure .devcontainer to download primer/css on post-create

hectahertz:
This configures Codespaces to download primer/css after creating the environment, so we can add the folder to the workspace later and work on both changes on both projects simultaneously with hot reloading.

See [the docs for hot reloading](https://github.com/primer/view_components/blob/main/contributor-docs/updating-components.md#hot-reloading-local-only) for more context.


langermank (approved):
🚀

camertron (approved):

PR #1101, opened 2022-04-05
Fix view_component deprecation warning

Tonkpils:
Requiring `view_component/engine` was deprecated in `view_component` version 2.46.0 https://github.com/github/view_component/commit/d24b26ae2697685e930d863839d7f8834b9ca4a3

Requiring `view_component` will automatically load the engine when in the correct context.

I'm also wondering if we should `require "view_component"` inside `primer/view_components/engine` so that we don't have to require `view_component` explicitly from apps using `primer/view_component/engine`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,12 @@ The category for changes related to documentation, testing and tooling. Also, fo
 
   _Sam Morrow_
 
+### Misc
+
+- Update documentation to reflect deprecation of `require "view_component/engine"`
+
+  _Leo Correa_
+
 ## 0.0.71
 
 ### Updates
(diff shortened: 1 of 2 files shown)


camertron:
Hey @Tonkpils, thanks for this! LOL we should have done this a while ago 😅 

I do think it makes sense to `require "view_component"` inside primer/view_components/engine, and I would gladly accept a PR that makes the change. For now, I'll approve this one though.

camertron (approved):

camertron:
Oh I forgot to mention: would you mind adding a CHANGELOG entry?

Tonkpils:
👍  @camertron I've added the changelog change. I'll also work on a PR to require `view_components` inside the engine and make the change to the docs again to reflect that 😄

Issue #1106, opened 2022-04-08
LinkComponent adding extraneous whitespace


SamMorrowDrums:
Hi, I just wanted to mention that the `Primer::LinkComponent` appears to add spaces automatically, which is often desirable, or a no-op depending on where it is in the markup. However, there is no way to remove them which means that in order to use a period/full-stop at the end of a sentence, you are forced to directly use an `<a>` tag manually.

As a concrete example, these were both ViewComponents, however the latter I had to convert into an anchor tag because otherwise I couldn't end the sentence without `Privacy Policy .` rendering with the extra space.

So:

```erb
<%= render(Primer::LinkComponent.new(href: "#")) { "Terms of Use" } %> and
<a href="/copilot/signup#privacy-placeholder">Privacy Policy</a>.
```

Renders `Terms of use and Privacy Policy.` and:

 ```erb
<%= render(Primer::LinkComponent.new(href: "#")) { "Terms of Use" } %> and
<%= render(Primer::LinkComponent.new(href: "#")) { "Privacy Policy" } %>.
```

Renders `Terms of use and Privacy Policy .` for me. I looked and I can see `<a> text </a>` rather than `<a>text</a>` which I believe is because [the erb](https://github.com/primer/view_components/blob/main/app/components/primer/link_component.erb) does this:

```
<%= render Primer::BaseComponent.new(**@system_arguments) do %>
  <%= content %>
  <%= tooltip %>
<% end %>
```
And I think unfortunately to actually behave correctly it might need to look like this:

```
<%= render Primer::BaseComponent.new(**@system_arguments) do %><%= content %><%= tooltip %><% end %>
```

guo-chris:
👋 Hi, I think this is still an issue. I have a couple places ([issue](https://github.com/github/c2c-actions-experience/issues/6071); usages in github/github code [1](https://github.com/github/github/blob/e9bc4e88cbb57d761387d2cc4bab836909cb9442/app/components/actions/workflow_runs/old_attempt_banner_component.html.erb#L3-L6), [2](https://github.com/github/github/blob/e9bc4e88cbb57d761387d2cc4bab836909cb9442/app/views/checks/_checks_header.html.erb#L20-L23)) where I am seeing this regression in prod and can still reproduce it locally even with this fix. I've tried poking around at it and I'm not able to figure out what's wrong, so I'd appreciate any help with this. 

[...]

PR #1107, opened 2022-04-08
bugfix: ensure that no whitespace is added inside LinkComponent

SamMorrowDrums:
Closes https://github.com/primer/view_components/issues/1106

Ensure that there is a test to catch extra whitespace, and change the `.erb` file so that it adds no whitespace.


jonrohan (approved):
thanks!

PR #1108, opened 2022-04-09
Introduce `DeprecatedComponents` rubocop rule to automatically lint against deprecated component

khiga8:
## What

This introduces a rubocop cop called `DeprecatedComponents` to notify consumers when they are using a deprecated component.

Similar to [DeprecatedInPrimer](https://github.com/github/github/blob/master/.erb-linters/deprecated_in_primer_counter.rb) (Staff-only link), we can make use of the `statuses.json` file to dynamically determine which components have been deprecated. 

We can keep a mapping of alternative components to render, if there are any to customize the message. Even when there are no alternative components, the value should be set. That allows us to have a validation that reminds people to update the map.

## Why

I noticed that there were deprecated components being newly used in dotcom which we DON'T want.
Currently we have a few individual deprecated component linters like `lib/rubocop/cop/primer/deprecated_layout_component.rb`. Having individual linters depend on us remembering to make a linter whenever we deprecated a component which can be easy to forget. Since we have `deprecated` info available to us, let's use it!

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -12,3 +12,7 @@ linters:
     rubocop_config:
       inherit_from:
         - lib/rubocop/config/default.yml
+      Primer/DeprecatedComponents:
+        Enabled: true
+        Exclude:
+          - app/components/primer/layout_component.html.erb
--- a/lib/rubocop/cop/primer/deprecated_components.rb
+++ b/lib/rubocop/cop/primer/deprecated_components.rb
@@ -7,7 +7,7 @@ require "parser/current"
 module RuboCop
   module Cop
     module Primer
-      # This cop ensures that components marked as "deprecated" in `static/statuses.json` are not used.
+      # This cop ensures that components marked as "deprecated" in `static/statuses.json` are discouraged from use.
       #
       # bad
       # Primer::BlankslateComponent.new(:foo)
@@ -21,8 +21,8 @@ module RuboCop
       # good
       # Primer::Alpha::Tooltip.new(:foo)
       class DeprecatedComponents < BaseCop
-        # If there is no alternative, set the value to nil.
-        ALTERNATIVE_COMPONENTS = {
+        # If there is no alternative to suggest, set the value to nil.
+        COMPONENT_TO_USE_INSTEAD = {
           "Primer::BlankslateComponent" => "Primer::Beta::Blankslate",
           "Primer::DropdownMenuComponent" => nil,
           "Primer::Tooltip" => "Primer::Alpha::Tooltip",
@@ -34,8 +34,8 @@ module RuboCop
           return unless node.source.include?("Primer::")
 
           deprecated_components.each do |component|
-            pattern = "(send #{pattern(component)} :new ...)"
-            add_offense(node, message: message(component)) if NodePattern.new(pattern).match(node)
+            pattern = NodePattern.new("(send #{pattern(component)} :new ...)")
+            add_offense(node, message: message(component)) if pattern.match(node)
           end
         end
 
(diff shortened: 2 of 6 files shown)


lib/rubocop/cop/primer/deprecated_components.rb
```diff
@@ -0,0 +1,47 @@
...
+        ALTERNATIVE_COMPONENTS = {
+          "Primer::LayoutComponent" => "Primer::Alpha::Layout",
+          "Primer::Tooltip" => "Primer::Alpha::Tooltip",
+          "Primer::BlankslateComponent" => "Primer::Beta::Blankslate"
+        }
+
+        def on_send(node)
+          load_deprecated_components
+          # Todo - find match with `@deprecated_components``
+          # save match to `component` variable
```
khiga8, line 22: todo: read rubocop docs

https://docs.rubocop.org/rubocop-ast/node_pattern.html

lib/rubocop/cop/primer/deprecated_components.rb
```diff
@@ -0,0 +1,77 @@
...
+          json = JSON.parse(
+            File.read(
+              File.join(File.dirname(__FILE__), "../../../../static/statuses.json")
+            )
+          ).freeze
+          deprecated_components = json.select { |_, value| value == "deprecated" }.keys
+          deprecated_components.each do |deprecated|
+            unless COMPONENT_TO_USE_INSTEAD.key?(deprecated)
+              raise "Please provide a component that should be used in place of #{deprecated} in COMPONENT_TO_USE_INSTEAD. "\
+                    "If there is no alternative, set the value to nil."
```
khiga8, line 69: The intent behind this check is so that whenever someone updates a component to `deprecated`, they don't forget to provide an alternative component to suggest to the consumer. 

The `deprecated_components_test.rb` should fail with this message whenever someone updates component to deprecated status.

.erb-lint.yml
```diff
@@ -12,3 +12,7 @@ linters:
     rubocop_config:
       inherit_from:
         - lib/rubocop/config/default.yml
+      Primer/DeprecatedComponents:
+        Enabled: true
+        Exclude:
+          - app/components/primer/layout_component.html.erb
```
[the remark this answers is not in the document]
khiga8: Looks like we'll need to introduce a custom config. I'd rather do this in a separate PR so let's go with this for now.

khiga8:
This is ready for review 🎉 @primer/rails-reviewers

lib/rubocop/cop/primer/deprecated_components.rb
```diff
@@ -0,0 +1,80 @@
...
+        def deprecated_components
+          deprecated_components = statuses_json.select { |_, value| value == "deprecated" }.keys
+          deprecated_components.each do |deprecated|
+            unless COMPONENT_TO_USE_INSTEAD.key?(deprecated)
+              raise "Please provide a component that should be used in place of #{deprecated} in COMPONENT_TO_USE_INSTEAD. "\
+                    "If there is no alternative, set the value to nil."
+            end
+          end
+          deprecated_components
+        end
```
camertron, line 76: What would you think about memoizing the return value of this method so we don't re-compute the list of deprecated components on every call?

```suggestion
        def deprecated_components
          @deprecated_components ||= begin
            statuses_json.select { |_, value| value == "deprecated" }.keys.tap do |deprecated_components|
              deprecated_components.each do |deprecated|
                unless COMPONENT_TO_USE_INSTEAD.key?(deprecated)
                  raise "Please provide a component that should be used in place of #{deprecated} in COMPONENT_TO_USE_INSTEAD. "\
                        "If there is no alternative, set the value to nil."
                end
              end
            end
          end
        end
```

camertron (approved):
Just a small comment re: performance 😄

khiga8:
Ahh @camertron I'm so sorry I had auto-merge enabled 😅. I'll add your suggest changes in a new PR.

camertron:
@khiga8 woops, I guess I shouldn't have approved it! Ok no problem :)

PR #1111, opened 2022-04-11
Contributor documentation - Experimental component lifecycle

pouretrebelle:
An approach for developing new Primer ViewComponents that are validated in a production environment.

https://github.com/primer/view_components/blob/pouretrebelle/candidate-lifecycle-contributor-docs/contributor-docs/adding-components.md

We've gone back and forth on whether it's better to build new components in .com and then upstream them or build new components directly in PVC and then potentially have to iterate them after release because they haven't been validated in .com. Currently we're building in .com and then upstreaming, but there's a lot of admin to upstreaming (docs, api consistency, testing) which could be streamlined with this approach - the only downside being a fair bit of copy/pasting.

This proposal doesn't touch how to develop with CSS changes because [that's its own minefield](https://github.com/primer/view_components/blob/main/contributor-docs/styling.md), but I'm interested to hear whether CSS co-development makes .com a more or less attractive dev environment 🤔 

Feedback welcome, especially from the Accessibility team (cc: @owenniblock, @khiga8, @inkblotty, @bolonio)

(Also sorry for all my passive voice uses, feel free to suggest rewordings as well as proposal feedback)


contributor-docs/adding-components.md
```diff
@@ -14,6 +14,22 @@ We recommend building new Primer ViewComponents in an application that uses them
 
 We classify our components into alpha, beta and stable statuses. You can read more about these milestones in the [Primer’s component lifecycle](https://primer.style/contribute/component-lifecycle). In addition to the criteria listed there, to promote a Primer ViewComponent from alpha to beta it must not use any deprecated [view_component](https://viewcomponent.org/CHANGELOG.html) framework features, such as Slots v1.
 
+### Experimental lifecycle
+
+All components added to Primer ViewComponents must already be validated in a production environment. We recommend building components in an application that uses them first, but to speed up the upstreaming process it may be easier to use a hybrid approach:
+
+1. Develop the component in a Primer ViewComponents draft pull request
+
+    This sandbox is optimised for component development so it may be faster than your production application for initial explorations. You can write tests and documentation as you develop and it’s easier to achieve API consistency when you can see it next to other Primer ViewComponents.
```
pouretrebelle, line 23: As @bolonio [mentioned on Slack](https://github.slack.com/archives/CSGAVNZ19/p1649855445595769) it would be useful to have the docs visible while it's still an experimental component although I can't see how that would be possible without committing the draft to the repo 🤔

I guess if you're writing YARD documentation for the component you'd copy over that context in code comments to the production environment, would that be adequate?

owenniblock: Ideally, we'd have somewhere to put rich, detailed documentation. The work @khiga8 and @ichelsea did on the Tooltip is a great example of the standard we should hold ourselves to:

- https://primer.style/design/accessibility/tooltip-alternatives
- https://primer.style/view-components/components/alpha/tooltip

I'm not sure how we could include these in the experimental dotcom phase - maybe alongside the .rb files and then copy them over as part of the upstreaming process?

pouretrebelle: Another thought - you'd only have the experimental component in .com while the PVC component was in draft, so you could add a comment to the .com component linking to the docs in the draft PR's preview app. A bit messy to link to a temporary app in merged code but the Vercel preview apps are retained even after a PR is merged so the link would be live long enough for you to swap it out with the real PVC component.
(I hope that makes sense, words are hard today)

inkblotty (approved):
Thank you for this. I think we can add detail as the process continues to evolve, but this is exactly what we need to point people to now 🚀

khiga8 (approved):
I love that we're codifying this process through this doc. Thank you for this @pouretrebelle!

contributor-docs/adding-components.md
owenniblock: That's a good point! @bolonio what do you think of this approach? Would mean we'd link the experimental linter help text for the dialog to a temporary Vercel instance. Seems like a good compromise to me thanks @pouretrebelle

pouretrebelle: (also N.B. we'll be moving off Vercel soon in favour of GitHub Pages, which somehow makes these temp links feel less dirty to me)

bolonio: Sounds good, I'm already writing the YARD documentation in the draft PR :) Thanks @pouretrebelle for the help

bolonio (approved):

contributor-docs/adding-components.md
owenniblock: We hit an issue with this (which is totally my fault!) because we're making changes to `primer/behaviors` - I was having to develop the PVC code locally side-by-side with the behaviors work. This makes it hard to re-open the PR I closed.

What I think we've settled on is that we'll add the YARD mark-up to the version of `dialog.rb` in dotcom - we'll link to that file in our experimental component's linter. Then when the upstreaming happens, the docs will naturally move to Vercel (or GitHub Pages :))

Hope that's OK @pouretrebelle - let me know if you have other ideas which you think would be better.

camertron (approved):

joelhawksley (approved):
Thanks for writing this down ❤️

PR #1112, opened 2022-04-11
Add performance improvement to `DeprecatedComponents` linter

khiga8:
@camertron had a performance improvement suggestion in https://github.com/primer/view_components/pull/1108 but I had auto-merge set.

This adds the review suggestion which is to use memoization in the `deprecated_components` method for performance improvements.


camertron (approved):
Thanks! 🎉

PR #1061, opened 2022-03-16
Bug: Ensure tooltip takes up full width instead of parent width

khiga8:
cc: @hectahertz 

Partially addresses https://github.com/github/primer/issues/771

This PR: 

- adds `width: max-content;` to ensure that when the tooltip and anchor element have a parent container with `position: relative;`, the tooltip width doesn't take up that parents width. This also adds an example so there's no regression in future.
- adjusts tooltip arrow position to ensure all directions are similarly positioned.

#### Before tooltip width

Screenshot of tooltip that has inherited width from its span parent, causing it to be very narrow

#### After tooltip width

Screenshot of tooltip with appropriate width

#### Before tooltip arrow position

https://user-images.githubusercontent.com/16447748/158689757-bf79eda3-2f78-40a5-8eda-fa9072140ae9.mov

In this video, I hover over buttons to show a tooltip. The `se` and `ne` tooltip arrow have a larger space from the edge than the `sw` and `nw` tooltip arrow.

#### After tooltip arrow position

https://user-images.githubusercontent.com/16447748/158689794-485a09b3-ec07-47d1-be84-ae804450471a.mov

In this video, I hover over buttons to show a tooltip. The `se`, `ne`, `nw`, and `sw` tooltip arrow have same amount of space from edge.

PR #1114, opened 2022-04-12
bug: fix tooltip position offset issue for south and north

khiga8:
## What
This fixes a bug where the tooltip arrow is slightly offset in the north and south direction. See [Tooltip direction example](https://primer.style/view-components/components/alpha/tooltip#with-direction). Notice how the tooltip arrows for north and south directly are slightly off-centered.

Screen Shot of South facing tooltip with an arrow that isn't totally centered

## Why
I noticed this in the docs. I thought it was weird because we don't see this bug in dotcom. This turns out to be because dotcom PVC package hasn't been updated to `0.0.71` which includes https://github.com/primer/view_components/pull/1061. While that PR fixes a bug, it also introduced an incorrect offset of 10 instead of 6 for tooltips in the north and south directions. We only want the 10 offset for tooltips that are in `nw`, `ne`, `se`, `sw`. 

This cleans up the CSS a bit and fixes the bug. 

## Note to reviewer

I generated this [Vercel preview pointing to these local changes](https://view-components-ds7hz3dro-primer.vercel.app/view-components/components/alpha/tooltip#with-direction) with a temp commit. You can see arrows look good for all directions.


khiga8:
[Here is the vercel preview with this change](https://view-components-git-refs-headskh-fix-tooltip-position-primer.vercel.app/view-components/components/alpha/tooltip). (I generated this vercel using a temp commit to point to our JS changes).

The tooltip arrow is looking good for all directions 👍

koddsson (approved):

PR #1116, opened 2022-04-12
Add a ADR to document decision to use Catalyst

koddsson:
I think we should start using Catalyst in Primer ViewComponents so that developers don't need to remove it before making an upstream change to PVC.


keithamus (approved):
LGTM :laughing:

adr/0251-use-catalyst-for-client-side-behaviours.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Context
+
+<!-- The issue motivating this decision, and any context that influences or constrains the decision. -->
+
+Existing Primer ViewComponents behaviours don't use [Catalyst](https://github.github.io/catalyst) or any other dependency, while GitHub developers often start writing a component using Catalyst.
+
+Removing the Catalyst dependency from the Web Component is an extra burden on Web Developers with few benefits. The library is ~1kb of JavaScript, and github.com already includes the library as part of its client-side bundle. Adding the library to Primer ViewComponents does not affect github.com and has a minimal effect on other applications using Primer ViewComponents.
+
+## Decision
```
BlakeWilliams, line 17: Most ADR's I've seen usually include alternatives considered listing out pros and cons of each approach. Thoughts on adding something similar here?

koddsson: Happy to add any thing here but didn't see anything like that in other ADRs for this repo or in the linked reading. I feel like anything I'd come up with would be arbitrary straw-man arguments.

Maybe I could add not using Catalyst and using other similar libraries, would that work here?

BlakeWilliams: We've done the pro/con approach in a good number of ADR's, like [Global Notices](https://github.com/github/fast/blob/main/adr/0000-storage-backing-global-notice-next.md) and [snek](https://github.com/github/snek/blob/main/docs/adrs/04-ownership-reporting-architecture.md). Memex does [something similar](https://github.com/github/memex/blob/main/docs/adr/adr-022-api-convention-in-the-frontend.md#alternatives-explored) via `Alternatives Explored`.

> Maybe I could add not using Catalyst and using other similar libraries, would that work here?

I think that would be perfect 👍. 

I view ADR's mostly as tools to help drive folks towards consensus to one solution over other solutions as well as being an artifact to give readers context on why a specific approach was chosen compared to alternatives. If y'all are using them differently here that's totally fine, and feel free to consider my comments non-blocking. 🙂

adr/0251-use-catalyst-for-client-side-behaviours.md
```diff
@@ -0,0 +1,28 @@
...
+
+Accepted
+
+## Context
+
+<!-- The issue motivating this decision, and any context that influences or constrains the decision. -->
+
+Existing Primer ViewComponents behaviours don't use [Catalyst](https://github.github.io/catalyst) or any other dependency, while GitHub developers often start writing a component using Catalyst.
+
+Removing the Catalyst dependency from the Web Component is an extra burden on Web Developers with few benefits. The library is ~1kb of JavaScript, and github.com already includes the library as part of its client-side bundle. Adding the library to Primer ViewComponents does not affect github.com and has a minimal effect on other applications using Primer ViewComponents.
```
camertron, line 15:
```suggestion
Removing the Catalyst dependency from the Web Component is an extra burden on Web Developers with few benefits. The library is ~1kb of JavaScript, and github.com already includes the library as part of its client-side bundle. Adding the library to primer_view_components does not affect github.com and has a minimal effect on other applications using primer_view_components.
```

camertron (approved):
LGTM, thanks for writing this up :)

adr/0251-use-catalyst-for-client-side-behaviours.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Decision
+
+<!-- The change that we're proposing or have agreed to implement. -->
+
+Start using Catalyst in Primer ViewComponent behaviours where applicable.
```
khiga8, line 21: Would you recommend creating new custom elements for Primer use using catalyst? Is there a preference?

adr/0251-use-catalyst-for-client-side-behaviours.md
koddsson: Is `primer_view_components` how we prefer to refer to the project?

adr/0251-use-catalyst-for-client-side-behaviours.md
koddsson: I think we should yeah. Should I clarify that somehow in this doc?

khiga8: if there's a preference for catalyst, that would be helpful to make explicit! @lindseywild and I are working on an `ActionMenu` component and a `action-menu` element. we're in the early stages and we're not using catalyst. maybe some point later in the process we should consider changing it to catalyst.. or just start using catalyst for any new element that's introduced

adr/0251-use-catalyst-for-client-side-behaviours.md
camertron: That's technically the name of the gem. Feel free to ignore this though.

joelhawksley: IMO we should use Primer ViewComponents or PVC. I don't think we refer to the project by the gem name anywhere in writing.

adr/0251-use-catalyst-for-client-side-behaviours.md
```diff
@@ -0,0 +1,43 @@
...
+  - Cons:
+    - Migrating components from the github.com codebase would need to be converted to Web Components without Catalyst.
+    - PVC JavaScript behaviours would differ from how GitHub engineers write JavaScript behaviours today.
+
+## Consequences
+
+<!-- What becomes easier or more difficult to do and any risks introduced by the change that will need to be mitigated.-->
+
+- The Primer ViewComponents JS bundle gets larger.
+- Developers unfamiliar with Catalyst need to get familiar with the library.
```
lesliecdubs, line 43: Not blocking feedback, but what's our plan for providing documentation for these developers? @joelhawksley @camertron

camertron: @lesliecdubs Catalyst has really good docs at https://github.github.io/catalyst/

lesliecdubs: Nice! I'll take some time to get more familiar with it 👀 

We may want to consider a good way to cross-link this from the Primer docs to help explain "how to create/work on a PVC component that uses Catalyst."

PR #1119, opened 2022-04-13
styles: refactor to use ActionList css

khiga8:
Very first pass at swapping to use [ActionList](https://primer.style/css/components/action-list) styles

cc: @langermank, @lindseywild


langermank (approved):
This is a great start! I left feedback for future iterations 😄

PR #1121, opened 2022-04-13
Put `menuitem` on `li` instead.

khiga8:
This always adds the `menuitem` on the `li` instead of the nested interactive element. There is no role set on the nested element.


khiga8:
@lindseywild mind some 👀 before I merge this in? I believe this is what we discussed today

lindseywild:
Yep, looks good thank you!

PR #1131, opened 2022-04-22
Hooking up `getAnchoredPosition` to position menu

khiga8:
cc: @langermank, @lindseywild


app/components/primer/alpha/action-menu-element.ts
```diff
@@ -143,11 +176,27 @@ class ActionMenuElement extends HTMLElement {
...
   openPopup() {
     if (!this.#buttonEl || !this.#menuEl) return
 
     this.#buttonEl.setAttribute('aria-expanded', 'true')
     this.#menuEl?.removeAttribute('hidden')
+    this.#menuEl.classList.add('opacity-none')
+
+    this.#updatePosition()
+
+    this.#menuEl.classList.remove('opacity-none')
```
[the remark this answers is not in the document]
khiga8: Ok, we can do that :)

app/components/primer/alpha/action-menu-element.ts
```diff
@@ -22,8 +45,31 @@ class ActionMenuElement extends HTMLElement {
     // eslint-disable-next-line no-invalid-this
```
khiga8, line 45: Note to selfs: we should try to fix these disables before web systems sees it 😆

app/components/primer/alpha/action-menu-element.ts
```diff
@@ -22,9 +23,43 @@ class ActionMenuElement extends HTMLElement {
...
+          position: absolute;
+          width: 160px;
+          background-color: var(--color-canvas-overlay);
+          z-index: 1000000;
+          border: $border-width $border-style var(--color-border-default);
+          border-radius: $border-radius;
+          box-shadow: var(--color-shadow-large);
+        }
+    `
+    document.querySelector('body')?.appendChild(style)
```
khiga8, line 61: @langermank, This is temporary CSS I added particularly for styling the `ul[role='menu']` to validate the positioning logic works. I'm assuming that there will be a Primer CSS we want here?

(Note: I originally had styles defined in Shadow DOM but had the worst time figuring out how to style nested elements after markup change. This is temporary anyways!)

app/components/primer/alpha/action-menu-element.ts
```diff
@@ -143,11 +177,27 @@ class ActionMenuElement extends HTMLElement {
...
   openPopup() {
     if (!this.#buttonEl || !this.#menuEl) return
 
     this.#buttonEl.setAttribute('aria-expanded', 'true')
     this.#menuEl?.removeAttribute('hidden')
+    this.#menuEl.style.visibility = 'hidden'
+
+    this.#updatePosition()
+
+    this.#menuEl.style.visibility = 'visible'
```
khiga8, line 200: I'm also assuming that this inline setting of visibility will be replaced by a CSS class that has animation on it! @langermank

khiga8:
@langermank, @lindseywild: 

I'm so excited by how close we are 😄. Just so I don't forget next week, I'm noting here what I think the most immediate To-Dos CSS wise are:

- Introducing a CSS class for styling the `ul[role='none']` bits. I I don't think we have this yet. We're mocking this temporarily with a temporary style tag as noted in https://github.com/primer/view_components/pull/1131#discussion_r856748915 but we definitely don't want this in production.
- Making sure that the above CSS class and the Overlay classes are available for production use.
- Verifying the menu is responsive and looks as we expect in various scenarios.

Less urgent in terms of MVP but still ideal to get done:
- Creating an Overlay component.
- Swapping the Overlay related parts of `ActionMenu` with the said Overlay component.

Issue #1140, opened 2022-05-03
Autocomplete: when `is_label_visible` is false, label still has bottom margin


maxbeizer:
### Steps to reproduce
Use a autocomplete with invisible label with another element at the same  level, i.e. 
```ruby
<%= render(Primer::Beta::AutoComplete.new(label_text: "Fruits", src: "/auto_complete", input_id: "fruits-input--non-visible-label", list_id: "fruits-popup--non-visible-label", is_label_visible: false)) %><%= render Primer::ButtonComponent.new(type: :submit, size: :small) do %>Search<% end %>
```

### Expected behavior
I would expect those two elements to have the same height

### Actual behavior

![autocomplete-label](https://user-images.githubusercontent.com/2006658/166562470-1b66aa94-f3a0-47f3-946b-8c192d99e3e6.png)

The label with sr-only text has a bottom margin of 6px

https://github.com/primer/view_components/blob/a177fb8f0a5a403a01610bbd507480f941ecfad0/app/components/primer/beta/auto_complete/auto_complete.html.erb#L2 still gets the stacked label class which has bottom margin

### System configuration
**Rails version**: current dotcom

**Ruby version**: current dotcom

**Gem version**:current dotcom

PR #1141, opened 2022-05-03
Only add label class if label is visible

maxbeizer:
Closes https://github.com/primer/view_components/issues/1140

This will only add the label classes if the label is visible. How's that look to y'all? 🙇


app/components/primer/beta/auto_complete/auto_complete.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= render Primer::BaseComponent.new(**@system_arguments) do %>
-  <label for="<%= @input_id %>" class="<%= @label_classes %>">
+  <label for="<%= @input_id %>" class="<%= @label_classes if @is_label_visible %>">
```
camertron, line 2: This should work just fine, but what about setting `@label_classes` to `nil` in the Ruby code?

camertron (approved):

app/components/primer/beta/auto_complete/auto_complete.html.erb
maxbeizer: So, I looked at that 😄 .

Currently label class is set in a ternary, so adding another conditional would be a bigger code change. If you think it's clearer that way I'm happy to update it 👍

camertron (approved):

maxbeizer:
Thanks @jonrohan!

camertron:
Sorry @maxbeizer, looks like @jonrohan got you sorted :)

PR #1118, opened 2022-04-12
Add `eslint-plugin-custom-elements`

khiga8:
## What

This PR adds [eslint-plugin-custom-elements](https://github.com/github/eslint-plugin-custom-elements) and updates our `npm run lint` script to only run eslint on `.ts` files because the `.js` files are auto-generated.

## Why

We've started housing some custom elements in PVC and should probably align with how we build custom elements at GitHub.

PR #1142, opened 2022-05-03
Move Shadow DOM attachment back into the constructor

khiga8:
In https://github.com/primer/view_components/pull/1118, we moved the Shadow DOM attachment code into the `connectedCallback` to address an `eslint-plugin-custom-elements` error which claims that there should be no constructor in custom elements. 

[No constructor rule documentation](https://github.com/github/eslint-plugin-custom-elements/blob/main/docs/rules/no-constructor.md)

We shipped tooltip update in dotcom via PVC bump and subsequently ran into several JS errors during deploy ([Sentry JS errors](https://sentry.io/organizations/github/releases/90732ec9c17c55ff358f042a17ad8a88a53d5426/?environment=canary&environment=production&project=1890375)):
```
Failed to execute 'attachShadow' on 'Element': Shadow root cannot be created on a host which already host
```

I came across this [Stack overflow thread](https://stackoverflow.com/a/56908658) which seems to explain the error we ran into:

> Do not call attachShadow in connectedCallback:
> Otherwise you will be attempting to create a new shadowRoot every time your component is attached to the body. Yes, your code may not do that, but always write your code assuming that someone will attach, remove and re-attach.

This moves the shadow DOM attachment back into the constructor.


app/components/primer/alpha/tool-tip-element.ts
```diff
@@ -175,14 +175,19 @@ class ToolTipElement extends HTMLElement {
     return this.ownerDocument.getElementById(this.htmlFor)
   }
 
-  connectedCallback() {
+  // eslint-disable-next-line custom-elements/no-constructor
```
khiga8, line 178: I've filed an issue to change this rule -- https://github.com/github/eslint-plugin-custom-elements/issues/48

camertron (approved):

keithamus:
You can instead guard against re-attaching by checking for existence of the shadowRoot:

```
connectedCallback() {
  if (!this.shadowRoot) this.attachShadow({ mode: 'open' })
}
```

khiga8:
@keithamus Sure, we can try that instead. I feel this conditional check can be easy to forget. Could we enforce this with an additional eslint rule? 

Also, might you have suggestions for reproducing `connectedCallback` being called twice? I want to repro this error and didn't run into it in my review lab testing.

khiga8:
Okay, I was able to verify the fix.
I manually trigger `connectedCallback()` on `main` and run into the errors:
Screen Shot of console errors after calling connectedCallback manually


On this branch I don't run into errors:
Screen Shot of no console errors after calling connectedCallback manually


I think we're good to go with this @jonrohan.

PR #1153, opened 2022-05-11
Add a linter for calls to `super` in templates

camertron:
A [recent issue](https://github.com/github/platform-ux/issues/825) was brought to our attention wherein slash commands were being double rendered in dotcom. We tracked the issue down to calling `super` inside a component's template. The code in question was emitting the result of the `super` call, eg:

```erb
<%= super %>
```

Removing the `=` indicator fixes the issue. To remove further confusion, [I wrote a helper method](https://github.com/github/view_component/pull/1353) called `render_parent` that functions as a call to `super` and returns `nil` (to avoid double rendering).

This PR adds a linter to automatically replace `super` calls with `render_parent`.

### DIFF
--- a/.changeset/popular-vans-sleep.md
+++ b/.changeset/popular-vans-sleep.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': minor
+---
+
+Add a linter for calls to super in templates
--- a/.changeset/popular-vans-sleep.md
+++ b/.changeset/popular-vans-sleep.md
@@ -1,5 +1,5 @@
 ---
-'@primer/view-components': minor
+'@primer/view-components': patch
 ---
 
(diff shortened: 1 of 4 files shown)


joelhawksley (approved):

.changeset/popular-vans-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/view-components': minor
```
jonrohan, line 2: For now we're doing all `patch` I would like to have a conversation thought when we should start versioning properly

```suggestion
'@primer/view-components': patch
```

camertron: Ah ok thanks!

jonrohan (approved):

PR #1164, opened 2022-05-23
Adding a stale PR/Issue workflow

jonrohan:
This PR introduces a stalebot workflow exactly like the one in primer/css https://github.com/primer/css/blob/main/.github/workflows/stale.yml


camertron (approved):
Awesome, thanks! 🎉

PR #1179, opened 2022-05-31
Add `role` to system args docs

maxbeizer:
Role was not included in the system args docs. @joelhawksley confirmed that it does indeed function as a system arg, so this adds it to the docs.

I figured a link to MDN would be better than enumerating all the roles. WDYT?


jonrohan (approved):
Looks great thanks!

maxbeizer:
Thanks @jonrohan !!

PR #1183, opened 2022-06-02
Remove trailing newlines from the output of LinkComponent

camertron:
The `LinkComponent` appends a trailing newline, which can appear as unwanted whitespace between the link and any text that follows it. The problem is that the link_component.erb file contains a trailing newline, and ERB faithfully reproduces it in the rendered output. This PR replaces the template with a `call` method.


jonrohan (approved):
Nice

PR #1194, opened 2022-06-10
Gitignore version manager files

camertron:
I propose we gitignore .ruby-version and .tool-versions, files used by Ruby version managers. Generally speaking, development is done in codespaces these days, which already have the correct Ruby version installed. Moreover, the existence of .ruby-version confuses the `bin/vendor-gem` script in dotcom.


langermank (approved):
Sounds good I think? 😄

PR #1200, opened 2022-06-29
Use Rubocop cops in .erb files

camertron:
I noticed we don't use any Rubocop cops for .erb files. We probably should?


jonrohan (approved):
I think there was some reason we did this, since erblint also runs rubocop lints, but I can't be certain. probably doesn't hurt to run it twice 🤷🏻

camertron:
@jonrohan I think it only runs rubocop linters if it's configured to do so, which I don't think it is currently?

PR #1143, opened 2022-05-03
Overlay component

langermank:
This is an abstracted version of the [Experimental Dialog component](https://github.com/github/github/blob/master/app/components/primer/experimental/dialog.rb) intended to be used like a private API by components like Dialog, ActionMenu, SelectPanel, HoverCard, ReactionMenu... you get the point.

More documentation to come ✨

PR #1214, opened 2022-07-22
Dialog Component

keithamus:
This is the initial implementation of the `Primer::Alpha::Dialog` component.

This takes some of the code we already have in `github/github` from `Primer::Experimental::Dialog`, however I've dropped some properties for it to avoid pulling in configurations which we may want to later remove - to avoid breaking changes. So for example this doesn't implement helpers around `<include-fragment>` (those helpers cause unnecessary coupling between the two components, and can be abstracted away into different helpers), nor does this support forms (we need to more carefully consider the interplay between dialogs and forms, especially with regard to how native `<dialog>` will play a role here).

This takes the `modal-dialog` web component in its entirety from `github/github`. While this is a _re-implementation_ of native `<dialog>`, and it probably makes sense to simply implement `<dialog>` instead, unfortunately we've noticed some areas where `<dialog>` isn't quite suitable: the support is only there in newer browsers, and some browsers have buggy implementations, and we need some time to look over the polyfill to make sure it handles what we want and the trade-offs are worth it. While we haven't had time to assess this yet, the `<modal-dialog>` component is ready to go, and I believe a later refactor from `<modal-dialog>` to `<dialog>` should be possible without any breaking changes.

Another area we should look into as a follow-up to this initial implementation is how we can leverage the overlay component (#1143) to simplify some areas of the dialog work.


app/components/primer/alpha/dialog.rb
```diff
@@ -0,0 +1,138 @@
...
+      def initialize(
+        title:,
+        subtitle: nil,
+        width: DEFAULT_WIDTH,
+        id: "dialog-#{(36**3 + rand(36**4)).to_s(36)}",
+        **system_arguments
+      )
+        @system_arguments = deny_tag_argument(**system_arguments)
+
+        @system_arguments[:tag] = "modal-dialog"
```
langermank, line 116: Do you mind adding back the header and footer divider params?

```
show_header_divider: true,
show_footer_divider: true,
```

app/components/primer/alpha/dialog.rb
```diff
@@ -0,0 +1,138 @@
...
+          WIDTH_MAPPINGS[fetch_or_fallback(WIDTH_OPTIONS, width, DEFAULT_WIDTH)],
+          "Overlay--height-auto",
+          "Overlay--motion-scaleFade",
+          system_arguments[:classes]
+        )
+
+        if subtitle.present?
+          @subtitle = subtitle
+          @subtitle_id = "#{id}-description"
+          @system_arguments[:aria].describedby ||= @description
```
langermank, line 131:
```suggestion
          @system_arguments[:aria].describedby ||= @subtitle
```

I _think_ this is meant to be `subtitle` now

app/components/primer/alpha/dialog.rb
```diff
@@ -0,0 +1,138 @@
...
+    #   - **Dialog unique id**: A dialog should be unique. Give a unique id
+    #   setting `:dialog_id`. If no `:dialog_id` is given, a default randomize
+    #   hex id is generated.
+    #
+    #   The combination of both `:title` and `:dialog_id` establishes an
+    #   `aria-labelledby` relationship between the title and the unique id of
+    #   the dialog.
+    class Dialog < Primer::Component
+      DEFAULT_WIDTH = :medium
+      WIDTH_MAPPINGS = {
```
langermank, line 22: I chatted with @vdepizzol more today and we'd like to keep `height` as a param for now. I think we might be able to create some guardrails with width/height combinations using some of the new helpers @krhkt has been working on if I remember correctly from his demo. So something we could add later when Dialog depends on Overlay, perhaps?

stories/primer/alpha/dialog_stories.rb
```diff
@@ -0,0 +1,20 @@
...
+      description ""
+      select(:tag, Primer::Alpha::Dialog::TAG_OPTIONS, Primer::Alpha::Dialog::DEFAULT_TAG)
+      href "#"
+    end
+
+    content do |c|
+
+    end
+  end
+end
```
jonrohan, line 20: Generally trying to stop writing stories and start writing previews `tests/previews/` you can run script/lookbook to see the dialog and test it

app/components/primer/alpha/modal-dialog-element.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+}
+
+declare global {
+  interface Window {
+    ModalDialogElement: typeof ModalDialogElement
+  }
+  interface HTMLElementTagNameMap {
+    'modal-dialog': ModalDialogElement
```
vdepizzol, line 142: _Sidenote_: I've been wondering whether we should normalize Primer custom elements with a `primer` prefix. In this case, we don't use the term "modal" to refer to dialogs except in this custom component due to the hyphen requirement. Another example is the [`tool-tip`](https://github.com/github/github/blob/master/node_modules/%40primer/view-components/app/components/primer/alpha/tool-tip-element.js) element.

Maybe `primer-dialog`, `primer-tooltip`, `primer-split-page-layout`, etc?

Related: ADR on [spelling of component names](https://github.com/github/primer/blob/main/adrs/2022-02-15-spelling-of-component-names.md)

app/components/primer/alpha/dialog.rb
```diff
@@ -0,0 +1,127 @@
...
+      #   @code
+      #     <%= render(Primer::Alpha::Dialog.new(
+      #       title: "Dialog Example",
+      #     )) do |d| %>
+      #       <% d.show_button { "Show Dialog" } %>
+      #       <% d.body do %>
+      #         <p>Some content</p>
+      #       <% end %>
+      #       <% d.footer do %>
+      #         <%= render(Primer::ButtonComponent.new(data: { "close-dialog-id": "my-dialog" })) { "Cancel" } %>
```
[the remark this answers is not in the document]
owenniblock: Yes, you'd need to send the id to the `Primer::Alpha::Dialog.new()` call to add `dialog_id: "my-dialog"`.

keithamus:
There's a little lint failure to fix here, but otherwise I think this is ready for review!

app/components/primer/alpha/dialog.rb
```diff
@@ -0,0 +1,142 @@
...
+      #         <%= render(Primer::ButtonComponent.new(scheme: :primary)) { "Submit" } %>
+      #       <% end %>
+      #     <% end %>
+      # @param id [String] The id of the dialog.
+      # @param title [String] The title of the dialog.
+      # @param subtitle [String] The subtitle of the dialog. This will also set the `aria-describedby` attribute.
+      # @param width [Symbol] The width of the dialog. <%= one_of(Primer::Alpha::Dialog::WIDTH_OPTIONS) %>
+      # @param height [Symbol] The height of the dialog. <%= one_of(Primer::Alpha::Dialog::HEIGHT_OPTIONS) %>
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+      def initialize(
```
langermank, line 101: I love your approach of pairing down some of the logic so we can address it with Overlay. However, I want to point out that [at least one team](https://github.com/github/github/pull/232953/files#diff-69396c583a76bad41c8de7b68c6a0c58afe813644b58e624eebb4544ab49aeefR28) has a use case for the `narrow` bottom sheet position, so I'm not sure how we should handle that here. I recommended they add just the position they need right now. Would we leave the Experimental version in place while Overlay gets going, and then ship another version of Dialog later with all the responsive properties?

app/components/primer/alpha/dialog/header.rb
```diff
@@ -0,0 +1,35 @@
...
+      # A `Dialog::Header` is a compositional component, used to render the
+      # Header of a dialog. See <%= link_to_component(Primer::Alpha::Dialog) %>.
+      class Header < Primer::Component
+        # @param title [String] The title of the dialog.
+        # @param subtitle [String] The subtitle of the dialog. This will also set the `aria-describedby` attribute.
+        # @param hide_divider [Boolean] If true the visual dividing line between the body and footer will be hidden
+        # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+        def initialize(
+          id:,
+          title:,
```
langermank, line 15: The `title` size option is another one that folks are using in dotcom (we have a default and a `large` title size). We were thinking it might make more sense to ship a `ConfirmationDialog` with that large title by default, and perhaps a bit more prescriptive about content? Thinking large title that must be visible, and two default buttons in the footer (danger and cancel). Regardless, should we add the title size back here in the meantime?

app/components/primer/alpha/dialog/header.rb
```diff
@@ -0,0 +1,35 @@
...
+          subtitle: nil,
+          subtitle_id: nil,
+          hide_divider: false,
```
langermank, line 18: Since this should be false by default, I wonder if we could rename it to `show_divider`. This prop name also aligns a bit better with how PRC has referenced "divider" props.

```suggestion
          show_divider: false,
```

app/components/primer/alpha/dialog/footer.rb
```diff
@@ -0,0 +1,31 @@
...
+module Primer
+  module Alpha
+    class Dialog
+      # A `Dialog::Footer` is a compositional component, used to render the
+      # Footer of a dialog. See <%= link_to_component(Primer::Alpha::Dialog) %>.
+      class Footer < Primer::Component
+        # @param hide_divider [Boolean] If true the visual dividing line between the body and footer will be hidden
+        # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+        def initialize(
+          hide_divider: false,
```
langermank, line 12: Same thought here as for header

langermank (commented):
A few thoughts on param names and some of the removed logic. I love how you broke out all the regions!

app/components/primer/alpha/dialog.rb
```diff
@@ -0,0 +1,142 @@
...
+        Primer::Alpha::Dialog::Header.new(
+          id: @id,
+          title: @title,
+          subtitle: @subtitle,
+          hide_divider: hide_divider,
+          **system_arguments
+        )
+      }
+
+      # Required body content.
```
owenniblock, line 67: Is this a required section any more given the introduction of the `content` in the `.erb` file?

app/components/primer/alpha/modal-dialog-element.ts
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        dialogId = button.getAttribute('data-submit-dialog-id')
+        if (dialogId === this.id) {
+          this.close(true)
+        }
+
+        dialogId = button.getAttribute('data-show-dialog-id')
+        if (dialogId === this.id) {
+          //TODO: see if I can remove this
```
owenniblock, line 93: This is my TODO - I believe that the `stopPropogation` is required here but I forgot to remove the comment!

app/components/primer/alpha/modal-dialog-element.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  disconnectedCallback(): void {
+    this.#abortController?.abort()
+  }
+
+  show() {
+    this.open = true
+  }
+
+  close(closed = false) {
```
owenniblock, line 113: I'm still not super happy with the wording I used here. There's nothing about the parameter which says "if it's not closed, it's cancelled". Maybe we could rename the variable to something like `closedNotCancelled` or something which better communicates the two states that result in changing this value.

app/components/primer/alpha/modal-dialog-element.ts
owenniblock: We also presumably need to use a different name because we'll be running the Experimental `modal-dialog` still until we've worked out how we're handling all of the bits of functionality that were added to that version. I like `primer-dialog`.

owenniblock (commented):
I love how you've split this in to different subcomponents! ❤️ The only reason I'm not giving it a 👍 is because I think we need to rename the WebComponent to avoid a clash with the experimental version in dotcom.

app/components/primer/alpha/dialog/header.rb
```diff
@@ -0,0 +1,35 @@
...
+      class Header < Primer::Component
+        # @param title [String] The title of the dialog.
+        # @param subtitle [String] The subtitle of the dialog. This will also set the `aria-describedby` attribute.
+        # @param hide_divider [Boolean] If true the visual dividing line between the body and footer will be hidden
+        # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+        def initialize(
+          id:,
+          title:,
+          subtitle: nil,
+          subtitle_id: nil,
```
jonrohan, line 17: Seems that `subtitle_id` isn't used anywhere?

app/components/primer/alpha/dialog.rb
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+        @id = id.to_s
+        @title = title
+
+        @subtitle = subtitle
+        if subtitle.present?
+          @system_arguments[:aria] ||= {}
+          @system_arguments[:aria][:describedby] ||= "#{@id}-description"
+        end
```
jonrohan, line 130:
```suggestion
        return if subtitle.present?
        @system_arguments[:aria] ||= {}
        @system_arguments[:aria][:describedby] ||= "#{@id}-description"
```

Linter wants guard clause

jonrohan (approved):
Ran this locally looking great 👍🏻 left a few code comments and I pushed lint fixes for you, hope that's ok.

app/components/primer/alpha/dialog.rb
keithamus: Yeah I'd be happy to do this :+1:

app/components/primer/alpha/dialog/header.rb
keithamus: > ConfirmationDialog

This was absolutely my intent as a fast-follow to Dialog.

app/components/primer/alpha/dialog.rb
keithamus: It is as the content is used for wrapping outside of the body content.

keithamus:
> We also presumably need to use a different name because we'll be running the Experimental modal-dialog still until we've worked out how we're handling all of the bits of functionality that were added to that version. I like primer-dialog.

@owenniblock  IIRC the code is almost identical so we can ship with this as a replacement.

andrialexandrou:
As a heads up @keithamus Owen is out for two weeks and will be back September 12

khiga8:
👋 I'm not sure familiar with the state of dialog is currently but wanted to bring this [audit comment from @jscholes ](https://github.com/github/accessibility-audits/issues/275#issuecomment-1228870632) to your attention in case it hasn't been surfaced yet.

langermank:
👋 Documenting some tips for rolling this out to dotcom (replacing `Experimental::Dialog`)

| Experimental param | Alpha param | note |
| -- | -- | -- |
| `footer_content_align` | deprecated | remove param and do visual check |
| `header_variant` | deprecated | remove param and do visual check |
| `motion` | deprecated | this has a default value now |
| `description` | `subtitle` | name change |



The `height` and `width` params are combined into a new `size` option. I'll try and map out how these swaps might look, but visual testing is likely required.

| width | new size |
| -- | -- |
| `:small`| `:small` |
| `:medium`| `:small` |
| `:large`| `:medium` |
| `:xlarge`| `:large` |
| `:xxlarge`| `:xlarge` |

The `size` param handles height and width. If there's a case where an `Experimental::Dialog` has a width of `:large` and a height set to anything above `:large`, set the `size` to `medium_portrait`.

For Dialogs wrapped in `<form>` tags, work with @keithamus on how to structure the Alpha Dialog.

PR #1215, opened 2022-07-22
Add `DeprecatedComponents` ERB lint rule

khiga8:
Related: https://github.com/github/accessibility/issues/1290

## Context

We previously added a rubocop rule against deprecated Primer components (https://github.com/primer/view_components/pull/1108) to flag when consumers are using deprecated components.

However, we haven't been able to turn this on for ERB files in dotcom due to limitations of disabling rubocop in ERB files which sucks because there's nothing that's stopping a consumer from using a deprecated primer component.

This PR adds an ERB lint counter equivalent for ERB.

## Updates

- added ERB rule
- pulled methods shared between Rubocop into a shared helpers file
- added tests


lib/primer/view_components/linters/deprecated_components_counter.rb
```diff
@@ -0,0 +1,80 @@
...
+          else
+            # add comment with counter if none
+            corrector.insert_before(processed_source.source_buffer.source_range, "#{offense.context}\n")
+          end
+        end
+      end
+
+      private
+
+      def counter_correct?(processed_source)
```
khiga8, line 51: I saw that `BaseLinter` already defines this. However, `BaseLinter` seems to be tied to a very particular linter implementation, and has a bunch of things that are irrelevant for the check we're trying to do here. Therefore, I am NOT using `BaseLinter` and re-defining the rule. 

I don't understand the `counter_correct?` in `BaseLinter` enough to know whether its refactorable.

camertron:
@khiga8 I'm not sure I understand 😅 Is this a re-implementation of the ERB counting mechanism available in dotcom?

Side note: I got around the ignore problem by implementing a [config option](https://github.com/primer/view_components/blob/main/lib/primer/view_components/linters/disallow_action_list.rb#L14).

khiga8:
> Is this a re-implementation of the ERB counting mechanism available in dotcom?

yes! there's a `counter_correct?` in `BaseLinter` but it has a bunch of extra things and I don't know what it's doing. 🤔 the `BaseLinter` seems to be specialized but maybe I'm mistaken.

khiga8:
side note, I installed this branch on a dotcom branch!

There are so many violations. here's a draft branch with the changes - [kh-deprecated-components](https://github.com/github/github/compare/kh-deprecated-components-rule?expand=1) (staff only)

lib/primer/view_components/linters/deprecated_components_counter.rb
khiga8: Update: I removed this definition.

Instead I added `require "erblint-github/linters/custom_helpers"` and pull in the method definition there so we aren't duplicating it.

jonrohan (approved):
Looks good 👍🏻

PR #1216, opened 2022-07-22
Move lookbook previews from the lookbook folder to the main project

jonrohan:
I think we should move the lookbook previews out of the `/lookbook/test/components/previews` folder and move them to the main project's `/test/components/previews/` folder.

The reason is, there's no need to keep them in `/lookbook/..` and we'll be using these previews for other testing in the future so having the location more general can be helpful.

cc @joelhawksley @langermank


langermank (approved):
I like it!

camertron (approved):
Yes! This is awesome.

PR #1203, opened 2022-06-30
Add erblint-github dependency

khiga8:
This PR adds [erblint-github](https://github.com/github/erblint-github) as a development dependency. This contains several accessibility-focused rules for ERB.

PR #1220, opened 2022-07-25
bump required ruby version to 2.6

khiga8:
## Context

This PR drops support for Ruby 2.5.xx because I want to add `erblint-github` for linting which requires Ruby >= 2.6.
I guess I could also update `erblint-github` to require `Ruby >= 2.5` but it seems better to support the latest versions of Ruby.

See https://github.com/primer/view_components/pull/1203#issuecomment-1171609491 for context.
I am pulling these changes to a separate PR from the lint one.

This PR also adds support for Ruby 3.1.x

## To reviewer
@primer/rails-reviewers are there concerns with dropping support for Ruby 2.5.xx?


.github/workflows/ci.yml
```diff
@@ -59,7 +59,7 @@ jobs:
     strategy:
       matrix:
         rails_version: [5.2.3, 6.1.1, main]
-        ruby_version: [2.5.x, 2.6.x, 2.7.x, 3.0.x]
+        ruby_version: [2.6.x, 2.7.x, 3.0.x]
```
khiga8, line 62: what do these builds do?

camertron: What do you mean?

camertron: Also, what would you think about adding 3.1.x?

camertron (approved):

.github/workflows/ci.yml
khiga8: what do these benchmark checks do? just curious

camertron: Ohhh I see. They don't really do anything as far as I know 😅 I think we have them in case we want to look back at the numbers to see when regressions were introduced. We should probably be comparing them to some baseline tho.

.github/workflows/ci.yml
```diff
@@ -59,22 +59,24 @@ jobs:
...
           - rails_version: 6.1.1
             ruby_version: 2.7.x
           - rails_version: 6.1.1
             ruby_version: 3.0.x
-          - rails_version: main
-            ruby_version: 2.5.x
+          - rails_version: 6.1.1
+            ruby_version: 3.1.x
           - rails_version: main
             ruby_version: 2.6.x
```
camertron, line 81: This can be bumped to Ruby 3.1.

.github/workflows/ci.yml
```diff
@@ -124,22 +126,24 @@ jobs:
...
```
camertron, line 148: Ruby 3.1 plz :)

test/benchmarks/bench_classify.rb
```diff
@@ -38,7 +38,7 @@ def setup
   end
 
   def bench_allocations
-    assert_allocations "3.0" => 6, "2.7" => 4, "2.6" => 4, "2.5" => 4 do
+    assert_allocations "3.0" => 6, "2.7" => 4, "2.6" => 4 do
```
camertron, line 41: I think you'll need to add an allocation count for Ruby 3.1 as well.

camertron (changes requested):

.github/workflows/ci.yml
khiga8: maybe I am tripping but can you check the latest changes?

khiga8: ohhh line 144 against `main`? is there a reason for this being behind so many?

test/benchmarks/bench_classify.rb
```diff
@@ -38,7 +38,7 @@ def setup
   end
 
   def bench_allocations
-    assert_allocations "3.0" => 6, "2.7" => 4, "2.6" => 4, "2.5" => 4 do
+    assert_allocations "3.1" => 6, "3.0" => 6, "2.7" => 4, "2.6" => 4 do
```
khiga8, line 41: I arbitrary set it to 6.

.github/workflows/ci.yml
khiga8: done in https://github.com/primer/view_components/pull/1220/commits/c06b8b175e7bee1d87d445d0c28128c689a7053e.

.github/workflows/ci.yml
khiga8: done in https://github.com/primer/view_components/pull/1220/commits/c06b8b175e7bee1d87d445d0c28128c689a7053e.

.github/workflows/ci.yml
camertron: Thanks! Yeah I think it was just an oversight.

test/benchmarks/bench_classify.rb
camertron: That's probably right. I would just set it to the number the test failure spits out, if any.

khiga8:
Ok new error with the 3.1.x build.

> LoadError: cannot load such file -- matrix

It turns out starting Ruby 3.1, the `matrix` gem is a dedicated gem. (https://github.com/gdelugre/origami/issues/81)
I just added `matrix` as a dev dependency. If this continues to fail builds, I think I'm going to save the 3.1 support for another day since this PR is blocking my other PRs.

khiga8:
idk why the 4 expected checks are waiting. it looks like the CI build is complete ._.

camertron (approved):
Oh yeah those two expected checks will need to be removed in settings since there's no ruby 2.5 build anymore.

khiga8:
@camertron I don't have access to PVC settings. mind doing this for me?

khiga8:
nvm thank you!!!

PR #1224, opened 2022-07-25
Add the ConditionalWrapper Component

camertron:
### The Problem

During the development of the forms framework last quarter I found myself needing a component that would only wrap content if a certain condition were `true`. I have a similar need this quarter as I start work on the ActionList effort.

It's possible to do this sort of thing today with a good 'ol `if` statement, but you have to duplicate a bunch of code in each branch:

```erb
<% if some_condition? %>
  <%= render(Primer::BaseComponent.new(tag: :div)) do %>
    <span class="foobar"> ... </span>
  <% end %>
<% else %>
  <span class="foobar"> ... </span>
<% end %>
```

It's also possible to capture the inner HTML and use it in both branches to reduce duplication a bit:

```erb
<% inner_html = capture do %>
   <span class="foobar"> ... </span>
<% end %>
<% if some_condition? %>
  <%= render(Primer::BaseComponent.new(tag: :div)) do %>
    <%= inner_html %>
  <% end %>
<% else %>
  <%= inner_html %>
<% end %>
```

Both of these approaches are pretty awkward, and I personally find them hard to read.

### Proposed Solution

What we really need IMHO is a component that renders a wrapper element if a particular condition is true and doesn't render the wrapper element if the condition is false. View Component gives us the `render?` method, but that's used to skip rendering an entire component, including its content.

I propose we introduce a new `ConditionalWrapper` component to encapsulate the desired behavior:

```erb
<%= render(ConditionalWrapper.new(condition: true, tag: :div)) %>
   <span class="foobar"> ... </span>
<% end %>
```

So clean. Much readable. Wow.

### Prior Art

The primer/react project has a [`ConditionalBox` component](https://github.com/primer/react/blob/959184386922cd09024ade05c3cf8e0356f1c7f1/src/ActionList/Item.tsx#L305) for their implementation of ActionList.

### DIFF
--- a/.changeset/tender-pugs-promise.md
+++ b/.changeset/tender-pugs-promise.md
@@ -2,4 +2,4 @@
 '@primer/view-components': patch
 ---
 
-Add the ConditionalBaseComponent for conditionally wrapping content
+Add the ConditionalWrapper for conditionally wrapping content
--- a/.changeset/tender-pugs-promise.md
+++ b/.changeset/tender-pugs-promise.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Add the ConditionalBaseComponent for conditionally wrapping content
--- a/app/components/primer/conditional_wrapper.rb
+++ b/app/components/primer/conditional_wrapper.rb
@@ -7,21 +7,21 @@ module Primer
   #
   # @example True conditional
   #   <%# condition is true, so content will be wrapped in a <span> tag
-  #   <%= render Primer::ConditionalBaseComponent.new(condition: true, tag: :span, class: "foobar")) do %>
+  #   <%= render Primer::ConditionalWrapper.new(condition: true, tag: :span, class: "foobar")) do %>
   #     <%# also rendered %>
   #     <p class="bazboo">Some text</p>
   #   <% end %>
   #
   # @example False conditional
   #   <%# condition is false so no <span> tag will render around the content (i.e. the <p> tag)
-  #   <%= render(Primer::ConditionalBaseComponent.new(condition: false, tag: :span, class: "foobar")) do %>
+  #   <%= render(Primer::ConditionalWrapper.new(condition: false, tag: :span, class: "foobar")) do %>
   #     <%# this content will be rendered %>
   #     <p class="bazboo">Some text</p>
   #   <% end %>
   #
   # @param condition [Boolean] Whether or not to wrap the content in a `Primer::BaseComponent`.
   # @param base_component_arguments [Hash] The arguments to pass to `Primer::BaseComponent`.
-  class ConditionalBaseComponent < Primer::Component
+  class ConditionalWrapper < Primer::Component
     def initialize(condition:, **base_component_arguments)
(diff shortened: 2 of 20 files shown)


khiga8:
> So clean. Much readable. Wow.

I agree with this! My only concern is potential for misinterpretation of what the component does.

Before I read the PR description, I interpreted `ConditionalBaseComponent` to mean that the entire component only renders if the condition is met. Someone who sees this in the wild could interpret this similarly.

I wonder if updating the name to something like `ConditionalWrapper` would help... I don't have a very strong preference either way.

Side note: I think more recent components don't have `-Component` at the end so I'd suggest removing that.

camertron:
@khiga8 those are great points, thank you 😄 I used the "-Component" suffix to match the naming of `BaseComponent`, but I agree the name is confusing. If we changed it to `ConditionalWrapper` then the need for consistency with `BaseComponent` goes away.

.changeset/tender-pugs-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Add the ConditionalBaseComponent for conditionally wrapping content
```
joelhawksley, line 5:
```suggestion
Add the ConditionalWrapper for conditionally wrapping content
```

app/components/primer/conditional_wrapper.rb
```diff
@@ -0,0 +1,36 @@
...
+  # @param condition [Boolean] Whether or not to wrap the content in a `Primer::BaseComponent`.
+  # @param base_component_arguments [Hash] The arguments to pass to `Primer::BaseComponent`.
+  class ConditionalWrapper < Primer::Component
+    def initialize(condition:, **base_component_arguments)
+      @condition = condition
+      @base_component_arguments = base_component_arguments
+    end
+
+    def call
+      return content unless @condition
```
BlakeWilliams, line 31: Just mentioning since it seemed relevant, but it's likely that we'll end up evaluating `content` even if `@condition` is `false` if https://github.com/github/view_component/pull/1428 comes back (which I believe it will).

If we wanted to get ahead of that change, we could call `content` before the `return content` call above.

BlakeWilliams: Maybe ignore me, I was thinking this may have been similar to the Avo issue where there was a conditional component that would render `nothing`, relying on `content` never being evaluated. It looks like we always expect `content` to be used here. 🤦

camertron: Right yeah, we always expect `content` to be evaluated/emitted.

joelhawksley (approved):

lesliecdubs:
👋🏻 This is a nice addition! @camertron do you think we need to write some light public docs for this one, or do we only want this used sparingly/internally?

camertron:
@lesliecdubs I envisioned that it would only be used internally :)

lesliecdubs:
Thanks, copy that @camertron! We were putting together the Primer changelog today and I wasn't sure whether to include or not. I'll remove.

PR #1225, opened 2022-07-25
SegmentedControl component

jonrohan:
closes https://github.com/github/primer/issues/760

👋🏻 This is ready for review http://primer-view-components-preview-1225.eastus.azurecontainer.io/view-components/lookbook/inspect/primer/alpha/segmented_control/default

## SegmentedControl

169250821-b90eaaf2-905f-4ae5-bae3-bf277dacba30

This PR adds a SegmentedControl component with status `:alpha` to the library. 

### Arguments

`SegmentedControl`

| Name | Type | Default | Description |
| :- | :- | :- | :- |
| `hide_labels`  | `Boolean` | `false` | Whether to hide the labels and only show the icons |
| `full_width` | `Boolean` | `false` | If the component should be full width |

`SegmentedControl::Item`

| Name | Type | Default | Description |
| :- | :- | :- | :- |
| `label` | `String` | `nil` | Label for the Button or IconButton tooltip |
| `icon` | `Symbol` | `nil` | Symbol for octicon |
| `selected` | `Boolean` | `false` | Setting the selected state of the button |
| `hide_labels`  | `Boolean` | `false` | Whether to hide the labels and only show the icons |

### Interaction 

For [Design Guidelines](https://primer.style/design/components/segmented-control)


app/components/primer/alpha/segmented-control-element.ts
```diff
@@ -0,0 +1,36 @@
...
+      if (!(selectedControl instanceof HTMLElement) || !selectedControl.closest('ul')) return
+
+      for (const control of controls) {
+        control.setAttribute('aria-current', 'false')
+      }
+
+      selectedControl.setAttribute('aria-current', 'true')
+      selectedControl.focus()
+    })
+  }
```
jonrohan, line 24: @keithamus This seems super generic, Is there an existing behavior or web-component that takes care of this `click -> selected state` action described here? https://primer.style/design/components/segmented-control#keyboard-navigation

app/components/primer/alpha/segmented_control.rb
```diff
@@ -0,0 +1,78 @@
...
+module Primer
+  module Alpha
+    # Add a general description of component here
+    # Add additional usage considerations or best practices that may aid the user to use the component correctly.
+    # @accessibility Add any accessibility considerations
+    class SegmentedControl < Primer::Component
+      status :alpha
+
+      ICON_ONLY_DEFAULT = false
+      ICON_ONLY_OPTIONS = [ICON_ONLY_DEFAULT, true, :whenNarrow].freeze
```
jonrohan, line 12: This is a bit of a strange compromise that I did until we make updates to our responsive guidelines for `when_narrow: { icon_only: true }` I'm willing to change how this is if someone has concerns.

mperrotti: I don't think `icon_only` should be on the parent - I'd recommend either:
- A different child component (this is what we do in React: https://primer.style/react/SegmentedControl#with-icons-only)
- A prop on the child Button component

I recommend doing a different child component because icon-only buttons will show a tooltip, so the functionality is a little different.

mperrotti: Are we able to support both responsive variants in PVC?
1. icon-only
2. swap with an Dropdown

mperrotti:
Overall this is looking good, but I'd like us to align a little closer closer on the API if possible. Here's the React API for reference: https://primer-e83bf66799-13348165.drafts.github.io/drafts/SegmentedControl#props (ignore the `loading` prop - we decided against that)

app/components/primer/alpha/segmented-control-element.ts
keithamus: Might be a good opportunity to bring in Catalyst which would save on boilerplate, and we could work on extracting that kind of behaviour.

app/components/primer/alpha/segmented-control-element.ts
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable custom-elements/no-constructor */
+export default class SegmentedControlElement extends HTMLElement {
```
camertron, line 2: Why not use a Catalyst controller for this?

app/components/primer/alpha/segmented-control-element.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+
+declare global {
+  interface Window {
+    SegmentedControlElement: typeof SegmentedControlElement
+  }
+}
+
+if (!window.customElements.get('segmented-control')) {
```
camertron, line 33: If we use Catalyst, this block can be removed.

app/components/primer/alpha/segmented_control.html.erb
```diff
@@ -0,0 +1,9 @@
+<segmented-control>
+  <%= render Primer::BaseComponent.new(**@system_arguments) do %>
```
camertron, line 2: Since the tag can't be customized, consider passing it directly instead of setting it in `system_arguments`.

```suggestion
  <%= render Primer::BaseComponent.new(tag: :ul, **@system_arguments) do %>
```

app/components/primer/alpha/segmented_control.rb
```diff
@@ -0,0 +1,59 @@
...
+      #     <%= c.with_button(icon: :eye, selected: true) { "Preview" } %>
+      #     <%= c.with_button(icon: :"file-code") { "Raw" } %>
+      #     <%= c.with_button(icon: :people) { "Blame" } %>
+      #   <% end %>
+      #
+      # @param full_width [Boolean] If the component should be full width
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+      def initialize(full_width: FULL_WIDTH_DEFAULT, **system_arguments)
+        @system_arguments = system_arguments
+        @system_arguments[:tag] = "ul"
```
camertron, line 51: Consider passing `tag: :ul` in the template as suggested above.

```suggestion
```

app/components/primer/alpha/segmented_control/button.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render Primer::Beta::BaseButton.new(**@system_arguments) do %>
```
camertron, line 1: Consider:

```suggestion
<%= render Primer::Beta::BaseButton.new(tag: :button, **@system_arguments) do %>
```

app/components/primer/alpha/segmented_control/button.rb
```diff
@@ -0,0 +1,53 @@
...
+        ICON_ONLY_OPTIONS = ICON_ONLY_MAPPINGS.keys
+
+        def initialize(icon: nil, icon_only: ICON_ONLY_DEFAULT, selected: false, **system_arguments)
+          @system_arguments = system_arguments
+
+          # Icons
+          @icon = icon
+          @icon_only = fetch_or_fallback(ICON_ONLY_OPTIONS, icon_only, ICON_ONLY_DEFAULT)
+
+          @system_arguments[:tag] = :button
```
camertron, line 25:
```suggestion
```

app/components/primer/alpha/segmented_control/button.html.erb
```diff
@@ -0,0 +1,9 @@
...
+  <div class="SegmentedControl-content">
+    <%= render Primer::OcticonComponent.new(@icon, classes: "SegmentedControl-leadingVisual") unless @icon.nil? %>
+    <% unless @icon_only == true %>
+      <span class="SegmentedControl-text" data-content="<%= trimmed_content %>"><%= trimmed_content %></span>
+    <% end %>
+  </div>
+<% end %>
+<%= render Primer::Alpha::Tooltip.new(direction: :n, hide: @icon_only == :when_narrow ? :whenRegular : nil, type: :description, for_id: @system_arguments[:id], text: trimmed_content) if tooltip? %>
```
camertron, line 9: Why is `:whenRegular` camel-cased?

app/components/primer/alpha/segmented_control/button.rb
```diff
@@ -0,0 +1,53 @@
...
+      # `SegmentedControl::Button` is used inside `SegmentedControl` to render a button.
+      class Button < Primer::Component
+        status :alpha
+
+        ICON_ONLY_DEFAULT = false
+        ICON_ONLY_MAPPINGS = {
+          ICON_ONLY_DEFAULT => "",
+          true => "SegmentedControl-button--iconOnly",
+          :when_narrow => "SegmentedControl-button--iconOnly-whenNarrow",
+        }.freeze
```
camertron: It's kinda weird to me to use both `true`/`false` and a symbol for keys here. What about something like

```suggestion
        ICON_ONLY_DEFAULT = :never
        ICON_ONLY_MAPPINGS = {
          ICON_ONLY_DEFAULT => "",
          :always => "SegmentedControl-button--iconOnly",
          :when_narrow => "SegmentedControl-button--iconOnly-whenNarrow",
        }.freeze
```

test/previews/primer/alpha/segmented_control_preview.rb
```diff
@@ -0,0 +1,83 @@
...
+        render(Primer::Alpha::SegmentedControl.new(full_width: full_width)) do |c|
+          Array.new(number_of_buttons || 3) do |i|
+            c.with_button(icon: :zap, selected: i.zero?, icon_only: :when_narrow) { "Button #{i + 1}" }
+          end
+        end
+      end
+
+      private
+
+      # Lookbook won't mix types, so this normalizes the argument to boolean or symbol
```
camertron, line 68: Another reason to perhaps consider `:never` and `:always` in place of `true` and `false`?

camertron (approved):
A few comments but nothing blocking. Looking good! 🎉

mperrotti:
I took a look at the rendered examples at https://jonrohan-primer-devspace-wr644pf5vq5-3000.githubpreview.dev/lookbook/primer/alpha/segmented_control_preview/default

The way this is marked up is preventing the dividers from being shown between each button, and `:hover`/`:active` styles are not being shown.

See this issue for more info: https://github.com/primer/css/issues/2204

mperrotti:
Looking at https://primer-e5b2c477c1-26611710.drafts.github.io/components/alpha/segmentedcontrol, I'm not seeing the  selected button styles.

Screen Shot 2022-08-16 at 4 51 20 PM

app/components/primer/alpha/segmented_control.rb
mperrotti: This got buried in "hidden items". Mentioning @jonrohan so it doesn't get lost.

mperrotti: It looks like the feedback about the `icon_only` arg moving from the parent to the children has been addressed 👍

app/components/primer/alpha/segmented_control/button.html.erb
```diff
@@ -0,0 +1,11 @@
+<li class="SegmentedControl-item" role="listitem" aria-current="<%= @selected %>">
```
mperrotti, line 1: In PRC, I have `aria-current` on the button not the list item. Should I update my implementation to put it on the list item?

cc @bolonio (this week's a11y first responder)

mperrotti: Nevermind @bolonio - this was answered in today's office hours. It should be on the button.

app/components/primer/alpha/segmented_control.pcss
```diff
@@ -0,0 +1,170 @@
...
+
+  &:hover .SegmentedControl-content {
+    background-color: var(--color-segmented-control-button-hover-bg);
+    transition-duration: var(--primer-duration-fast, 80ms);
+  }
+
+  &:active .SegmentedControl-content {
+    background-color: var(--color-segmented-control-button-active-bg);
+    transition-duration: 0;
+  }
```
mperrotti, line 25: `:hover` and `:active` styles shouldn't apply to active segmented control buttons.

Kapture 2022-08-17 at 20 36 32

mperrotti (approved):

mperrotti (approved):

app/components/primer/alpha/segmented_control.ts
```diff
@@ -0,0 +1,38 @@
+import {controller, targets} from '@github/catalyst'
+
+@controller
+export class SegmentedControlElement extends HTMLElement {
+  @targets buttons: HTMLElement[]
```
camertron, line 5: It doesn't look like this is used anywhere

app/components/primer/alpha/segmented_control.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  connectedCallback() {
+    this.#updateButtonLabels()
+  }
+
+  select(event: Event) {
+    const button = event.currentTarget as HTMLButtonElement
+    for (const item of this.querySelectorAll('li.SegmentedControl-item')) {
```
camertron, line 13: Why not use the `buttons` target defined above instead of using `querySelector`?

app/components/primer/alpha/segmented_control.ts
```diff
@@ -0,0 +1,38 @@
...
+  // This if for selection styling to avoid the text jumping. It only needs to be
+  // setup when the component is first loaded.
+  #updateButtonLabels() {
+    for (const label of this.querySelectorAll('.Button-label')) {
+      label.setAttribute('data-content', label.textContent || '')
+    }
+  }
+}
+
+declare global {
```
camertron, line 32: Is this necessary if you use the `@controller` decorator?

camertron (changes requested):

app/components/primer/alpha/segmented_control.ts
jonrohan: I believe it is? At least some of our other catalyst components do this, 🤷🏻

app/components/primer/alpha/segmented_control.ts
jonrohan: Good catch 👍🏻

app/components/primer/alpha/segmented_control.html.erb
```diff
@@ -0,0 +1,12 @@
+<segmented-control>
+  <%= render Primer::BaseComponent.new(tag: :ul, role: "list", **@system_arguments) do %>
+    <% buttons.each do |button| %>
+      <li class="<%= class_names(
```
camertron, line 4: Does it make sense for this `<li>` to be emitted by the `Button` instead of here?

app/components/primer/alpha/segmented_control.ts
```diff
@@ -0,0 +1,41 @@
...
+      item.classList.remove('SegmentedControl-item--selected')
+      item.querySelector('[aria-current]')?.setAttribute('aria-current', 'false')
+    }
+
+    button.closest('li.SegmentedControl-item')?.classList.add('SegmentedControl-item--selected')
+    button.setAttribute('aria-current', 'true')
+  }
+
+  // Updates the button labels to have a data-content attribute with the text
+  // This if for selection styling to avoid the text jumping. It only needs to be
```
camertron, line 23:
```suggestion
  // This is for selection styling to avoid the text jumping. It only needs to be
```

app/components/primer/alpha/segmented_control/button.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if @icon_only == :always %>
+  <%= render Primer::Beta::IconButton.new(icon: @icon, "aria-label": content, **@system_arguments) %>
+<% else %>
+  <%= render Primer::Beta::Button.new(**@system_arguments) do |button| %>
+    <% button.with_leading_visual_icon(icon: @icon) unless @icon.nil? %>
+    <% button.tooltip(text: content) if @icon_only == :when_narrow %>
+    <%= content %>
```
camertron, line 7: Is it ok to allow arbitrary content in the button? Could we get by with a text-only attribute like `label:` instead?

camertron (approved):

test/components/component_test.rb
```diff
@@ -10,6 +10,13 @@ class PrimerComponentTest < Minitest::Test
   COMPONENTS_WITH_ARGS = [
     [Primer::Beta::IconButton, { icon: :star, "aria-label": "Star" }],
     [Primer::Beta::Button, {}],
+    [Primer::Alpha::SegmentedControl, {
+      icon_only: :never,
```
camertron, line 14: Did this option get removed?

camertron (approved):

PR #1226, opened 2022-07-25
Fix for flash_preview rename

jonrohan:
Screen Shot 2022-07-25 at 4 41 02 PM

Noticed that the FlashPreview didn't get renamed when the file was renamed.


langermank (approved):
This has been driving me crazy 😅

joelhawksley:
Is there a way we could have a test fail on this error next time?

PR #1233, opened 2022-07-27
LiveReload lookbook when primer_view_components.js

jonrohan:
I wanted to have lookbook watch for `*.ts` changes in `app/components` and reload when changes are made. This adds config to watch the compiled version for changes `app/assets/javascripts` and hard reload when that happens.

1. Procfile ts watches and compiles *.ts files
2. Procfile rollup watches compiled *.js files and builds final js
3. Hotwire watches final js file and reloads


camertron (approved):
Awesome! 🎉

PR #1238, opened 2022-07-28
Merge primer/rails_forms into primer/view_components

camertron:
### Purpose

Last quarter we made the conscious decision to home all the forms framework stuff in its own repo called https://github.com/primer/rails_forms with the acknowledgement that doing so enabled us to move faster in the short-term but that the code should ultimately live in primer/view_components. We also decided it made sense to rename primer/view_components to primer/rails, since the repo would no longer only contain components, but also forms-related code.

As you know, we decided to halt work on the forms framework this quarter, but would like to make good on our promise to unite the two repositories to enable other use-cases (eg. autocomplete). I've written up an issue delineating what I believe to be the concrete steps we need to take to achieve those goals: https://github.com/github/primer/issues/1143 (staff-only).

This PR is the result of tackling the first and second items on the list. We won't be attempting the other items on the list until later in the quarter.

### Changes

In addition to copy/pasting all the files from PRF into PVC, making sure the tests run as expected, and exposing the documentation for `TextField`, I made several other facilitating changes:

1. Dropped support for Ruby 2.6 and Rails 5.2. It is fairly common practice in the Ruby world to support the last three minor versions, and the consensus on the team was that these older versions weren't important to keep around. We now support Ruby 2.7, 3.0, and 3.1, and Rails 6.1, 7.0, and main.
2. Added a number of tests missing from the original primer/rails_forms code (thanks for catching those, Simplecov).
3. Augmented the Lookbook's bin/dev script to accept a `--debug` option (ala dotcom) that starts all the non-web processes in the background and then runs the Rails server in the foreground. Doing so means debugging tools like `binding.irb` and `binding.pry` work as expected.
4. Upgraded form inputs to handle CSS classes with `Primer::Classify` the same way PVC components do. This means arguments like `my: 4`, etc can be passed to inputs.
5. Added documentation for the new `TextField` component, including examples.


app/components/primer/alpha/text_field.rb
```diff
@@ -0,0 +1,116 @@
...
+      #   ) %>
+      #
+      # @param name [String] Value for the HTML name attribute.
+      # @param id [String] Value for the HTML id attribute.
+      # @param class [String] A list of CSS classes to add to the input. Exists for compatibility with Rails form builders.
+      # @param classes [String] A list of CSS classes to add to the input. Combined with the `:class` argument.
+      # @param caption [String] Caption text to render below the input.
+      # @param label [String] Label text displayed above the input.
+      # @param show_label [Boolean] Whether or not to visually display the label. If `true` the label will still be available to screen readers.
+      # @param trailing_label [String] Text to display to the right of the input.
```
langermank, line 99: Do we have a use case to support this param? I don't think we want to support this design, so I would remove this option.

app/components/primer/alpha/text_field.rb
```diff
@@ -0,0 +1,116 @@
...
+      # @param disabled [Boolean] Whether or not the input should accept keyboard and mouse input.
+      # @param invalid [Boolean] If `true`, renders the input in a visually invalid state.
+      # @param placeholder [String] Placeholder text.
+      # @param inset [Boolean] If `true`, renders the input in a visually inset state.
+      # @param monospace [Boolean] If `true`, uses a monospace font for the input field.
+      # @param leading_visual [Hash] Renders a leading visual icon before the text field's cursor. The hash will be passed to Primer's [Octicon component](https://primer.style/view-components/components/octicon).
+      # @param validation_message [String] A validation message to display beneath the input. Implicitly sets `invalid` to `true`.
+      # @param label_arguments [Hash] System arugments passed to the Rails builder's `#label` method. These arguments will appear as HTML attributes on the `<label>` tag.
+      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+      # @param block [Proc] Unused.
```
langermank, line 113: Can we remove this since `full_width` replaced it?

app/components/primer/alpha/text_field.rb
```diff
@@ -0,0 +1,116 @@
...
+      #     )
+      #   ) %>
+      #
+      # @param name [String] Value for the HTML name attribute.
+      # @param id [String] Value for the HTML id attribute.
+      # @param class [String] A list of CSS classes to add to the input. Exists for compatibility with Rails form builders.
+      # @param classes [String] A list of CSS classes to add to the input. Combined with the `:class` argument.
+      # @param caption [String] Caption text to render below the input.
+      # @param label [String] Label text displayed above the input.
+      # @param show_label [Boolean] Whether or not to visually display the label. If `true` the label will still be available to screen readers.
```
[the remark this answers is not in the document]
camertron: Ah right, thank you. Forgot to change this.

app/components/primer/alpha/text_field.rb
camertron: I added this to support the "Artifact and Log Retention" form:

image

However, I never ended up landing the migration for it, so we could remove it I guess?

app/components/primer/alpha/text_field.rb
camertron: Oh this is a dumb Ruby thing, i.e. is referring to a Ruby block (`Proc` object) rather than a `display: ` value. Some form inputs like `RadioButtonGroupInput` take a Ruby block that lets you define constituent radio buttons, eg:

```ruby
my_form.radio_button_group(name: :foo) do |group| # <-- this thing between the do/end keywords is a block
  group.radio_button(label: "bar")
  group.radio_button(label: "baz")
end
```

The constructor defined for this component lives in form_components.rb and accepts a block that may or may not be provided depending on what the input accepts. The code in form_components.rb simply passes it along to the input. I tried excluding it from the docs but Yard complained it existed in the code but wasn't documented.

app/components/primer/alpha/text_field.rb
langermank: Ahh right. Lets remove it for now and re-evaluate. A similar case came up recently for Primer Brand, so we can align with that solution if they decide to add in a property.

jonrohan (approved):
This looks great to me

PR #1242, opened 2022-07-29
Make sure `COMPONENT_TO_USE_INSTEAD` keys are aligned with officially deprecated components

khiga8:
This should be merged after https://github.com/primer/view_components/pull/1241.

Adding a CI check that makes sure officially deprecated components and components in `COMPONENT_TO_USE_INSTEAD` are aligned. 

This ensures:

- if someone adds a component to `COMPONENT_TO_USE_INSTEAD` linter without officially deprecating the component, the check fails.
- if someone deprecates a component officially with update to `statuses.json` and doesn't add the component to `COMPONENT_TO_USE_INSTEAD`, the check fails.
- both these will ensure components are officially deprecated & it is linted against by the `DeprecatedComponents` linter.


khiga8:
hmm actually I think I'll convert this to an actual check in `ci.yml` since this is dependent on the current test set up. it would stop working if someone changes the test to not use the actual JSON.

lib/primer/view_components/linters/helpers/deprecated_components_helpers.rb
```diff
@@ -31,14 +31,7 @@ def statuses_json
         end
 
         def deprecated_components
-          @deprecated_components ||= statuses_json.select { |_, value| value == "deprecated" }.keys.tap do |deprecated_components|
-            deprecated_components.each do |deprecated|
-              unless COMPONENT_TO_USE_INSTEAD.key?(deprecated)
```
khiga8, line 36: I decided this makes more sense as a CI check.

.github/workflows/ci.yml
```diff
@@ -174,6 +174,20 @@ jobs:
...
+    - name: Setup Ruby
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 3.0.x
+    - name: Make sure deprecated components can be linted
+      run: |
+        gem install bundler:2.2.9
+        bundle config path vendor/bundle
+        bundle install --jobs 4 --retry 3
+        bundle exec rake deprecated:check
```
khiga8, line 190: should I move this check elsewhere?

would this make more sense in `.github/workflows/docs-build.yml` after the `bundle exec rake static:dump`? 

last I remember the docs build action was funky so wasn't sure

camertron: I think it makes sense here since it doesn't really have to do with docs. What if we added `bundle exec rake static:dump` before the `rake deprecated:check` here? It's doing the work again but should be fast enough.

lib/tasks/deprecated.rake
```diff
@@ -0,0 +1,28 @@
...
+    require "json"
+    require_relative "./../primer/view_components/linters/helpers/deprecated_components_helpers"
+
+    puts "Checking that officially deprecated components are linted by `DeprecatedComponents` linter...."
+
+    statuses_json = JSON.parse(
+      File.read(
+        File.join(File.dirname(__FILE__), "../../static/statuses.json")
+      )
+    ).freeze
```
camertron, line 14:
```suggestion
    require_relative "../primer/view_components/statuses"
    statuses = Primer::ViewComponents::STATUSES
```

camertron (changes requested):

.github/workflows/ci.yml
khiga8: Oooh makes sense to me 👍

camertron (approved):
🎉

PR #1244, opened 2022-07-29
`ClipboardCopy` only needs `aria-label` when empty

### LINKED ISSUE #1081: `ClipboardCopy` component requires `aria-label` even when text is set
https://primer.style/view-components/components/clipboardcopy#with-text-instead-of-icons

#### Current behavior
`aria-label` is required even when button has text content.  This causes people to set `aria-label` that doesn't match, and overrides the button content.

#### Desired behavior
- Don't require `aria-label` when button has text content

neall:
We were requiring `aria-label` on `ClipboardCopy` in all cases, but when the component already contains suitable text this is not recommended.

This changes when we validate the presence of `aria-label`. At instantiation time the component doesn't know if it will be rendered with a block, so we check at render time instead.

Fixes #1081

### DIFF
--- a/app/components/primer/clipboard_copy.rb
+++ b/app/components/primer/clipboard_copy.rb
@@ -12,7 +12,7 @@ module Primer
     #   <%= render(Primer::ClipboardCopy.new(value: "Text to copy", "aria-label": "Copy text to the system clipboard")) %>
     #
     # @example With text instead of icons
-    #   <%= render(Primer::ClipboardCopy.new(value: "Text to copy", "aria-label": "Copy text to the system clipboard")) do %>
+    #   <%= render(Primer::ClipboardCopy.new(value: "Text to copy")) do %>
     #     Click to copy!
     #   <% end %>
     #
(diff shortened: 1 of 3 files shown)


app/components/primer/clipboard_copy.rb
```diff
@@ -34,10 +34,16 @@ def initialize(value: nil, **system_arguments)
       @system_arguments[:value] = value if value.present?
     end
 
+    # :nodoc:
+    def render_in(*args, &block)
+      validate_aria_label unless block_given?
```
camertron, line 39: Hmm, I'm not sure checking the presence of a block is the right approach here. The component could ostensibly be passed an empty block, or someday have slots that must be defined in the block. What would you think about something like:

```ruby
def before_render
  validate_aria_label if content.blank?
end
```

camertron (changes requested):

app/components/primer/clipboard_copy.rb
neall: They can always give some content that is *not* suitable for screen readers, but explicitly checking for blank sounds good. `before_render` also sounds like a better place for this check than hooking into `render_in` as well.

camertron (approved):

jonrohan (approved):

PR #1247, opened 2022-07-29
Set `aria-hidden="true"` when type is label

khiga8:
This PR makes a change so that when the tooltip type is `label` establishing an `aria-labelledby` relationship, the tooltip element will have `aria-hidden="true"`. The reason for this change is to avoid screen reader users needing to hear duplicated content that they already hear when they navigate on the control itself.

This change has been made based on an [internal Slack conversation](https://github.slack.com/archives/C0FSWLQ0Y/p1659132347317399) and recommendation from @jscholes.


khiga8:
To do on Monday: share codespaces URL with @jscholes

camertron (approved):

Issue #1048, opened 2022-03-04
Codespaces environment doesn't load the local JavaScript bundle


hectahertz:
# Problem

When working on Codespaces, the `primer_viewcomponents.js` bundle is not properly loaded as it always tries to fetch it from `localhost:4000` which is not available.

This is also an issue when modifying the `bin/rails` port on the `Procfile` in local environments (which can useful sometimes).

See the code with this logic [here](https://github.com/primer/view_components/blob/main/docs/src/@primer/gatsby-theme-doctocat/components/head.js#L25).

Related to https://github.com/primer/view_components/issues/1042

# Proposed solution

Modify that logic so it is dynamically set and supports Codespaces and, ideally, running processes on different ports.

hectahertz:
Related to https://github.com/primer/view_components/issues/1042

khiga8:
I think I addressed this in: https://github.com/primer/view_components/pull/1248, https://github.com/primer/view_components/pull/1259 🚀

Issue #1042, opened 2022-03-03
Preview Deploys on Pull Requests with JavaScript changes don't include those changes


hectahertz:
# Problem

When there's a Pull Request with JavaScript changes the Preview Deploy doesn't include those changes as it loads the released version instead.

See the code with this logic [here](https://github.com/primer/view_components/blob/main/docs/src/@primer/gatsby-theme-doctocat/components/head.js#L25).

# Proposed solution

Modify that logic so it treats Preview Deploy environments as development environments.

hectahertz:
Related https://github.com/primer/view_components/issues/1048

khiga8:
I think I addressed this in: https://github.com/primer/view_components/pull/1248, https://github.com/primer/view_components/pull/1259 🚀

PR #1248, opened 2022-07-30
copy JS to static docs directory and use that instead

khiga8:
Trying to fix:

- https://github.com/primer/view_components/issues/1048
- https://github.com/primer/view_components/issues/1042

Currently I'm unable to share a documentation site with my JS changes because the docs preview and codespaces both pull in the published JS package in npm rather than the changes on my branch. 

This means even when code is merged, people are unable to see JS changes on the docs site until a new PVC release is cut which is not ideal since the JS on the docs site should reflect whats in the codebase.

This is a workaround I used in the past but I'm not sure it's actually what we want.


package.json
```diff
@@ -26,7 +26,7 @@
   "scripts": {
     "build:docs": "cd docs && yarn gatsby build --prefix-paths",
     "build:docs:preview": "cd docs && yarn gatsby build",
-    "prepare": "tsc && rollup -c",
+    "prepare": "rm -rf docs/static && mkdir docs/static && tsc && rollup -c && cp app/assets/javascripts/primer_view_components.js docs/static/primer_view_components.js",
```
camertron, line 29: Is there anything else in docs/static that this command is blowing away?

camertron (approved):
I think this is probably fine. I do wish the js world had a better way of including local packages though.

package.json
khiga8: Nope, this is a new directory!

PR #1256, opened 2022-08-01
Use `visibility: 'hidden'` instead of `hidden` attribute

khiga8:
## Context

Screen reader users may use shortcuts like `B` to navigate to a button control without actually focusing on the button element itself. The expected behavior is for any accessible name and description to be announced.

@jscholes discovered that when navigating to a button control in this way without focusing the element, the accessible name or description provided by the tooltip wasn't being announced. We observed that the label and description is only announced after the button is focused on and `hidden` attribute removed.

This is unexpected behavior, likely due to some bug in the accessibility API or web component API. The use of `hidden` attribute shouldn't prevent an element from being labelled. I tested this with a regular `span` and observed the association was created fine. 

## Impact 

To work around this bug, this PR swaps out the `hidden` attribute for `visibility: hidden;`. This will achieve the expected behavior of:

- tooltip labels or describes the element on screen reader navigation (such as with `B`)
- tooltip labels or describes the element on focus of the element.
- tooltip can be navigated to when the element is focused.

This behavior has been tested with:
- Safari/VoiceOver
- Chrome/NVDA

@jscholes and I discussed that in the future, we may want to revisit the idea of using `sr-only` class so the tooltip content is available at all times for screen readers.

## Before video

In this video I am on my Windows VM using NVDA. I navigate the `With direction` tooltip page examples using the `B` shortcut to jump between button controls. The focus does not change. I hear "North button, South button, East button", etc. However, the tooltip content associated as a description is not being announced.

https://user-images.githubusercontent.com/16447748/182234517-137fe1a8-93a3-4d04-8a60-0225376903f3.mp4

## After video

In this video, I have a similar setup as before. However this time, when I navigate using the `B` shortcut, I hear "North button this is a north facing tooltip... South button this is a south facing tooltip."

https://user-images.githubusercontent.com/16447748/182234503-55da726b-da92-4fc8-9dee-8ad22b8d507a.mp4


test/components/component_test.rb
```diff
@@ -110,8 +110,8 @@ def test_all_components_pass_through_classes
   def test_all_components_support_inline_styles
     default_args = { style: "width: 100%;" }
     COMPONENTS_WITH_ARGS.each do |component, args, proc|
-      render_component(component, default_args.merge(args), proc)
-      assert_selector("[style='width: 100%;']", visible: :all)
+      node = render_component(component, default_args.merge(args), proc)
+      assert_equal(true, node.children.attribute("style")&.value&.include?("width: 100%"))
```
khiga8, line 114: `assert_selector` looks for an exact match which we won't see with `Primer::Alpha::Tooltip` so I updated this assertion.

app/lib/primer/join_style_arguments_helper.rb
```diff
@@ -8,7 +8,7 @@ module JoinStyleArgumentsHelper
     # join_style_arguments("width: 100%", "height: 100%") =>
     #   "width: 100%;height: 100%"
     def join_style_arguments(*args)
-      args.compact.join(";")
+      args.compact.join(";").gsub(";;", ";")
```
khiga8, line 11: I saw that this `join_style_arguments` can result in extra semicolons being added so this strips it.

test/components/component_test.rb
```diff
@@ -111,7 +111,11 @@ def test_all_components_support_inline_styles
     default_args = { style: "width: 100%;" }
     COMPONENTS_WITH_ARGS.each do |component, args, proc|
       render_component(component, default_args.merge(args), proc)
-      assert_selector("[style='width: 100%;']", visible: :all)
+      if component == Primer::Alpha::Tooltip
+        assert_selector("[style='width: 100%;visibility: hidden']", visible: :all)
+      else
+        assert_selector("[style='width: 100%;']", visible: :all)
+      end
```
khiga8, line 118: `assert_selector` looks for an exact match. I couldn't find an alternate matcher. I found [assert_matches_style](https://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FMatchers:assert_matches_style) but its not available even when I upgraded capybara.
if reviewer has ideas here...

camertron, line 118: It looks like `#assert_matches_style` is designed to be called on a Capybara `Node` object. However it also looks like `assert_selector` supports the `style:` argument, so maybe this works?

```suggestion
      assert_selector(style: { "width" => "100%" })
```

app/lib/primer/join_style_arguments_helper.rb
camertron: Hmm why not remove trailing semicolons before joining? Maybe something like

```suggestion
      args.compact.map { |a| a.strip.chomp(";") }.join(";")
```

camertron (changes requested):

test/components/component_test.rb
khiga8: Hmm this syntax appears correct but looks like something could be broken:

```
Nokogiri::CSS::SyntaxError: unexpected '$' after ''
    (eval):3:in `_racc_do_parse_c'
    (eval):3:in `do_parse'
```

khiga8: I think we might need to do somethign gross like:

```.rb
rendered = render_component(component, default_args.merge(args), proc)
assert_equal(true, rendered.inner_html.include?('style="width: 100%;'))
```

camertron: Hmm that's a good idea. You should be able to use `page.body` I think? The result of the render is also available as `@rendered_content` or `rendered_content`, but that's technically part of the private API.

jonrohan (approved):

PR #1257, opened 2022-08-01
Revert "copy JS to static docs directory and use that instead"

khiga8:
Reverts primer/view_components#1248

I think something broke with the JS on the Primer site.
See [tooltip page](https://primer.style/view-components/components/alpha/tooltip#as-a-description-for-a-button-with-visible-label). JS doesn't load.


khiga8:
actually lets try https://github.com/primer/view_components/pull/1259 first

khiga8:
fixed by https://github.com/primer/view_components/pull/1259

PR #1259, opened 2022-08-01
fix script tag to fix production JS

khiga8:
follow up to https://github.com/primer/view_components/pull/1248 which seems to have broken the JS in production docs

I think this should fix it but if not lets revert.

### DIFF
--- a/docs/src/@primer/gatsby-theme-doctocat/components/head.js
+++ b/docs/src/@primer/gatsby-theme-doctocat/components/head.js
@@ -1,3 +1,4 @@
+import { withPrefix } from "gatsby"
 import React from 'react'
 import Helmet from 'react-helmet'
 import useSiteMetadata from '@primer/gatsby-theme-doctocat/src/use-site-metadata'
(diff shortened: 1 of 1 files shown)


docs/src/@primer/gatsby-theme-doctocat/components/head.js
```diff
@@ -28,7 +29,7 @@ function Head(props) {
       <meta property="og:description" content={description} />
       <meta property="og:image" content={siteMetadata.imageUrl} />
       <meta property="twitter:card" content="summary_large_image" />
-      <script src={primerViewComponentsSrc}></script>
+      <script src={withPrefix(primerViewComponentsSrc)} type="text/javascript"></script>
```
khiga8, line 32: [from gatsby docs:](https://www.gatsbyjs.com/docs/how-to/previews-deploys-hosting/path-prefix/#add-the-path-prefix-to-paths-using-withprefix)

> For pathnames you construct manually, there’s a helper function, [withPrefix](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#add-the-path-prefix-to-paths-using-withprefix) that prepends your path prefix in production (but doesn’t during development where paths don’t need to be prefixed).

camertron (approved):
Hmm ok let's give it a try?

PR #1260, opened 2022-08-01
Move tooltip to be sibling of link

khiga8:
The `Link` component now supports rendering of tooltip through its API.

Currently the tooltip element is nested within the anchor element. However this will result in the tooltip element becoming a touch area for the link which is unexpected behavior. 

Instead the tooltip element should be rendered adjacent to the anchor.

Additionally, to ensure the tooltip (which uses absolute positioning) is positioned correctly, this PR adds a `position: 'relative` container that wraps the tooltip and the anchor element. I made this a `span` to mimic `inline` positioning of  link.

### DIFF
--- a/app/components/primer/link_component.rb
+++ b/app/components/primer/link_component.rb
@@ -17,6 +17,8 @@ module Primer
 
     # `Tooltip` that appears on mouse hover or keyboard focus over the link. Use tooltips sparingly and as a last resort.
     # **Important:** This tooltip defaults to `type: :description`. In a few scenarios, `type: :label` may be more appropriate.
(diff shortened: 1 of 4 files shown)


app/components/primer/link_component.rb
```diff
@@ -43,9 +45,6 @@ class LinkComponent < Primer::Component
     # @example Without underline
     #   <%= render(Primer::LinkComponent.new(href: "#", underline: false)) { "Link" } %>
     #
-    # @example Span as link
```
khiga8, line 46: I don't think this is a good usecase so we should remove this example.

PR #1261, opened 2022-08-01
Make all subdirs of docs/static in prepare script

camertron:
I think this is causing the storybook deploy issue evidenced [here](https://github.com/primer/view_components/runs/7621342012?check_suite_focus=true).


khiga8 (approved):
thank you!

jonrohan (approved):

PR #1265, opened 2022-08-02
Update IconButton and Button tooltips to be sibling instead of child

jonrohan:
This work does the same thing @khiga8 did in https://github.com/primer/view_components/pull/1260 for similar reasons that the Tooltip content shouldn't exist inside the button.

### DIFF
--- a/app/components/primer/button_component.html.erb
+++ b/app/components/primer/button_component.html.erb
@@ -1,5 +1,5 @@
 <% if tooltip.present? %>
-  <%= render Primer::BaseComponent.new(tag: :span, position: :relative) do %>
+  <%= render Primer::BaseComponent.new(tag: :div, position: :relative, display: :inline_block) do %>
     <%= render Primer::Beta::BaseButton.new(**@system_arguments) do -%>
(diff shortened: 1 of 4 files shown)


app/components/primer/button_component.html.erb
```diff
@@ -1,4 +1,12 @@
-<%= render Primer::Beta::BaseButton.new(**@system_arguments) do -%>
-  <%= leading_visual %><%= trimmed_content %><%= trailing_visual %><%= primer_octicon("triangle-down", ml: 2, mr: -1) if @dropdown %>
-  <%= tooltip %>
-<% end -%>
+<% if tooltip.present? %>
+  <%= render Primer::BaseComponent.new(tag: :span, position: :relative) do %>
```
khiga8, line 2: should we set the display to [inline-block](https://github.com/primer/css/blob/main/src/buttons/button.scss#L6) to preserve what .btn usually renders as? maybe`tag: :div` and `display: :inline-block`. 

similar comment for `IconButton`

khiga8 (approved):
Niceeee! Thanks for doing this!

PR #1268, opened 2022-08-02
ActionList and NavList components

camertron:
This PR introduces the `ActionList` and `NavList` components. Interface guidelines for `ActionList` can be found [here](https://primer.style/design/components/action-list). `NavList` replaces the existing experimental `NavigationList` component inside dotcom, and uses `ActionList` as its base. `ActionList` is intended to be used only as a base class and should not be used directly by consumers.

Please see the Lookbook examples and documentation pages for more details.

image


app/components/primer/alpha/action_list_item.rb
```diff
@@ -0,0 +1,136 @@
...
+      DESCRIPTION_VARIANT_OPTIONS = DESCRIPTION_VARIANT_MAPPINGS.keys.freeze
+
+      DEFAULT_VARIANT = :default
+      VARIANT_MAPPINGS = {
+        DEFAULT_VARIANT => nil,
+        :danger => "ActionList-item--danger"
+      }.freeze
+      VARIANT_OPTIONS = VARIANT_MAPPINGS.keys.freeze
+
+      DEFAULT_SELECT_MODE = :none
```
langermank, line 30: I'm wondering if we might be able to remove this select mode check. For single select, we'll want `aria-selected="true"` on the `li`, and for multi `aria-checked="true"`. So maybe it can be either `checked` or `selected`?

Additionally, we should make sure `aria-etc` is only set when these are true and removed from the dom when `false`. Its just extra noise we don't need.

app/components/primer/alpha/action_list_item.rb
```diff
@@ -0,0 +1,136 @@
...
+      SELECT_MODE_ICON_MAPPINGS = {
+        DEFAULT_SELECT_MODE => nil,
+        :single => "ActionList-item--singleSelectCheckmark",
+        :multiple => "ActionList-item-multiSelectIcon"
+      }.freeze
+      SELECT_MODE_OPTIONS = SELECT_MODE_ICON_MAPPINGS.keys.freeze
+
+      renders_one :description
+
+      renders_one :leading_visual, types: {
```
langermank, line 40: I'd like to offer something similar for both `leading_action` and `trailing_action` while still configuring support for the two checked/selected states. Is that doable?

app/components/primer/alpha/action_list_item.html.erb
```diff
@@ -0,0 +1,59 @@
+<%= render(Primer::BaseComponent.new(tag: :li, **@system_arguments)) do %>
+  <%= render(Primer::BaseComponent.new(tag: :span, **@content_arguments)) do %>
```
langermank, line 2: For this tag, we want people to use `span` with an `onClick` for button type interactions, and `a href` for links. How can we pass those options down?

app/components/primer/alpha/action_list_item.html.erb
```diff
@@ -0,0 +1,59 @@
+<%= render(Primer::BaseComponent.new(tag: :li, **@system_arguments)) do %>
```
langermank, line 1: This tag needs a `role` param.. I suppose with `listitem` as default but I may change that later :)

app/components/primer/alpha/action_list_item.rb
camertron: Hmm ok I see what you mean. Is the idea then that `checked` would imply a checkbox and `selected` a check mark?

app/components/primer/alpha/action_list_item.rb
camertron: I think so. What sorts of actions did you have in mind?

app/components/primer/alpha/action_list_item.html.erb
camertron: Oh good call. Dotcom's `NavigationList` checks for the presence of an `href` attribute and uses an `a` tag if one is given. I can copy that logic over here. Sounds like we also need to expose an `on_click` attribute.

app/components/primer/alpha/action_list_item.html.erb
camertron: Ok, I think it's better if the role is passed in, but we can use `listitem` as a default.

app/components/primer/alpha/action_list_item.rb
camertron: It now occurs to me you were actually saying we might not need the `:none` case. I agree and will remove it.

langermank: Okay, I think having the select mode option is correct. We do need a `:none` case, but that should be default. So you should be able to opt into single or multi select. I think the none state isn't working right now.

app/components/primer/alpha/action_list_item.rb
langermank: I added this as WIP 😄

app/components/primer/alpha/nav_list/item.rb
```diff
@@ -0,0 +1,20 @@
...
+      class Item < Primer::Alpha::ActionList::Item
+        def initialize(selected_item_id: nil, selected_by_ids: [], **system_arguments)
+          @selected_item_id = selected_item_id
+          @selected_by_ids = Array(selected_by_ids)
+
+          overrides = { select_mode: :none, "data-item-id": @selected_by_ids.join(" ") }
+          overrides[:active] = true if @selected_by_ids.include?(@selected_item_id)
+
+          super(**system_arguments, **overrides)
+        end
```
jonrohan, line 16:
```ruby        
        def before_render
          self.private_leading_action(...)
        end
```

app/components/primer/alpha/action_list/item.rb
```diff
@@ -0,0 +1,230 @@
...
+            DESCRIPTION_SCHEME_OPTIONS, description_scheme, DEFAULT_DESCRIPTION_SCHEME
+          )
+
+          @system_arguments[:classes] = class_names(
+            @system_arguments[:classes],
+            SCHEME_MAPPINGS[@scheme],
+            "ActionListItem",
+            # "ActionListItem--withActions",
+            # "ActionListItem--trailingActionHover",
+            # "ActionListItem--withActions" if trailing_action.present? this isn't working,
```
langermank, line 136: @joelhawksley I couldn't get these conditional classnames to work, can you take a look?

app/components/primer/alpha/action_list/item.rb
```diff
@@ -0,0 +1,246 @@
...
+        def expand!
+          # @expanded = true
+        end
+
+        private
+
+        def before_render
+          raise "Cannot render a trailing visual for an item with subitems" if items.present? && trailing_visual.present?
+
+          if items.present?
```
langermank, line 220: @joelhawksley this bit thats commented out is what's missing from the "sub group" logic.

app/components/primer/alpha/action_list/heading.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render(Primer::BaseComponent.new(**@system_arguments)) do %>
+  <%= render(Primer::BaseComponent.new(tag: :h3, classes: "ActionList-sectionDivider-title", id: @section_id)) do %>
```
langermank, line 2: @joelhawksley this heading level came up last week from an accessibility issue in Discussions. Can we add a heading level param so it can be set per implementation, with a default of `h3`?

langermank: PR for reference: https://github.com/github/github/pull/236445/files

test/previews/primer/alpha/action_list_preview.rb
```diff
@@ -0,0 +1,176 @@
+# frozen_string_literal: true
+
+module Primer
+  module Alpha
+    # @label ActionList
+    class ActionListPreview < ViewComponent::Preview
+      # @label Sub lists
+      def sub_lists
```
langermank, line 8: I'm still a little bit confused with `sub-list`. It feels funny to me to repeat the `<action-list>` wrapper:

image

Also, if we want to move back towards the entire list always being a `ul`, the divider component should become an `li` instead of a `div

app/components/primer/alpha/action_list.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+module Primer
+  module Alpha
+    # An `ActionList` is a TBD
+    class ActionList < Primer::Component
+      DEFAULT_ROLE = :list
+      DEFAULT_TAG = :ul
```
langermank, line 8: @joelhawksley do you think we need this? I don't see a scenario where this _isn't_ a `ul`/`li`

app/components/primer/alpha/action_list.rb
```diff
@@ -0,0 +1,97 @@
...
+        end
+      }
+
+      def initialize(
+        tag: DEFAULT_TAG,
+        role: DEFAULT_ROLE,
+        item_classes: nil,
+        scheme: DEFAULT_SCHEME,
+        show_dividers: false,
+        child: false,
```
langermank, line 47: @joelhawksley should we call this subGroup? Thats what I had in the API doc but I'm not sure where it came from. I don't have a strong opinion either way.

test/previews/primer/alpha/action_list_preview.rb
camertron: I was able to get rid of the `<action-list>` wrapper. It only renders once per list now.

test/previews/primer/alpha/nav_list_preview.rb
```diff
@@ -0,0 +1,364 @@
+# frozen_string_literal: true
+
+module Primer
+  module Alpha
+    # @label NavList
+    class NavListPreview < ViewComponent::Preview
+      def playground
```
langermank, line 7: Can you add params that NavList has to this `Playground` preview, and also add a `Default` preview with a simple example?

I like the example with show more. Can you delete all the others copied from ActionList?

app/components/primer/primer.ts
```diff
@@ -1,4 +1,6 @@
+import './alpha/action_list_element'
 import './alpha/modal_dialog'
+import './alpha/nav_list_element'
```
jonrohan, line 3: Can you drop the `_element` suffix from these file names?

demo/config/routes.rb
```diff
@@ -12,6 +12,7 @@
   scope path: Rails.env.production? ? "/view-components" : "/" do
     get "/auto_complete", to: "auto_complete_test#index"
     resources :toggle_switch, only: [:create]
+    resources :nav_list_items, only: [:index]
```
jonrohan, line 15: I had trouble hooking up these routes for production, if you goto the live site now https://primer.style/view-components/lookbook/ the controller actions don't work. It's cause they'd need passthroughs from the DNS magic we have. But I thought maybe we could fix it here by changing the scope for them to `/view-components/lookbook-controllers/` ? Not blocking this pr, just be aware for when it merges.

camertron: Oh that's really interesting. Is it as simple as adding another scope around these `resources` calls?

test/previews/primer/alpha/nav_list_preview.rb
camertron: If I recall correctly, we decided to rename this playground preview to "default" and not expose any params, since the preview renders an entire list + sub items.

Yes, I have deleted the other previews copied from ActionList 😄

langermank: Yup! Give me a nice capitol `Default` for my mental health please 😆

app/components/primer/alpha/action_list/heading.rb
```diff
@@ -0,0 +1,38 @@
...
+        # @param tag [Symbol] Semantic tag for the heading.
+        # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
+        def initialize(list_id:, title:, tag: :h3, scheme: DEFAULT_SCHEME, subtitle: nil, **system_arguments)
+          @tag = tag
+          @system_arguments = system_arguments
+          @list_id = list_id
+          @title = title
+          @subtitle = subtitle
+          @scheme = fetch_or_fallback(SCHEME_OPTIONS, scheme, DEFAULT_SCHEME)
+          @system_arguments[:tag] = :div
```
langermank, line 28: I think this should be an `li` and it should render inside the list.. or did we decide the entire list won't be a `ul` like it is in dotcom?

jonrohan (approved):
Seems good 👍🏻 

Had one non-blocking question. Should we add to the Deprecations class for migrating from the old Navigation element in dotcom to this?

camertron:
Hey @jonrohan, yes! I have a dotcom branch that does just that for the old `Experimental::NavigationList` component.

app/components/primer/alpha/action_list/heading.rb
camertron: Discussed this with @langermank in-person. We decided to bring back the concept of "sections." While refactoring, I also turned the heading into an `<li>`.

langermank (approved):
Beautiful! 🚀

PR #1275, opened 2022-08-04
move Primer::ButonGroup to Primer::Beta::ButtonGroup

mxriverlynn:
split from #676 

this PR runs the component status migrator script against the `Primer::ButtonGroup` component, moving it to `Primer::Beta::ButtonGroup`. it also creates a deprecated `Primer::ButtonGroup` in place of the original, updates all references to the button group, and adjusts doc site and storybook navigation

## files to note in review

this PR is not terribly large compared to some, as there are few references to the button group from other components or examples. all files should be given thorough review

## screenshots

storybook:
image

docs site:
image


jonrohan (approved):
This is good to go once the merge conflicts are fixed

PR #1283, opened 2022-08-08
Updating label component to not always have large

jonrohan:
image

https://primer.style/view-components/components/label

Currently the Label is always rendered with the `Label--large` class. I fixed the issue here and added a label component preview.


camertron (approved):
Thanks :) 🎉

PR #1284, opened 2022-08-08
Don't render tooltip on IconButton when tag is summary

jonrohan:
Currently there's an issue with the `IconButton` component when being rendered with a `tag: :summary` option. 

image

The `<summary>` gets wrapped with a `div` that breaks any details dialog triggers.

I'm proposing that we don't render a tooltip when the tag is a summary. This is because to place it properly we would need to put the tooltip outside of the `<details>` element.

### DIFF
--- a/.changeset/poor-elephants-impress.md
+++ b/.changeset/poor-elephants-impress.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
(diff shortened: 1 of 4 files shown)


prashkan (approved):
Thanks for the quick remediation here! 🎉

PR #1291, opened 2022-08-10
Continue axe tests when there's a failure

owenniblock:
### Motivation

I've been looking over the axe tests to understand what they're doing. While doing so, I introduced an intentional issue and noticed that the axe tests halt on error.

This code change collates all axe errors and then raises an error at the end of the process.

### Additional Notes

I switched off the skip `:"color-contract"` rule and there were 13 axe failures reported:

> Error:
> IntegrationDocExamplesAxeTest#test_accessibility_of_doc_examples:
> RuntimeError: There were 13 axe failures:
> ["button_group/default", "button_group/sizes", "dropdown/customizing_the_button", "state_component/schemes", "alpha_tooltip/as_a_description_for_a_button_with_visible_label", "alpha_tooltip/with_relative_parent", "beta_blankslate/primary_and_secondary_actions", "beta_blankslate/primary_action", "beta_button_group/default", "beta_button_group/sizes", "blankslate_component/action_button", "button_component/block", "button_component/schemes"]

It might be worth us putting some effort in to resolving these contrast errors so we can switch this rule on.

I also tried to enable rule-sets using [according_to](https://github.com/dequelabs/axe-core-gems/blob/develop/packages/axe-core-rspec/README.md#according_to---accessibility-standard-tag-clause) and I expected to see failures related to the pages not having an `h1` heading but the tests all passed. This is strange behavior so I would like to see if there's a way to confirm which tests are actually run by this code.

### DIFF
--- a/test/system/doc_examples_axe_test.rb
+++ b/test/system/doc_examples_axe_test.rb
@@ -18,6 +18,8 @@ class IntegrationDocExamplesAxeTest < ApplicationSystemTestCase
     puts "Running axe-core checks on previews generated from documentation examples..."
     puts "============================================================================="
 
+    failure_urls = []
+
     Primer::Docs.constants.each do |klass|
       next if NOT_STANDALONE.include?(klass)
       next if IGNORE.include?(klass)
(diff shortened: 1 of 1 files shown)


joelhawksley (approved):

khiga8:
@owenniblock 

oooh, I can speak to:
 
> I expected to see failures related to the pages not having an h1 heading but the tests all passed.

If I remember correctly, these accessibility checks are running against each individual component examples in the doc, rather than testing the doc page directly. You can think of these as tests to make sure the component examples that people add to the docs are accessible rather than an accessibility check of the doctocat site itself.

We use [within: option](https://github.com/primer/view_components/blob/main/test/application_system_test_case.rb#L37) which makes it so axe runs against a specified component. I believe `within` option results in automatic exclusion of rules that depend on the full context of the page like headings and landmarks. I think there was a list somewhere but I'll have to dig around for it..

owenniblock:
Aha, thanks @khiga8 that makes total sense. I basically _wanted_ to run as many tests as possible and exclude the ones (like the h1 test) which don't make sense but it sounds like it does that automagically!

khiga8:
@owenniblock Yes!!! I'm pretty sure it does but now I can't find docs on which rules exactly are being run 😭. 

Maybe if we run the test locally and put a breakpoint in the browser, we can check the axe object and get the rules. Might be worth a shot when we have time!

PR #1292, opened 2022-08-10
Update slot use in examples for all components in the alpha folder

mxriverlynn:
this PR updates all components in the `alpha/` folder to use the new `.with_#{slot_name}` syntax in docs examples

additionally, an error was found with the `Alpha::Layout` component that prevented it's docs from generating and updating correctly. the `status :alpha` flag was missing, and the gatsby `nav.yml` link was pointing to an incorrect (old) location. both of these have been corrected.

## Updated components

* Alpha::AutoComplete (deprecated)
* Alpha::Layout
* Alpha::TabNav
* Alpha::TabPanels
* Alpha::UnderlineNav
* Alpha::UnderlinePanels

## Screenshots

layout:
image

tab nav:
image

tab panels:
image

underline nav:
image

underline panels:
image

### DIFF
--- a/app/components/primer/alpha/layout.rb
+++ b/app/components/primer/alpha/layout.rb
@@ -14,6 +14,8 @@ module Primer
     #   Keyboard navigation follows the markup order. Decide carefully how the focus order should be be by deciding whether
     #   `main` or `sidebar` comes first in code. The code order won’t affect the visual position.
     class Layout < Primer::Component
+      status :alpha
+
       FIRST_IN_SOURCE_DEFAULT = :sidebar
       FIRST_IN_SOURCE_OPTIONS = [FIRST_IN_SOURCE_DEFAULT, :main].freeze
 
--- a/app/components/primer/alpha/layout.rb
+++ b/app/components/primer/alpha/layout.rb
@@ -80,8 +80,8 @@ module Primer
       # @example Default
       #
       #   <%= render(Primer::Alpha::Layout.new) do |c| %>
-      #     <% c.main(border: true) { "Main" } %>
-      #     <% c.sidebar(border: true) { "Sidebar" } %>
+      #     <% c.with_main(border: true) { "Main" } %>
+      #     <% c.with_sidebar(border: true) { "Sidebar" } %>
       #   <% end %>
       #
       # @example Main widths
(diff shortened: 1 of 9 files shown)


app/components/primer/alpha/layout.rb
```diff
@@ -14,6 +14,8 @@ module Alpha
     #   Keyboard navigation follows the markup order. Decide carefully how the focus order should be be by deciding whether
     #   `main` or `sidebar` comes first in code. The code order won’t affect the visual position.
     class Layout < Primer::Component
+      status :alpha
```
mxriverlynn, line 17: this was missing, and i think it was causing issues with the docs being rendered

docs/content/components/label.md
```diff
@@ -31,7 +31,7 @@ Use `aria-label` if the `Label` or the context around it don't explain the label
...
+<Example src="<span data-view-component='true' class='Label'>Default</span><span data-view-component='true' class='Label Label--primary'>Primary</span><span data-view-component='true' class='Label Label--secondary'>Secondary</span><span data-view-component='true' class='Label Label--accent'>Accent</span><span data-view-component='true' class='Label Label--success'>Success</span><span data-view-component='true' class='Label Label--attention'>Attention</span><span data-view-component='true' class='Label Label--danger'>Danger</span><span data-view-component='true' class='Label Label--severe'>Severe</span><span data-view-component='true' class='Label Label--done'>Done</span><span data-view-component='true' class='Label Label--sponsors'>Sponsors</span>" />
```
mxriverlynn, line 34: i don't know why this file has changed, honestly. but it keeps showing up as modified on my local system whenever i do any work so i figured it should be committed. let me know if these changes should be dropped, tho

docs/src/@primer/gatsby-theme-doctocat/nav.yml
```diff
@@ -68,7 +68,7 @@
   - title: Label
     url: "/components/label"
   - title: Layout
-    url: "/components/layout"
+    url: "/components/alpha/layout"
```
mxriverlynn, line 71: this didn't get updated when layout was moved into the alpha folder, and that was causing old docs to show up on the site (production is currently showing old docs)

joelhawksley: 🤔 I wonder if we could find a way to automatically check for this?

docs/content/components/label.md
joelhawksley: I'm not sure. Maybe we just commit this diff and see what happens? 😂

app/components/primer/alpha/layout.rb
joelhawksley: Could we add a test to catch this next time?

joelhawksley (approved):

docs/content/components/label.md
mxriverlynn: @joelhawksley honestly, i'm a big fan of that approach! once at a job almost 20 years ago, the IT Director found a box connected to a giant cable coming out of a hole in a wall, under a folding table in a closet. literally just a hole roughly cut out of the wall... no one knew what it did, so he unplugged it. three days later, the order entry people were wondering why they had not received any new orders in three days. turns out that box connected customer orders to the order entry team. oops 🤣

docs/src/@primer/gatsby-theme-doctocat/nav.yml
mxriverlynn: @joelhawksley the current `component_status_migrator` should handle this now. i think the layout component was a leftover from before i made those changes, and i didn't think to check it. 

🤔 probably worth writing a script to verify all the component links in `nav.yml` are correct, checking against the actual components' status. this could be a good CI check for builds

PR #1293, opened 2022-08-10
Update slot use in examples for all components in the beta folder

mxriverlynn:
this PR updates all components in the `beta/` folder to use the new `.with_#{slot_name}` syntax in docs examples

it also adds the missing `status :beta` to `Beta::BorderBox::Header`

## Updated components

* Beta::AutoComplete
* Beta::AvatarStack
* Beta::BorderBox / BorderBox::Header
* Beta::BreadCrumbs
* Beta::ButtonGroup
* Beta::Flash
* Beta::Truncate

## Screenshots

auto complete:
image

avatar stack:
image

border box:
image

bread crumbs:
image

button group:
image

flash:
image

truncate:
image

### DIFF
--- a/app/components/primer/beta/border_box/header.rb
+++ b/app/components/primer/beta/border_box/header.rb
@@ -5,7 +5,7 @@ module Primer
     class BorderBox
       # `BorderBox::Header` is used inside `BorderBox` to render its header slot.
       #
-      # @accessibility When using `header.title`, set `tag` to one of `h1`, `h2`, `h3`, etc. based on what is appropriate for the page context. <%= link_to_heading_practices %>
+      # @accessibility When using `header.with_title`, set `tag` to one of `h1`, `h2`, `h3`, etc. based on what is appropriate for the page context. <%= link_to_heading_practices %>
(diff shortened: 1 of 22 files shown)


app/components/primer/beta/border_box/header.rb
```diff
@@ -5,8 +5,10 @@ module Beta
     class BorderBox
       # `BorderBox::Header` is used inside `BorderBox` to render its header slot.
       #
-      # @accessibility When using `header.title`, set `tag` to one of `h1`, `h2`, `h3`, etc. based on what is appropriate for the page context. <%= link_to_heading_practices %>
+      # @accessibility When using `header.with_title`, set `tag` to one of `h1`, `h2`, `h3`, etc. based on what is appropriate for the page context. <%= link_to_heading_practices %>
       class Header < Primer::Component
+        status :beta
```
mxriverlynn, line 10: i was previously told sub components don't need a `status`, but i found that they show up as `alpha` status in `static/statuses.json` if they don't have one specifically set. so i think we should include them

PR #1305, opened 2022-08-12
Unblock codespaces githubpreview.dev domain

keithamus:
This allows us to use the `githubpreview.dev` domain which means we can use `github.dev` with lookbook without it blocking us.


jonrohan (approved):
Nice!

Issue #1299, opened 2022-08-11
The `Primer::Alpha::Tooltip` examples should be using the component API


khiga8:
Since `Button`, `Link`, and `IconButton` all support tooltip now through it's API, the examples on the page should be updated to not have direct use of the `Primer::Alpha::Tooltip` component. One of the example has `IconButton` so we're seeing double-rendering.

PR #1306, opened 2022-08-15
Update docs to discourage standalone tooltip use.

khiga8:
Fixes: https://github.com/primer/view_components/issues/1299

Since some Primer components now support tooltip and we want to emphasize not using tooltip directly, the examples in the tooltip docs should use `Button`, `IconButton`, or `Link`. This PR swaps the examples and makes some edits to the tooltip docs.

### DIFF
--- a/.changeset/serious-fireants-tie.md
+++ b/.changeset/serious-fireants-tie.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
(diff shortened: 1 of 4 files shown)


jonrohan (approved):
I noticed we're missing the Tooltip link in the sidebar nav of the docs site. Not sure if that was intentional. Should we add it?

Either way, not related to your update so 👍🏻

Issue #1296, opened 2022-08-11
Improving screen reader support for tooltip


khiga8:
## Context

- In https://github.com/primer/view_components/pull/1256, we swapped `hidden` to `visibility: hidden`. The motivation is to workaround a surfaced incompatibility with the accessibility API and the web component API that prevents a `hidden` element from serving as an accessible name or description until the `hidden` is removed once.  It is expected for `hidden` elements to label or describe another element. This seems to be a bug specific to custom elements usage.
- `visibility:hidden` helped us get around the issue and ensured labeling/description relations are established correctly for screen reader users. However, it introduced layout issues that is apparent when tooltips are on the far rights side of the page. This is due to fact that the tooltip will take up the full space. The tooltip isn't positioned where it needs to be until it is interacted with. We don't want to position tooltip on page load because doing so will cause performance issues if a page has hundreds of tooltips.

## Suggestion

- use the `sr-only` class to hide or show the tooltip to sighted users
- always keep `aria-hidden` on the tooltip when it is a label type to reduce verbosity

PR #1307, opened 2022-08-15
Improving screen reader support for tooltip with `sr-only`

khiga8:
Fixes: https://github.com/primer/view_components/issues/1296

## Context
This is take 2 of improving screen reader support. For more on the current issue, read the description of the issue linked above.

Recently we updated our tooltips to use `visibility: hidden;` to improve tooltip screen reader support because there appears to be a bug in accessibility API and custom elements and use of `hidden` attribute (https://github.com/primer/view_components/pull/1256). We couldn't use `display: none;` either because it seemed to result in the same bug. Using `visibility: hidden;` let us achieve what we wanted but it also introduced layout bugs on some pages which lead to a revert. This is due to `visibility: hidden` taking up space.

This PR aims to improve screen reader support and avoid layout issues by using `sr-only` to hide or show the tooltip visually.  
Upon discussion with @jscholes we landed on the following:

- When tooltip is label, it will always have `aria-hidden` to avoid duplicate announcements for screen reader users.
- When tooltip is description, we will not have `aria-hidden` because duplicate announcement is acceptable for accessible descriptions to allow screen reader users to access the description text at any time.

## Changes include

Other changes include:
- greater system test coverage to test the JS behavior. the JS originally had unit tests in dotcom but we didn't import them here because PVC doesn't seem to have JS unit test setup. I added tests in system tests since I noticed we don't have much JS coverage at all.
- ensure that `aria-labelledby` set by JS respects existing `aria-labelledby` on the button.

## Reviewer

Please review the changes carefully. The thing to watch out for most are layout issues but I don't believe we should encounter it given `sr-only` should not take up space.


test/previews/primer/alpha/tooltip_preview.rb
```diff
@@ -14,6 +14,22 @@ def default(type: :description, direction: :s, tooltip_text: "Tooltip text")
...
 
+      # @param tooltip_text text
+      def label_tooltip_on_button_with_existing_labelledby(type: :label, direction: :s, tooltip_text: "Tooltip text")
+        render(Primer::ButtonComponent.new(id: "button-with-existing-label", "aria-labelledby": "existing-label-id")) do |c|
+          c.tooltip(text: tooltip_text, type: type, direction: direction)
+          "Button"
+        end
+      end
+
+      # @param tooltip_text text
```
langermank, line 25: > Please review the changes carefully. The thing to watch out for most are layout issues but I don't believe we should encounter it given sr-only should not take up space.

I wonder if we could add a mock preview for an edge case where it could potentially break layout? Maybe use the bugs from last week as a reference. There's a way to spec out some basic HTML for a preview (you can see some examples of that with Autocomplete I believe). What do you think?

khiga8: That would be ideal! I was having trouble figuring out how to have multiple elements/components within a PVC preview. I'll have to take a look at `Autocomplete`. Mind if I tag you for extra eyes once I figure it out?

langermank: @khiga8 here's an example: https://github.com/primer/view_components/tree/main/test/previews/primer/beta/auto_complete_item_preview

Yes please let me know how I can help!

khiga8: Added a preview and a test in https://github.com/primer/view_components/pull/1307/commits/162afb0ae34b0d0a53c2b9fb48c043b0a4a9abf8 ✅.

(FYI I couldn't get the background color change that we talked about over slack)

khiga8:
I pointed dotcom to these changes and verified the previously reported regression in a PR page for mobile is not present.

Previously reported regression in mobile:
Screenshot of page layout in the PR page shifted to the left leaving awkward whitespace on the right


This PR in mobile:
Screenshot of PR page in mobile


GitHub Staff can also test these changes by checking out [kh-test-view-changes](https://github.com/github/github/pull/new/kh-test-pvc-changes) (this change) or [kh-bad-changes](https://github.com/github/github/compare/kh-test-bad-change?expand=1) (for the regressed view).

test/system/alpha/tooltip_test.rb
```diff
@@ -8,8 +8,94 @@ def test_renders
...
+      refute_selector("tool-tip.sr-only", visible: :hidden)
+      assert_selector("tool-tip", visible: :visible)
+      refute_selector("tool-tip[aria-hidden]", visible: :visible)
+    end
+
+    def test_does_not_overflow_and_affect_layout
+      visit_preview(:with_right_most_position)
+      tooltip_in_viewport = evaluate_script("(function(el) {
+        return (
+          el.getBoundingClientRect().right <= (window.innerWidth || document.documentElement.clientWidth) + 1
```
khiga8, line 94: In this PVC preview environment, the button control is at the right-most side of the page and has no padding. The sr-only will give the tooltip a width of 1px causing a 1px overflow in this preview environment.

In practice however, all GitHub controls contained within a UI will have a padding of at least a few pixels so a tooltip with 1px should not cause any horizontal overflow.

I verified this 1px doesn't cause any problems in https://github.com/primer/view_components/pull/1307#issuecomment-1215997460.

(chatted about this with @langermank)

khiga8:
Hey @langermank, @primer/rails-reviewers, this is ready for review. Thank you!!

langermank (approved):
This looks good to me! I would just like for @jonrohan to review the JS changes 🚀 

Thank you for testing in previews and in dotcom!

app/components/primer/alpha/tool-tip-element.ts
```diff
@@ -176,6 +177,15 @@ class ToolTipElement extends HTMLElement {
     return this.ownerDocument.getElementById(this.htmlFor)
   }
 
+  set hiddenFromView(value: true | false) {
+    value ? this.classList.add(TOOLTIP_SR_ONLY_CLASS) : this.classList.remove(TOOLTIP_SR_ONLY_CLASS)
```
keithamus, line 181:
```suggestion
    this.classList.toggle(TOOLTIP_SR_ONLY_CLASS, value)
```

app/components/primer/alpha/tool-tip-element.ts
```diff
@@ -253,15 +263,17 @@ class ToolTipElement extends HTMLElement {
     if (name === 'id' || name === 'data-type') {
       if (!this.id || !this.control) return
       if (this.type === 'label') {
-        this.control.setAttribute('aria-labelledby', this.id)
+        let labelledBy = this.control.getAttribute('aria-labelledby')
+        labelledBy ? (labelledBy = `${labelledBy} ${this.id}`) : (labelledBy = this.id)
```
keithamus, line 267: I'd like us to avoid ternary statements, as they can be a little unwieldy.
```suggestion
        if (labelledBy) {
          labelledBy = `${labelledBy} ${this.id}`
        } else {
          labelledBy = this.id
        }
```

keithamus (approved):
This LGTM :+1: 

A couple of style notes:

PR #1319, opened 2022-08-18
[ADR] Build component CSS in primer view components with ruby code

jonrohan:
Writing up an ADR for the changes made in this PR https://github.com/primer/view_components/pull/1316

### DIFF
--- a/adr/0252-build-component-css-with-ruby-code.md
+++ b/adr/0252-build-component-css-with-ruby-code.md
@@ -1,0 +1,54 @@
+# 252. Build component css with ruby code
+
+Date: 2022-08-17
+
+## Status
+
+Accepted
+
+## Context
+
+We've talked a few times about the overhead it takes to develop components twice for rails architecture. Once in [primer/css](https://github.com/primer/css/pull/2083) and again in [primer/view_components](https://github.com/primer/view_components/pull/1225). This disconnect can lead to multiple versions and iterations happening in primer/css before any component is actually ready. [^1] [^2] [^3] [^4] [^5] [^6].
+## Decision
+
+Any new View Components should include corresponding CSS in this repository instead of in Primer CSS.
+
+### How
+
+Because this is essentially a new build system, we want to be more forward thinking with the code written here and opt for using [PostCSS] with the [preset-env] plugin that allows us to write more CSS spec features like, nested selectors, match functions, container queries etc.
+
+In our github.com production bundle we have a few options:
+
+1. We could include the compiled css file in our main bundle by adding a `@import "@primer/view-components/app/assets/styles/primer_view_components.css"` import line. This is most likely what we'll start with.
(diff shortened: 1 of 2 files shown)


adr/0252-build-component-css-with-ruby-code.md
```diff
@@ -0,0 +1,64 @@
...
+- Keep component CSS separate in primer/css
+  - Pros:
+    - Doesn't require us to change already established norms.
+    - Has storybook site
+  - Cons:
+    - Requires extra releases in both libraries when iterating on component architecture.
+    - Cognitive disassociation between Primer::ViewComponent and .PrimerCssComponent
+    - Stuck using old SCSS technology
+
+## Consequences
```
langermank, line 59: I think one consequence of this is that we're moving _away_ from potentially sharing CSS with Primer React or other frameworks by moving it into the PVC repo. 

Another one is that we'll have Primer CSS in two repos, which might be a bit confusing. Should we also discuss the impact this has on the PCSS repo? In the decisions we say "new component CSS" but we might want to mention something about "old" component CSS.

langermank: @simurai @jonrohan I love the idea of meeting with PRC to discuss. I'm going to meet with @siddharthkp this week first so we can get some thoughts down in a FigJam so that our larger discussion can be a bit more straight forward! 🎉 I wonder if we should agree upon the consequences before merging this ADR.

joelhawksley: Agreed. Let's chat about it at the React Tooling Sync tomorrow 😄

adr/0252-build-component-css-with-ruby-code.md
```diff
@@ -0,0 +1,54 @@
...
+Any new View Components should include corresponding CSS in this repository instead of in Primer CSS.
+
+### How
+
+Because this is essentially a new build system, we want to be more forward thinking with the code written here and opt for using [PostCSS] with the [preset-env] plugin that allows us to write more CSS spec features like, nested selectors, match functions, container queries etc.
+
+In our github.com production bundle we have a few options:
+
+1. We could include the compiled css file in our main bundle by adding a `@import "@primer/view-components/app/assets/styles/primer_view_components.css"` import line. This is most likely what we'll start with.
+2. We could use CSS modules to load specific bundles for each component dynamically when the component is included on the page. This is more work, but would be worth the reward.
```
siddharthkp, line 23: I have questions around rewards and tradeoff of #2, is it helpful to talk about it now or should we save that for when we get there?

mperrotti (approved):

adr/0252-build-component-css-with-ruby-code.md
jonrohan: It's probably worth another discussion for this.

PR #1323, opened 2022-08-19
Support submitting a check box group as an array

camertron:
In Rails, check boxes usually have their own names and result in sending either a `0` (false, unchecked) or `1` (true, checked) to the server on submit. For single check boxes that's probably what you want. However, Rails also supports grouping check box values into an array. If a check box is selected, Rails will add its value to an array of checked values and make that array available to your application. Users can opt into the array functionality by passing `multiple: true` to the `#check_box` form builder method. This will generate an `<input>` tag with a `name` attribute that ends in square brackets, eg. `places[]`. Rails will gather the values of all check boxes with this name into an array and stick it in the controller's `params`.

@neall discovered this is how the webhooks form under repo settings works and needs the forms framework to support it. This PR is an attempt to do so.

Passing a `name:` parameter to the check box group will now automatically cause the array behavior described above. If not given at the group level, `name:` parameters are required on each constituent check box and will be submitted individually with a value of `0` or `1` (i.e. the non-array behavior).

### DIFF
--- a/.changeset/tiny-meals-serve.md
+++ b/.changeset/tiny-meals-serve.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Support submitting a check box group as an array
--- a/lib/primer/forms/builder.rb
+++ b/lib/primer/forms/builder.rb
@@ -14,8 +14,8 @@ module Primer
         super(*args, classify(options), &block)
       end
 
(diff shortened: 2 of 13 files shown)


neall (approved):
This is great.

jonrohan (approved):

PR #1325, opened 2022-08-19
Beta::Button and Beta::IconButton visual refinements

langermank:
Using the `Beta` namespace, this PR updates style and parameters for Button components based on this [discussion](https://github.com/github/primer/discussions/1016) and design recommendations.

More refinement to come in a followup PR as we begin integrating `IconButton` into `CommentBox`, `Dialog`,  and `ActionList::Item`


jonrohan (approved):
Tested this in #1339 branch looks good 👍🏻 

image

test/lib/css_coverage_test.rb
```diff
@@ -45,8 +36,4 @@ def setup
   def test_classify_does_not_generate_primer_css_classes_that_do_not_exist
     assert_empty(@classes_from_utilities - @css_data - @allowed_missing_classes_for_now)
   end
-
-  def test_docs_do_not_generate_primer_css_classes_that_do_not_exist
-    assert_empty(@classes_from_docs_build - @css_data - @allowed_missing_classes_for_now)
-  end
```
jonrohan, line 51: This test was looking for classes being used in the docs that weren't in the primer/css bundle. It's beginning to fail more as we transition to #1319 So I'm removing that part of the test

PR #1339, opened 2022-08-25
ActionBar component

jonrohan:
https://github.com/github/primer/issues/932

Screen Shot 2022-08-30 at 16 49 35

## TODO

- [x] Hide divider when it's the first item in the overflow menu. [Details](https://github.com/primer/view_components/pull/1339#issuecomment-1231248352)
- [x] Look into alignment option [Details](https://github.com/primer/view_components/pull/1339#issuecomment-1231279420)
- [x] Do we need `ActionBar-item` class anymore now that we're using `Primer::Beta::IconButton`? https://github.com/primer/view_components/pull/1339/commits/0452d1fe453a13b7921613dd4ed40fcc6c776ba7
- [x] Focus arrows don't focus menu icon


app/components/primer/alpha/action_bar.pcss
```diff
@@ -0,0 +1,56 @@
...
+
+.ActionBar--condensed {
+  gap: 0;
+}
+
+.ActionBar--spacious {
+  gap: var(--primer-controlStack-small-gap-spacious, 16px);
+}
+
+.ActionBar--vertical {
```
simurai, line 46: Added a `direction` option. Works ok so far:

Screen Shot 2022-08-26 at 20 50 17

But yeah, it's kinda limited to `IconButton`. So we'd have to see what we wanna do for anything else.

simurai: Revered this again (for now).

simurai:
Re. dividers... I think we should **keep** the divider when in front of the "more" icon (`...`). Makes it clear that the "group of icons" is still shown in its entirety and that the "more" icon is not part of that group.

image

Just once there is no space for the divider and it's shown at the top of the `ActionMenu`, then I think we should **hide** it since it looks a bit odd:

Screen Shot 2022-08-30 at 16 15 29

simurai:
Re. left aligned. We probably will need this to be changed programmatically? E.g. when the `ActionBar` is moved to the left at runtime the alignment of the items also needs to be changed to the left.

I think we could have the `ActionBar` items always be left aligned. And then the parent can decide if the entire `ActionBar` should be aligned to the right. The extra gap in front of the "more" icon doesn't personally bother me and I think it's even a good thing because it visually separates it from the rest:

Screen Shot 2022-08-30 at 16 43 41

And another nice thing about keep it always left aligned is that when the items get hidden, the rest of the icons don't jump to the right. Here a just a video simulating it with adding `margin-left: auto` to the "more" icon.

https://user-images.githubusercontent.com/378023/187380182-293c3c38-87b4-44a8-8be1-91f8350539af.mov

Anyways, we can chat about it tomorrow.

app/components/primer/experimental/action_bar.rb
```diff
@@ -0,0 +1,53 @@
...
+    class ActionBar < Primer::Component
+      status :experimental
+
+      SIZE_DEFAULT = :medium
+      SIZE_MAPPINGS = {
+        SIZE_DEFAULT => nil,
+        :small => "ActionBar--small",
+        :large => "ActionBar--large"
+      }.freeze
+      SIZE_OPTIONS = SIZE_MAPPINGS.keys.freeze
```
simurai, line 17: Added this back to make sure there is enough spacing between items so that the touch targets don't overlap.

Screen Shot 2022-08-31 at 14 08 40

app/components/primer/beta/button.pcss
```diff
@@ -329,4 +329,19 @@ summary.Button {
   &.Button--large {
     width: var(--primer-control-large-size, 40px);
   }
+
+  /* Increase touch target */
```
simurai, line 333: This increases the touch target for `.Button--iconOnly`. There is some small risk that it will overlap something else, but maybe still ok to enable this by default.

image

/cc @langermank

app/components/primer/experimental/action_bar.pcss
```diff
@@ -0,0 +1,32 @@
...
+
+/* Divider */
+
+.ActionBar-divider {
+  margin: 0 var(--primer-control-medium-gap, 8px);
+  height: calc(var(--primer-control-medium-size, 32px) / 2);
+  border-left: var(--primer-borderWidth-thin, 1px) solid var(--color-border-subtle);
+}
+
+/* Increase spacing so touch targets don't overlap */
```
simurai, line 18: Added this back to make sure there is enough spacing between items so that the touch targets don't overlap.

Screen Shot 2022-08-31 at 14 08 40

app/components/primer/beta/button.pcss
langermank: Yeah, I haven't thoroughly tested this yet. We merged it a bit hastily 😅 we can write some previews to try and simulate how it would behave.

app/components/primer/experimental/action-menu-element.ts
```diff
@@ -0,0 +1,387 @@
+// eslint-disable-next-line prettier/prettier
```
jonrohan, line 1: `TODO: Remove this file before merging`

app/components/primer/beta/button.pcss
```diff
@@ -329,4 +329,19 @@ summary.Button {
...
+  @media (pointer: coarse) {
```
langermank, line 334: @jonrohan I think this is already being accomplished at line 32 with a mixin. You can change the values to 44px instead 😄

PR #1340, opened 2022-08-28
Let forks ✅ on `main`

### LINKED ISSUE #1033: docs-build fails on forks
https://github.com/primer/view_components/blob/439cc7629fd650454dac3607036af6957cef295a/.github/workflows/docs-build.yml#L17

https://github.com/jsoref/view_components/runs/5349636877?check_suite_focus=true
```
Run actions/checkout@v[2](https://github.com/jsoref/view_components/runs/5349636877?check_suite_focus=true#step:2:2)
  with:
    repository: jsoref/view_components
    ssh-strict: true
    persist-credentials: true
    clean: true
    fetch-depth: 1
    lfs: false
    submodules: false
Error: Input required and not supplied: token
```

Most workflows in this repository are fairly nice. I've seen some primer repositories with workflows that check to see if the workflow is running in the primer organization / specific primer repository.

It's also possible to check for a secret, and if it isn't present skip things, instead of failing.

jsoref:
closes #1033

https://github.com/jsoref/view_components/actions?query=branch%3Amain

### Before
image

### After
image

(Note that I've disabled CI because it's expensive, but I'm not changing anything relevant to CI.

The change to `@return` is included to show that there's a little value in not just skipping the entire workflow. Note that I have no particular preference between skipping an entire workflow or skipping a portion of a workflow.


jonrohan (approved):
Thanks!

PR #1350, opened 2022-09-06
Responsive Component subsystem

krhkt:
This PR adds a Responsive Component abstraction that can be used to create View Components in a more robust way, avoid the current boilerplate.

For more information, read the following ADR: _this is still incomplete_
[0252-responsive-component-base.md](https://github.com/primer/view_components/blob/d0ebf5c3bc5344412a66a0cfda097f6e1f335dce/adr/0252-responsive-component-base.md)


krhkt:
quick update: my FR rotation was really busy (and in P&T it lasts 3 weeks)

We had an incident that's stil ongoing:
- https://github.com/github/memex/issues/12669

Tonight I'll have time to resume work on this:
- fix linting
- fix tests
- wrap up documentation

lesliecdubs:
@jonrohan @camertron could you both take a look at this and weigh in before too many more conflicts stack up?

camertron:
Hey @krhkt, thanks for resurrecting this PR 😄 

I'm conflicted here. It looks to me as though you've put in _a lot_ of work on this, and I want to respect that. My reservations are three-fold:

1. It's been long enough since we talked in-person about this that I don't think I understand the code well enough to properly review it.
2. I'm concerned about the ongoing maintenance burden a complex feature like this will put on our (small) team. Ultimately we're responsible for maintaining, documenting, and testing the code that makes it into the library on an ongoing basis, and I don't feel confident anyone on our team can do that for this contribution.
3. I don't remember if we discussed this in-person, but to my knowledge responsive functionality isn't something I hear teams at GitHub asking for. That's not to say it's a good or bad idea, only that in our day-to-day component authoring work I personally don't see the need this addresses.

With that said, let me propose an alternative: what about creating a separate gem for the responsive component subsystem? That way our team wouldn't have to shoulder the maintenance burden. If the subsystem becomes well-adopted at GitHub, we can consider moving the project under the Primer umbrella, eventually merging it into primer/view_components (PVC) when we feel the time is right. Most of the code that makes it in is upstreamed from dotcom, so we'd be following a similar pattern here.

Thoughts?

krhkt:
@camertron hi, thank you for replying. Sorry the late response on my part.

> With that said, let me propose an alternative: what about creating a separate gem for the responsive component subsystem? That way our team wouldn't have to shoulder the maintenance burden. If the subsystem becomes well-adopted at GitHub, we can consider moving the project under the Primer umbrella, eventually merging it into primer/view_components (PVC) when we feel the time is right. Most of the code that makes it in is upstreamed from dotcom, so we'd be following a similar pattern here.

Sounds like a solid plan and makes thing way easier. I just don't have enough bandwidth currently to work that out.

> I don't remember if we discussed this in-person, but to my knowledge responsive functionality isn't something I hear teams at GitHub asking for.

Yes, I think the discussion around that has died down and I don't even know if that responsive ADR is going to be followed at all. So I think it's better to just close this PR.

Thank you for the time helping me with this. =]

PR #1357, opened 2022-09-07
Setup preview deploys for pull requests

jonrohan:
This setups up preview deploys for the current `/demo/` app, which is only storybook.

The script runs by creating a new container instance on azure with the PR number in it. This was heavily inspired by the workflow in https://github.com/github/docs/blob/main/.github/workflows/azure-preview-env-deploy.yml

image

### Follow up todo

I'd like to follow up this PR with the following

- [x] Separate workflow to cleanup unused deployments https://github.com/github/docs/blob/main/.github/workflows/azure-preview-env-destroy.yml
- [ ] Combine `/demo/` and `/lookbook/` servers into one server https://github.com/primer/view_components/pull/1347
- [ ] Migrate all stories to `/loobook/` and delete storybook installation

---

@peterbe @mikesurowiec Would you mind reviewing this PR to let me know if I need anything else?


mikesurowiec:
Nice! I'd also recommend adding a "destroy" workflow, similar to this: https://github.com/github/docs-internal/blob/main/.github/workflows/azure-preview-env-destroy.yml

It removes the docker images, the running container instance, and the github "deployment" (https://github.com/github/docs-internal/deployments) for that environment.

.github/workflows/deploy-demo-preview.yml
```diff
@@ -0,0 +1,92 @@
...
+      # This saves time because the PR can merge sooner.
+      if: ${{ !github.event.pull_request.auto_merge }}
+      uses: azure/arm-deploy@841b12551939c88af8f6df767c24c38a5620fd0d
+      with:
+        resourceGroupName: primer
+        subscriptionId: 550eb99d-d0c7-4651-a337-f53fa6520c4f
+        template: ./.github/workflows/deploy-demo-preview-template.json
+        deploymentName: ${{env.DEPLOYMENT_NAME}}
+        parameters: appName="${{env.APP_NAME}}"
+          containerImage="primer.azurecr.io/${{ env.IMAGE_REPO }}:latest"
```
mikesurowiec, line 96: I think this should be 
```suggestion
          containerImage="${{ DOCKER_IMAGE }}"
```

.github/workflows/demo-preview-deploy.yml
```diff
@@ -0,0 +1,101 @@
+
+
+name: Demo Preview
+
+on:
+  pull_request:
```
peterbe, line 6: Shouldn't this be ?
```suggestion
  pull_request_target:
```

.github/workflows/demo-preview-destroy.yml
```diff
@@ -0,0 +1,46 @@
+name: Demo Preview
+
+on:
+  pull_request:
```
peterbe, line 4: ditto

peterbe:
Is http://primer-view-components-preview-1357.eastus.azurecontainer.io/ and https://primer-020b8eb3dd-26611710.drafts.github.io/ as you'd expect it? If so, r=me

.github/workflows/demo-preview-deploy.yml
jonrohan: I think we want to restrict it running on forks, so keeping it as pull_request only is what I'd prefer for now, might change this in the future though

PR #1365, opened 2022-09-12
Convert the rest of the primer storybook stories to view component previews

jonrohan:
This PR converts the rest of the `stories/` to ViewComponent previews to be used in lookbook.

image


joelhawksley (commented):
These generally look good. Won't removing the existing stories break the docs, which link to storybook for each component?

jonrohan:
> These generally look good. Won't removing the existing stories break the docs, which link to storybook for each component?

We could remove the links for now

joelhawksley:
> We could remove the links for now

Hmm. I feel like that would be a regression in our docs experience. How about we _add_ the Lookbook previews, update the docs to point to Lookook, remove the links to Storybook, _then_ remove the stories?

PR #1374, opened 2022-09-14
Demo lookbook

jonrohan:
Finally getting review deployments working http://primer-view-components-preview-1374.eastus.azurecontainer.io/view-components/lookbook/

### DIFF
--- a/demo/config/routes.rb
+++ b/demo/config/routes.rb
@@ -12,6 +12,8 @@ Rails.application.routes.draw do
     get '/auto_complete', to: 'auto_complete_test#index'
     resources :toggle_switch, only: [:create]
 
-    mount Lookbook::Engine, at: "/lookbook"
+    unless Rails.version.to_i >= 7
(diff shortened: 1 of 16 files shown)


demo/config/routes.rb
```diff
@@ -1,9 +1,18 @@
 # frozen_string_literal: true
 
+# :nocov:
 Rails.application.routes.draw do
-  get '/view-components', to: redirect('/view-components/stories/')
-  get '/', to: redirect('/view-components/stories/')
+  if Rails.env.production?
+    get '/view-components', to: redirect('/view-components/stories/')
```
joelhawksley, line 6: Does this mean that this route is only available in prod now? I think we'd want to have it available in development still.

jonrohan: The storybook server runs separately at :5000. This doesn't effect that

joelhawksley (approved):

PR #1375, opened 2022-09-15
Deleting the old lookbook folder

jonrohan:
This is a follow up to https://github.com/primer/view_components/pull/1374

I'm deleting the old `/lookbook/` folder to reduce development into the `/demo/` folder.

I moved some of the config over from the old folder to make development a bit nicer.


langermank (approved):
🚀

PR #1376, opened 2022-09-15
Adding lookbook links to docs output

jonrohan:
image

- The links will only show up if the component has a preview written.
- I had to suffix the url with `/default/` which assumes all components have a `def default` preview. I did this because leaving it off rails ends up redirecting to the azure server url example https://primer.style/view-components/lookbook/inspect/primer/beta/avatar/

### DIFF
--- a/lib/tasks/docs.rake
+++ b/lib/tasks/docs.rake
@@ -131,7 +131,6 @@ namespace :docs do
         f.puts("componentId: #{data[:component_id]}")
         f.puts("status: #{data[:status]}")
         f.puts("source: #{data[:source]}")
-        f.puts("storybook: #{data[:storybook]}")
         f.puts("lookbook: #{data[:lookbook]}") if preview_exists?(component)
         f.puts("---")
         f.puts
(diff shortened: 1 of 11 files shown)


lib/tasks/docs.rake
```diff
@@ -485,6 +487,18 @@ namespace :docs do
     "https://primer.style/view-components/stories/?path=/story/#{path}"
   end
 
+  def lookbook_url(component)
+    path = component.name.underscore
+
+    "https://primer.style/view-components/lookbook/inspect/#{path}/default/"
```
langermank, line 493: Does this mean that there needs to be a preview called "default"?

jonrohan: I don't mind what we call it, but we do need to make sure they all have the same method

langermank: I started an ADR for preview/story format and recommend that every component has a "default" so I like this 😄

langermank (approved):

lib/tasks/docs.rake
joelhawksley: @jonrohan perhaps we could add a linter or test that makes sure each component has a preview called `default`, then?

PR #1377, opened 2022-09-15
Run axe scanner on written previews in test/previews

jonrohan:
I noticed we're running axe scan on the previews generated from our docs, but not on the manually written `test/previews` 

This PR turns scanning on for those previews, and fixes any currently failing previews


khiga8 (approved):
🚀

PR #1378, opened 2022-09-15
Separate run for slow tests

jonrohan:
Separating out faster component tests from the system tests.


.github/workflows/ci.yml
```diff
@@ -129,3 +128,25 @@ jobs:
         bundle config path vendor/bundle
         bundle install --jobs 4 --retry 3
         bundle exec rake coverage:report
+  system:
+    name: Accessibility
```
khiga8, line 132: Should we just generally call this `System` or `Browser`? Though the a11y test, `doc_examples_axe_test` is taking the most time, there's more general component tests as well for testing behavior.

[system test folder](https://github.com/primer/view_components/tree/main/test/system)

khiga8 (approved):
just one comment! great idea!

PR #1379, opened 2022-09-16
Run deprecated job as a test instead of ci job

jonrohan:


test/components/component_test.rb
```diff
@@ -162,4 +163,11 @@ def render_component(component, args, proc)
       proc.call(c) if proc.present?
     end
   end
+
+  def test_deprecated_components_by_status_match_list
+    deprecated_by_status = Primer::ViewComponents::STATUSES.select { |_, value| value == "deprecated" }.keys.sort
+    deprecated_by_list = ::Primer::Deprecations::DEPRECATED_COMPONENTS.keys.sort
+
+    assert_empty(deprecated_by_status - deprecated_by_list, "Please make sure that components are officially deprecated by setting the `status :deprecated` within the component file.\n")
```
khiga8, line 171: Can we also include:

"Make sure to provide an alternative component for each deprecated component in `Primer::Deprecations::DEPRECATED_COMPONENTS` (lib/primer/deprecations.rb). If there is no alternative to suggest, set the value to nil."

or some version of it for extra clarity?

PR #1381, opened 2022-09-16
Simplify CI configuration to a single build per Ruby/Rails version.

joelhawksley:
Our existing CI configuration runs a multi-build matrix for the main test suite. I think this is overkill for our needs. In my experience, our matrix builds have mainly served us by testing each version of Rails and Ruby we support, not every combination of Ruby and Rails we support. I think we can get by with just 3 builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rails_version: [6.1.1, 7.0.3, main]
-        ruby_version: [2.7.x, 3.0.x, 3.1.x]
+        include:
+          - rails_version: "6.1.1"
+            ruby_version: "2.7.x"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -13,8 +13,13 @@ jobs:
...
       matrix:
-        rails_version: [6.1.1, 7.0.3, main]
-        ruby_version: [2.7.x, 3.0.x, 3.1.x]
+        include:
+          - rails_version: "6.1.1"
+            ruby_version: "2.7.x"
+          - rails_version: "7.0.3"
+            ruby_version: "3.0.x"
+          - rails_version: "main"
+            ruby_version: "3.1.x"
```
jonrohan, line 22: We have a second matrix for the `test` section

jonrohan (approved):
Great improvement 👍🏻

PR #1383, opened 2022-09-17
Remove storybook from the repo

jonrohan:
Now that we're deploying lookbook, https://primer.style/view-components/lookbook/ it doesn't seem that we need to keep storybook around.

This PR removes the code for storybook. An added benefit from doing this is a lot of the builds and deploys are faster.


joelhawksley (approved):
👏🏻

PR #1386, opened 2022-09-19
Make sure duplicate ID associations aren't created

khiga8:
I observed that some buttons labelled or described by a tooltip can end up with multiple duplicated `aria-labelledby` and `aria-describedby` association resulting in duplicate screen reader announcements. The button attribute can end up like:

```
aria-labelledby="tooltip-1663617441252-3721 tooltip-1663617441252-3721 tooltip-1663617441252-3721"
```

which is definitely unintentional/ an annoying bug.

I observed this particularly on pages with soft navigation where `attributeChangedCallback` can end up getting triggered multiple times for `name === 'id' || name === 'data-type'` attributes.

We want to both make sure that `aria-labelledby` or `aria-describedby` includes any IDs that a user may have set in addition to the IDs that our tooltip JS provides. At the same time, we want to ensure that IDs don't get set multiple times. Our current logic is resulting in the same IDs being appended in some scenarios while trying to accomplish the former.

This PR introduces a conditional check which will ensure that IDs don't get duplicated. I've also added a test that mimics the `attributeChangedCallback` triggering. 

I'd like to get this out soon since it's likely very annoying for screen reader users. 🙏

### DIFF
--- a/test/system/alpha/tooltip_test.rb
+++ b/test/system/alpha/tooltip_test.rb
@@ -66,6 +66,42 @@ module Alpha
       assert_equal("existing-description-id #{tooltip_id}", find("button")["aria-describedby"])
     end
 
+    def test_does_not_render_duplicate_describedby_id_when_attribute_callback_triggered
+      visit_preview(:description_tooltip_on_button_with_existing_describedby)
+
+      tooltip_id = find("tool-tip", visible: :hidden)["id"]
+      assert_equal("existing-description-id #{tooltip_id}", find("button")["aria-describedby"])
+      evaluate_script("(function(el) {
+        return (
+          el.setAttribute('data-type', 'label')
+        );
+      })(arguments[0]);", find("tool-tip", visible: :hidden))
+      evaluate_script("(function(el) {
+        return (
+          el.setAttribute('data-type', 'description')
+        );
+      })(arguments[0]);", find("tool-tip", visible: :hidden))
+      assert_equal("existing-description-id #{tooltip_id}", find("button")["aria-describedby"])
+    end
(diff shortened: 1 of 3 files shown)


test/system/alpha/tooltip_test.rb
```diff
@@ -66,6 +66,42 @@ def test_appends_to_existing_aria_describedby
...
+      evaluate_script("(function(el) {
+        return (
+          el.setAttribute('data-type', 'label')
+        );
+      })(arguments[0]);", find("tool-tip", visible: :hidden))
+      evaluate_script("(function(el) {
+        return (
+          el.setAttribute('data-type', 'description')
+        );
+      })(arguments[0]);", find("tool-tip", visible: :hidden))
```
khiga8, line 83: Mimics an `attributeChangedCallback` for `data-type` .

test/system/alpha/tooltip_test.rb
```diff
@@ -66,6 +66,42 @@ def test_appends_to_existing_aria_describedby
...
+      evaluate_script("(function(el) {
+        return (
+          el.setAttribute('data-type', 'label')
+        );
+      })(arguments[0]);", find("tool-tip", visible: :hidden))
+      assert_equal("existing-label-id #{tooltip_id}", find("button")["aria-labelledby"])
+    end
+
```
khiga8, line 104: These same tests will fail in `main` with an ID output like:

`"existing-description-id tooltip-1663619871156-8889 tooltip-1663619871156-8889"`

jonrohan (approved):
Nice catch

PR #1392, opened 2022-09-20
Converting arguments.yml to arguments.json

jonrohan:
This converts the arguments.yml file we've been generating since to json file instead. The reason is because of the new docs work next quarter, it will be more easily processed by gatsby in json form.


langermank (approved):
Thank you!

PR #1393, opened 2022-09-20
Add the Requested-With header to toggle switch requests

camertron:
This is apparently something libraries like jQuery do to inform the server the request was made via AJAX. It's something dotcom expects as well.


joelhawksley (commented):
Care to provide a test?

jonrohan (approved):

PR #1395, opened 2022-09-21
Add accessible column in component statuses

josepmartins:
- [x] Add an accessible column on the component statuses page.
- [x] Update `docs.rake` to add a11yReviewed to the frontmatter

Tracking issue: https://github.com/github/primer/issues/1212
Related: https://github.com/primer/react/pull/2258


lib/tasks/docs.rake
```diff
@@ -131,6 +131,7 @@ namespace :docs do
         f.puts("componentId: #{data[:component_id]}")
         f.puts("status: #{data[:status]}")
         f.puts("source: #{data[:source]}")
+        f.puts("a11yReviewed: #{false}")
```
josepmartins, line 134: I temporarily used this hardcoded value for testing reasons. 

For context, there is no component at the moment with the `a11yReviewed` variable set true (so not reviewed by the accessibility team). 

Any idea how to handle the `a11yReviewed` variable for all components? any help here will be much appreciated ❤️  @jonrohan @camertron

camertron: Hmm I'm not 100% sure, but I believe anything in the beta or stable categories should have had an accessibility review as per the [lifecycle criteria](https://primer.style/contribute/component-lifecycle). There may be other components that have also gone through an a11y review but are not yet in either of those categories. I would reach out to the a11y team for guidance. It would be great to have this data in PVC 😄

camertron (approved):
This looks good to me pending getting the actual data into the frontmatter 😄

lib/tasks/docs.rake
josepmartins: @camertron It should be true for new components, but at the moment the accessibility team has reviewed 2 components in the react implementation, and the idea with these labels is to temporarily give more visibility to that until we reach the lifecycle criteria. I'm working closely with the a11y team on this btw 👍 https://github.com/github/primer/issues/1212

Any idea what would be a good starting point to create a basic infrastructure for the a11y or the design team to update this status on a PVC component?

camertron: Actually, I just found static/audited_at.json that's supposed to contain a mapping from component -> date of a11y audit, but unfortunately all the date values are empty strings 😞 

However, this discovery also uncovered that there's a mechanism for marking components as a11y reviewed, which I didn't know existed! Here's how it works:

```ruby
class MyComponent < Primer::Component
  include Primer::Audited::Dsl
  audited_at 'YYYY-MM-DD'
end
```

So that's where I'd put the data. Then you can run `rake static:dump` to generate static/audited_at.json, or pull the dates straight from the component classes themselves by doing something like

```ruby
Primer::Component.descendants.each do |component|
  component.audited_at  # do something with this
end
```

josepmartins: @camertron thank you! didn't know about the `audited_at`. I moved on by pulling the dates straight from the component class, so from now on it should be as easy as adding the `audited_at` value in the class as we do with the status:

```ruby
    class Avatar < Primer::Component
      status :beta
      audited_at "2022-04-11"
```

PR #1396, opened 2022-09-21
Deprecate Primer::ButtonComponent in favor of Primer::Beta::Button

mxriverlynn:
with the CI changes that have been made in the dotcom repo, we can safely deprecate components in PVC without requiring an all-or-nothing update in dotcom.

this PR marks the `Primer::ButtonComponent` as deprecated in favor of `Primer::Beta::Button`. it also corrects the primer.style navigation so the `Button` link in the nav takes you to the beta button instead of the deprecated version.

image

### DIFF
--- a/.changeset/hot-walls-arrive.md
+++ b/.changeset/hot-walls-arrive.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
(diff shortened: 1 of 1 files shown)


jonrohan (approved):
Woo hoo 🥇

PR #1400, opened 2022-09-22
Add Lookbook previews for Beta Button and IconButton

langermank:
Previews!


jonrohan:
Nice! While you're here, what do you think about deleting the preview for the old IconButton, I think it's `test/previews/primer/icon_button_preview.rb`

jonrohan (approved):

PR #1402, opened 2022-09-23
move Octicon cache preload to initializer

joelhawksley:
I believe we can move this code to be with the other compilation optimizations and not at the component level. This will allow us to remove `_after_compile` from ViewComponent.


jonrohan:
I don't know enough about the octicon cache helper to make a judgement here.

camertron (approved):
Nice, this makes sense to me 👍

PR #1404, opened 2022-09-26
Move previews into root and include in gemspec

steves:
I paired on this with @joelhawksley. We've begun including Lookbook in `github/github`. If we include our primer/view_components previews in the gem then we can pull them into the local Lookbook instance. After some debate we also decided to move them out of `test` and into their own top level directory. It felt wrong including test files in the gem and previews really aren't tests. They can be used by tests but they are, themselves, not tests.


jonrohan (approved):
Love it 👍🏻

camertron (approved):
Sweet!

PR #1407, opened 2022-09-26
Automatically publish the package when a GitHub release is published

jonrohan:
This is based on how `octicons` currently publishes. https://github.com/primer/octicons/pull/681

The `tl;dr;` is that everything currently stays how it is.

1. Changeset creates release PR
2. Release PR is merged
3. Changeset creates GitHub release

Previously after that last step, we would need to manually run `script/publish`.

This PR setups a workflow to listen for the `release: published` event and then prepares and then runs `script/publish`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,16 +11,33 @@ jobs:
         uses: actions/setup-ruby@v1
         with:
           ruby-version: "3.1.x"
-      - uses: actions/setup-node@v2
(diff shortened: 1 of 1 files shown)


camertron (approved):
Oh heck yes! 🎉

jonrohan:
I'm fairly sure this will work, but won't know until it's used so going to merge and we'll revisit if the next release experiences problems

PR #1412, opened 2022-09-27
Change header to div on dialog

owenniblock:
### Motivation

When the dialog appears outside of a sectioning element (like `<main>`) it looks like some browser add the `banner` role to the element. This behaviour is undesired so we're changing the header to a div.

Looks like the `<footer>` has already been removed.


keithamus (approved):
LGTM

PR #1417, opened 2022-09-28
Introducing Stylelint to lint component *.pcss files

jonrohan:
This PR introduces [Stylelint](https://stylelint.io/) to run against our component `*.pcss` files. I'm creating a new config with two of the standard config settings. `"stylelint-config-standard", "stylelint-config-prettier"`.

**Why a new config?** Why a new config, if we already have [@primer/stylelint-config](https://github.com/primer/stylelint-config)? 

That config is specific to SCSS and contains a lot of SCSS rules that would need to be inspected in order to work.


langermank (approved):
Looks good in Lookbook 👍

PR #1419, opened 2022-09-28
Add VS Code fileNesting patterns for view_components

jonrohan:
This adds the file nesting patterns to the vscode settings to combine view component files into matching. [File Nesting information](https://code.visualstudio.com/updates/v1_67#_explorer-file-nesting)

**Before**
image

**After**
image

However, I left out turning it on/off so the user can decide if they want to have this feature on.

image


langermank (approved):
OH HECK YEAH

PR #1421, opened 2022-09-29
Fix Beta::Button visual regressions

langermank:
- Adjusts `:invisible` param to match more closely with original
- Add back proper "link" styles for buttons pretending to be links
- Add notes for upgrading to beta

several invisible buttons in a row with different params set


simurai (approved):
I _think_ this will affect the `CommentBox`, but we can address that separately. Added a note to: https://github.com/github/primer/discussions/1313#discussioncomment-3760520

/cc @jonrohan

jonrohan (approved):
Nice 👍🏻

PR #1422, opened 2022-09-29
Add SVG and custom content options to NavList's leading visual slot

camertron:
This PR adds SVG and custom content as options to NavList's leading visual slot.

### DIFF
--- a/app/components/primer/alpha/action_list/item.rb
+++ b/app/components/primer/alpha/action_list/item.rb
@@ -32,14 +32,24 @@ module Primer
         # for layout options.
         renders_one :description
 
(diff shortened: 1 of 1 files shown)


langermank (approved):

joelhawksley (commented):
Looks good. Mind adding tests and a preview that shows this feature?

camertron:
@joelhawksley I don't think tests are necessary since they'd only be exercising polymorphic slots and `BaseComponent`, which are already well-covered. I'd be happy to add some previews though.

jonrohan (approved):

PR #1424, opened 2022-09-29
Allow specifying an unchecked value for check box inputs

camertron:
Right now, the check box input doesn't allow specifying an unchecked value. This PR adds that option.


jonrohan (approved):
This seems ok to me, not sure if you also want @neall to weigh in

camertron:
@jonrohan as a matter of fact, @neall and I paired on this, so that's tacit approval I'd say.

PR #1428, opened 2022-09-30
Remove external classes from the the button content `<span>` in `Primer::beta::Button`

bolonio:
## Context

We (the accessibility team) are using the `Primer::beta::Button` for the QueryBuilder component. We’ve seen that the prop `classes` is applied both to the `<button>` element and the `<span>` element inside the `<button>` element.

## Expected behavior

The prop `classes` should only be to the `<button>` element and not to the `<span>` element inside the `<button>` element.


jonrohan (approved):
Thanks for this

PR #1432, opened 2022-09-30
Add a labeler workflow that will categorize our pulls

jonrohan:
This should add labels to incoming prs for the types of files that they change. To help us get a sense of what a PR is updating.


camertron (approved):
Wow, this is really cool 😎

PR #1437, opened 2022-10-03
Save component snapshots in test/snapshot folder

jonrohan:


.github/workflows/test.yml
```diff
@@ -0,0 +1,82 @@
...
+        cache-dependency-path: |
+          package-lock.json
+          demo/package-lock.json
+    - run: |
+        npm ci
+        cd demo && npm ci
+    - name: Build and test with Rake
+      run: |
+        bundle exec rake test:snapshots
+    - id: auto-commit
```
rezrah, line 42: This is super interesting because you're not relying on maintainers to update snapshots manually. Is that correct? All of it is done on the server, which eliminates the need for additional Docker config too. 

I guess the trade-off here is that folks may just ignore / forget to look at the visual differences and scroll past it over time? Keen to see how this plays out either way. I like this approach.

rezrah (approved):
Nice work @jonrohan. Very cool! 🙌

.github/workflows/test.yml
```diff
@@ -0,0 +1,82 @@
...
+    - name: "Changes detected"
+      if: steps.auto-commit.outputs.changes_detected == 'true'
+      uses: phulsechinmay/rewritable-pr-comment@v0.2.1
+      with:
+        message: |
+          ### ⚠️ Visual differences found
+
+          Our visual comparison tests found UI differences. Please review the differences by viewing the files changed tab to ensure that the changes were intentional.
+
+          [Review visual differences](https://github.com/primer/view_components/pull/${{ github.event.number }}/files?file-filters%5B%5D=.png&show-viewed-files=false)
```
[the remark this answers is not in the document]
jonrohan: I linked to the PR tab because we do have those features in GitHub. If there is a change in the image you'll see the swipe and onion skin. You can see it in action on this commit where some images changed. https://github.com/primer/view_components/pull/1437/commits/32002d5aecdfb644e48a21645e0b79e7372627a4

.github/workflows/test.yml
jonrohan: I think a benefit here too is that we'll ensure the images are always run on the same system, so if there's subtle differences in font rendering it won't show.

PR #1438, opened 2022-10-03
ActionListItem `line-height` bug fix

langermank:
- Adds `line-height` declarations to all CSS that sets a `font-size`


camertron (approved):
Thanks :)

PR #1443, opened 2022-10-04
Renaming benchmark ci run as Test > Performance and improving build times

jonrohan:
In response to a feeling that our `test/` folder could use some organization I strarted looking up testing schemes and came across [Rails Test Types](https://blog.railwaymen.org/rails-tests-ruby-on-rails-apps) and [The Testing Pyramid](https://semaphoreci.com/blog/testing-pyramid).

Both seem like sane methodologies, I'm not trying to advocate for one or the other, but it would be nice to have some kinda guidance on them.

However for this PR, it seemed sensible to create a `Test > Performance` test since we were already doing that and calling it benchmarks. The rename makes it clearer for contributors, and allows us to include different kinds of Performance tests.

This PR doesn't change anything that's currently being tested.


camertron (approved):
I can dig it

PR #1446, opened 2022-10-04
Allow NavLists to define sub-items as well as a trailing visual

camertron:
The current implementation of `NavList` disallows attaching a trailing visual to items that also define sub items. This used to be because the trailing visual slot was used for the expand/collapse chevron icon that shows/hides the list of sub items. However, the component uses the `private_trailing_action_icon` slot for the chevron now, and should no longer raise an error if sub items are present and the `trailing_visual` slot has been provided. I believe it should instead raise an error if sub items are present and the `trailing_action` slot is provided, since it doesn't make sense to render two trailing actions.

### DIFF
--- a/.changeset/long-boats-remain.md
+++ b/.changeset/long-boats-remain.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Allow NavLists to define sub-items as well as a trailing visual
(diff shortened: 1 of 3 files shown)


jonrohan (approved):

langermank (approved):
Yay tests!

PR #1447, opened 2022-10-04
Also run forms tests

jonrohan:
I noticed we were missing running any `test/forms/` tests

### DIFF
--- a/lib/tasks/test.rake
+++ b/lib/tasks/test.rake
@@ -22,6 +22,7 @@ namespace :test do
     t.test_files = FileList[
       "test/components/**/*_test.rb",
       "test/lib/**/*_test.rb",
+      "test/forms/**/*_test.rb",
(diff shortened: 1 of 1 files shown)


joelhawksley (commented):
What do you think about adding a SimpleCov change first, making it so that this change is necessary to make CI pass?

jonrohan:
> What do you think about adding a SimpleCov change first, making it so that this change is necessary to make CI pass?

I'm not sure how to do that?

camertron (approved):
Oh crap, my bad! Thank you!

joelhawksley:
> > What do you think about adding a SimpleCov change first, making it so that this change is necessary to make CI pass?
> 
> I'm not sure how to do that?

I'm not sure either. @camertron any idea why our coverage checks didn't work here?

PR #1449, opened 2022-10-04
Actions snapshots

jonrohan:
## Description

First thing, this adds snapshots for `:focus` state and whatever happens right after pressing enter on the first focus item. I accomplish this by sending a keyboard event for `tab` then snapshot whatever has the current focus. I then send another keyboard event for `enter` and take another snapshot.

A bit of re-organizing of the file structure. I wanted to be able to accomidate more states and color themes in the future, so I moved the folder structure a bit to better account for these. Here's an example of how that might look.

```
test/snapshots/primer/
├─ component1/
│  ├─ preview_name/
│  │  ├─ light/
│  │  │  ├─ initial.png
│  │  │  ├─ focus.png
│  │  │  ├─ enter.png
```


camertron (approved):
Wow ok cool! So the idea is to make sure every component supports being focused?

PR #1452, opened 2022-10-05
Move sample forms out of test folder

steves:
In #1404 preview files were moved out of `test` and into their own top level directory. We later discovered that form previews reference form classes that are defined in `test/forms`. Because they're in the `test` folder they don't get included in the gem and the previews that use them don't work.

Because these sample forms are only used in the previews, and not in any tests that I could find, I moved them into `app/forms`. This lets us autoload them in the local Lookbook as well as bundle them into the gem for use in our `github/github` Lookbook instance.


camertron:
Hey @steves, thanks for taking this on. I don't remember if this is the case for all of them, but I know some of those forms are indeed used in tests. As long as tests are passing though we should be ok.

camertron (approved):

jonrohan (approved):

PR #1454, opened 2022-10-05
Don't build snapshots on release

jonrohan:
I'm doing this because the new Snapshots build is running on the Release PRs https://github.com/primer/view_components/pull/1444/commits/8faf01e282730d78174273b8b82449ed97938568

This tells it not to run when the actor is `primer-css`


.github/workflows/demo-preview-destroy.yml
```diff
@@ -13,7 +13,7 @@ concurrency:
 
 jobs:
   destory:
-    if: ${{ github.event.pull_request.head.repo.full_name == 'primer/view_components' && github.actor != 'dependabot[bot]' && github.actor != 'primer-css' }}
+    if: ${{ github.event.pull_request.head.repo.full_name == 'primer/view_components' && github.actor != 'dependabot[bot]' }}
```
jonrohan, line 16: realized this shouldn't be here or else we're never deleting release pr deployments

PR #1457, opened 2022-10-06
Remove dead Flex and FlexItem links from nav

orhantoy:
primer-dead-links

These components were deprecated in https://github.com/primer/view_components/pull/511 and then removed some time later on.


jonrohan (approved):
🤦🏻‍♂️ Thanks @orhantoy! Totally missed that

PR #1460, opened 2022-10-06
Remove workflow that updates status project board

jonrohan:
This removes the workflow that was introduced in https://github.com/primer/view_components/pull/373 to update this project https://github.com/primer/view_components/projects/3 with cards representing component statuses.

I asked in https://github.slack.com/archives/CGJTTJ738/p1664838618429209 if it was useful, and didn't get a response.


camertron (approved):
Aw dang, this looks like it was a lot of work 😓

PR #1465, opened 2022-10-07
Apply browser options to driver in ApplicationSystemTestCase

erinnachen:
I was poking around in this file while investigating swapping out the cuprite driver for selenium in order to be able to bump axe-core-api to v4.4.2. It looks like the intent of `Ferrum::Browser.new(process_timeout: 240, timeout: 240)` was to apply these options to the system tests.


jonrohan (approved):
🤘🏻 This is probably why we've been getting timeouts

PR #1470, opened 2022-10-10
Run script with bash

koddsson:
`pushd` and `popd` exist in `bash` but not in `sh`. `script/dev` therefor errors on my codespace but this patch fixes that issue.


jonrohan (approved):
Thanks!

Issue #1366, opened 2022-09-13
Nested dialogs are dismissed immediately


koddsson:
The global click handler for dialogs doesn't differentiate what dialog is "at the top", so any click on the top dialog dismisses any dialog "below" if that click is outside that dialogs area.

A concrete example is a Dialog with `position: :right`, which includes a nested Dialog with default settings. Any clicks to the nested dialog will register as "outside" of the root dialog and close it.

lesliecdubs:
Thanks for opening this up, @koddsson!

I'm curious about the use case for nested dialogs, and whether that's a design pattern we want to support. @langermank I know you worked on defining the [Dialog interface guidelines](https://primer.style/design/components/dialog); do you have any thoughts on this?

koddsson:
@lesliecdubs; Sure! I think nested dialogs are a bit 😬 but I was given a design for `github/github` which uses a side panel which includes a trigger for a more conventional dialog. I don't wanna give too much away in public since this is a unreleased feature but I can share additional info with you or the team in Slack.

PR #1471, opened 2022-10-10
Nested dialogs

koddsson:
Make sure that nested dialogs work correctly.

I also removed some dead code, added a preview for nested dialog and made sure `script/dev` exists if `overmind` is not installed.

Closes https://github.com/primer/view_components/issues/1366


keithamus (approved):
This looks reasonable to me

jonrohan (approved):

PR #1478, opened 2022-10-11
Update eslint plugin GitHub

koddsson:
### Description

- Use shareable browserslist config from https://github.com/github/browserslist-config/.
- Update `eslint-plugin-github` and use the correct typescript config.
- Also update the `caniuse-lite` db to include Safari 16.Finally fix a shadowed variable to silence lint rule.

### References

https://github.com/primer/view_components/pull/1474


keithamus (approved):
LGTM :+1:

PR #1480, opened 2022-10-11
Only register click handler once

koddsson:
### Description

The `<modal-dialog>` registers a global event handler each time it connects to the DOM. We actually just want one global event handler so that we don't open or close multiple dialogs when clicking around on the page.

This PR makes sure we only register the event handler once by using a function reference instead of a inlined function for the click handler.

Closes https://github.com/primer/view_components/issues/1366 (again :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "cssnano": "^5.1.13",
     "eslint": "^8.23.1",
     "eslint-plugin-custom-elements": "^0.0.6",
-    "eslint-plugin-github": "^4.3.7",
+    "eslint-plugin-github": "^4.4.0",
     "eslint-plugin-prettier": "^4.2.1",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -83,5 +84,6 @@
     "tslib": "^2.4.0",
     "typescript": "^4.7.4"
   },
-  "prettier": "@github/prettier-config"
+  "prettier": "@github/prettier-config",
+  "browserslist": "extends @github/browserslist-config"
```
jonrohan, line 88: There's a browserslistrc file I added recently, could you delete that in favor of this one?

jonrohan: Edit: saw that you did 😂

jonrohan (approved):

PR #1482, opened 2022-10-11
Deploy production docs and lookbook app only when a release is created

jonrohan:
### Description

This might be a useful change because it will keep our docs/lookbooks pinned to the previous release. This could help any confusion caused by unreleased docs making their way into the site.

However there might be emergencies where we would need to deploy the site manually. For that I'm adding a `workflow_dispatch` trigger which will allow us to deploy at the click of a button. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

### Integration

no

### DIFF
--- a/.github/workflows/demo-production-deploy.yml
+++ b/.github/workflows/demo-production-deploy.yml
@@ -3,6 +3,7 @@
 name: Demo Production
 
 on:
+  workflow_dispatch:
   release:
     types: [published]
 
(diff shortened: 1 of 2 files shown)


langermank (approved):
This is excellent. Next we should try and print out PVC version number in Lookbook so its clear what folks are viewing 😄

PR #1494, opened 2022-10-13
Updates to the flash component to use new banner styles

camertron:
### Description

A number of months ago @vdepizzol [refactored](https://github.com/primer/css/pull/1979) the CSS for our flash messages. Unfortunately at the time there wasn't capacity on the Rails team to integrate it with PVC, so it has sadly languished for some time. This PR adds it into PVC as a .pcss file and includes the necessary markup changes.

### Integration

No.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews

Co-authored with @vdepizzol and @langermank


app/components/primer/beta/flash.rb
```diff
@@ -8,21 +8,15 @@ class Flash < Primer::Component
...
-      # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
-      renders_one :action, lambda { |**system_arguments|
-        deny_tag_argument(**system_arguments)
-        system_arguments[:tag] = :div
-        system_arguments[:classes] = class_names(system_arguments[:classes], "flash-action")
-
-        Primer::BaseComponent.new(**system_arguments)
-      }
+      # @param system_arguments [Hash] The same arguments as <%= link_to_component(Primer::Beta::Button) %>.
+      renders_one :action, Primer::Beta::Button
```
jonrohan, line 12: I saw a few cases using things other than buttons in the `with_action` https://github.com/search?q=repo%3Agithub%2Fgithub+.with_action&type=code

app/components/primer/beta/flash.rb
```diff
@@ -42,29 +36,27 @@ class Flash < Primer::Component
...
       #   <% end %>
       #
       # @param full [Boolean] Whether the component should take up the full width of the screen.
-      # @param spacious [Boolean] Whether to add margin to the bottom of the component.
+      # @param full_when_narrow [Boolean] Whether the component should take up the full width of the screen when rendered inside smaller viewports.
       # @param dismissible [Boolean] Whether the component can be dismissed with an X button.
       # @param icon [Symbol] Name of Octicon icon to use.
       # @param scheme [Symbol] <%= one_of(Primer::Beta::Flash::SCHEME_MAPPINGS.keys) %>
       # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
-      def initialize(full: false, spacious: false, dismissible: false, icon: nil, scheme: DEFAULT_SCHEME, **system_arguments)
```
jonrohan, line 56: We'll need some updates to code in production on github for this.

previews/primer/beta/flash_preview/with_action.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= render(Primer::Beta::Flash.new(full: full, dismissible: dismissible, icon: icon, scheme: scheme)) do |component| %>
+  <% component.with_action(size: :small) { "Take action" } %>
+  <%= content %>
+<% end %>
```
jonrohan, line 4: Not sure if it's meant to be this way, but the button alignment felt off here for a single line 
image

jonrohan (commented):
There's currently `680` uses of `Primer::Beta::Flash` in dotcom, I worry a bit about the changes required when we are updating the gem. I wonder if it'd be easier on us to create a new `Primer::Alpha::Banner` class here and deprecate the `Flash` component.

app/components/primer/beta/flash.rb
langermank: What if we rename it to `Banner` 😛

previews/primer/beta/flash_preview/with_action.html.erb
langermank: This felt off to me as well. I center aligned it.

.changeset/empty-ghosts-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Use @vdepizzol's new banner CSS for the `Flash` component.
```
jonrohan, line 5:
```suggestion
Updates to `Primer::Beta::Flash` to use new `.Banner` styles
```

app/components/primer/beta/flash.rb
```diff
@@ -11,18 +11,16 @@ class Flash < Primer::Component
       # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
       renders_one :action, lambda { |**system_arguments|
         deny_tag_argument(**system_arguments)
-        system_arguments[:tag] = :div
-        system_arguments[:classes] = class_names(system_arguments[:classes], "flash-action")
 
-        Primer::BaseComponent.new(**system_arguments)
+        Primer::Beta::Button.new(**system_arguments)
```
jonrohan, line 15: Apologies if still working on this, are we adding back support for `forms, include-fragment, details` here?

previews/primer/beta/flash_preview.rb
```diff
@@ -4,16 +4,28 @@ module Primer
...
+
+      # @label With action button
       #
       # @param full toggle
-      # @param spacious toggle
       # @param dismissible toggle
       # @param icon [Symbol] select [alert, check, info, people]
       # @param scheme [Symbol] select [default, warning, danger, success]
       # @param content text
-      def default(full: false, spacious: false, dismissible: false, icon: :people, scheme: Primer::Beta::Flash::DEFAULT_SCHEME, content: "This is a flash message!")
```
jonrohan, line 15: We'll need to keep at least one `default` preview. see https://github.com/github/primer/pull/1387

app/components/primer/beta/flash.rb
camertron: Yeah sorry, I haven't gotten a chance to switch back to `BaseComponent` yet. Actually thinking about making this a polymorphic slot.

app/components/primer/beta/flash.rb
camertron: Added a linter.

app/components/primer/beta/flash.rb
```diff
@@ -6,65 +6,108 @@ module Beta
...
+      # To render custom content, call the `with_action_content` method and pass a block that returns HTML.
+      renders_one :action, types: {
+        button: lambda { |**system_arguments|
+          deny_tag_argument(**system_arguments)
+
+          Primer::Beta::Button.new(**system_arguments)
+        },
 
-        Primer::BaseComponent.new(**system_arguments)
+        content: lambda { |**system_arguments|
```
jonrohan, line 21: 🙌🏻

lib/primer/view_components/linters/migrate_deprecated_flash_arguments.rb
```diff
@@ -0,0 +1,140 @@
...
+      end
+
+      def maybe_close_block(code)
+        match = code.match(BLOCK_EXPR)
+        return code unless match
+
+        if match.captures[0].strip == "do"
+          "#{code}; end"
+        else
+          "#{code} }"
```
jonrohan, line 135: The coverage linter is complaining this line isn't being run

jonrohan (approved):
Looks good, just a small coverage problem in the CI tests

app/components/primer/beta/flash.rb
```diff
@@ -6,65 +6,108 @@ module Beta
...
       #
       # @example Full width
       #   <%= render(Primer::Beta::Flash.new(full: true)) { "This is a full width flash message!" } %>
       #
       # @example Dismissible
-      #   <%= render(Primer::Beta::Flash.new(dismissible: true)) { "This is a dismissible flash message!" } %>
+      #   <%= render(Primer::Beta::Flash.new(dismissible: true, reappear: true)) { "This is a dismissible flash message!" } %>
       #
-      # @example Icon
+      # @example Custom icon
```
langermank, line 58: Can we change this to be the default icons per variant?

camertron: @langermank if you don't provide an icon, it defaults to one based on the scheme. In other words, if you don't specify which icon you want and the scheme is `:danger`, it will automatically choose the [`:stop` Octicon](https://primer.style/octicons/stop-24). Is that what you had in mind?

khiga8:
Hi! I was recently investigating flash as well and received feedback from @jscholes which I believe is a Primer-level update. 
I filed an issue here: https://github.com/github/primer/issues/1449. 

Could this be included here?

app/components/primer/beta/flash_banner.ts
```diff
@@ -0,0 +1,62 @@
...
+  connectedCallback() {
+    this.lastFocusedElement = (document.activeElement as HTMLElement) || document.body
+
+    if (document.readyState.toString() === 'ready' || document.readyState.toString() === 'complete') {
+      this.announceMessage()
+    } else {
+      document.addEventListener('readystatechange', () => {
+        if (document.readyState.toString() === 'complete') {
+          this.announceMessage()
+        }
```
khiga8, line 19: One thing that I noticed in dotcom was that with VoiceOver especially, was that the announcements would get interrupted. I believe that in dotcom, there are a lot of scripts that are still loading on the page and that somehow ends up interrupting the VoiceOver announcements. As a result, I ended up needing to wait for the `load` state that indicates when all the scripts have loaded, and delay on that. Wondering if you'd ran into that.

app/components/primer/beta/flash_banner.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (!parentElement) return
+
+    parentElement.removeChild(this)
+    this.lastFocusedElement?.focus()
+  }
+
+  private announceMessage() {
+    setTimeout(() => {
+      this.message.focus()
+    }, 200)
```
khiga8, line 45: Is there concern that focusing on delay can interrupt a user who is already starting to interact with the page? Something I jotted down in [my coaching findings was](https://github.com/github/accessibility/pull/2101/files#diff-5f93c27804a1cf56989f041af5356de22a12434616f5a05508873c22b57f5817R60-R62):

> In testing, I found that if we have a `role=”alert”` element and update the content with a non-breaking space on `load` state with a 200ms delay, we can pretty consistently get screen reader announcements. It is possible that focus management on load & delay alone will lead to SR announcement, but by the time page load completes, user may already be interacting with the page so we should not

lib/primer/view_components/linters/migrate_deprecated_flash_arguments.rb
camertron: Ah ok that makes sense. I can add a test that exercises this codepath.

app/components/primer/beta/flash_banner.ts
camertron: Right yeah, that makes sense. I haven't tested yet in dotcom, which I'm sure will uncover a number of issues 😅  I tried adding a `load` listener but it never fired 🤷 so that's why I ended up with this `readystatechange` approach.

app/components/primer/beta/flash_banner.ts
camertron: Hmm yeah that's a valid concern. Maybe we can just remove the delay then?

khiga8: > Maybe we can just remove the delay then?

Yeah, I think that removing the delay for focus events will minimize potential issues.

> Adding a non-breaking space to the content didn't work when I was testing in Lookbook 😢

It's possible the delay wasn't long enough given waiting on the `load` state in dotcom will result in a longer wait. Also if you use ` ` instead of `nbsp;`, the SR announcement will not kick off. Let me know if there's a branch I can check out!

app/components/primer/beta/flash_banner.ts
khiga8: Here is the load state in [document-ready.ts](https://github.com/github/github/blob/master/app/assets/modules/github/document-ready.ts#L24) in dotcom. Is this what you tried? 

I wonder if this might take the 75% success rate to 100%. If we cannot, we could try increasing the delay to see if it improves the announcements being made.

app/components/primer/beta/flash.rb
```diff
@@ -3,68 +3,61 @@
...
-          "flash-full": full
-        )
-        @system_arguments[:mb] ||= spacious ? 4 : nil
+      # @param system_arguments [Hash] The same arguments as <%= link_to_component(Primer::Beta::Banner) %>.
+      def initialize(**system_arguments)
+        super
+
+        @message_arguments[:aria] ||= {}
+        @message_arguments[:aria][:atomic] = "true"
+        @message_arguments[:role] = @scheme == :danger ? "alert" : "status"
```
khiga8, line 46: Here, I notice that we're setting Flash of type `:danger` to `role="alert"`,  and the other types of flash to `role="status"`. 

Since Flashes are displayed as a result of some user action, I think it could be argued that all flashes (whether it be of type `:danger`, `:success`, or `:warning`) are important enough to have `role="alert"` .

But, I can also understand the logic behind this update. 

Tagging @jscholes for thoughts on if we should distinguish what role to set based on the type of Flash.

jscholes: @khiga8 I think it is semantically correct to distinguish between roles.  However:

1. `role="status"` would need to be tested, in a similar way to how you tested your approach to alerts on dotcom.  It doesn't, AFAIK, have a corresponding system role like an alert, within the accessibility APIs on native operating systems.  So it's more of a standard live region.
2. Given that GitHub doesn't use alert that much, I don't think using it for all flashes would be inappropriate, if the testing fails/you don't have the cycles for it.

camertron: I was able to test this and can confirm `role="status"` works as well as `role="alert"`.

app/components/primer/beta/flash_banner.ts
camertron: I ended up increasing the delay to 500ms and it worked in all browsers with VoiceOver and NVDA 🎉

app/components/primer/beta/flash_banner.ts
camertron: In a Slack convo with folks from web-systems, we determined the `readystatechange` code is equivalent to the logic in document-ready.ts. Increasing the delay to 500ms seemed to bring the success rate up to 100% 😄

app/components/primer/beta/banner.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    <% end %>
+    <% if @dismissible %>
+      <div class="Banner-close">
+        <%= render(
+          Primer::Beta::IconButton.new(
+            scheme: :invisible,
+            icon: :x,
+            aria: { label: "Dismiss" },
+            data: { action: catalyst_action(event: "click", function: "dismiss") },
+            autofocus: true
```
khiga8, line 27: Should `Banner` component have `autofocus` too? Maybe this should be a conditional and set to `true` for `Flash` only?

app/components/primer/beta/flash.rb
jscholes: @camertron What screen readers and browsers did you test with, specifically?

app/components/primer/beta/banner.html.erb
camertron: Good call! I actually moved all the `Banner` stuff into a [separate PR](https://github.com/primer/view_components/pull/1533) and have removed `autofocus` from it there.

app/components/primer/beta/flash.rb
camertron: On Mac I tested with VoiceOver on Chrome, Safari, and Firefox. On Windows I tested with NVDA on Edge, Chrome, and Firefox.

PR #1500, opened 2022-10-17
Upgrade simplecov to latest version, and adjust testing of coverage in actions

jonrohan:
### Description

This PR upgrades `simplecov` to `0.21.2` and also consolidates the simplecov config into 1 file `.simplecov` that can be shared between processes.

The benefit of this is faster execution of coverage tests. Previously we were running coverage in `CI > test` runs and uploading to an asset that was then re-downloaded to test for pass/fail.

In this updated pr, the `test.yml` workflow will run the tests, then run a `bundle exec rake test:coverage` command that tests for coverage. https://github.com/primer/view_components/actions/runs/3268991155/jobs/5376035261#step:7:1

### Integration

Doesn't effect anything outside of the repository.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


lib/tasks/test.rake
```diff
@@ -9,14 +9,17 @@ namespace :test do
   Rake::TestTask.new(:single) do |t|
     ENV["TZ"] = "Asia/Taipei"
 
+    t.warning = false
     t.libs << "test"
     t.libs << "lib"
-    t.test_files = FileList[ENV["TESTS"]]
+    t.test_files = FileList[ENV["TEST"]]
```
jonrohan, line 15: Also fixed that typo cc @camertron 😅

PR #1507, opened 2022-10-18
Adding testing around writing previews

jonrohan:
### Description

This PR adds tests to codify some of the adr requirements written up here https://github.com/github/primer/pull/1387.

The document says that every preview should have a 

- Playground story _"Playground stories make use of controls and are intended to be fully configurable, exposing all available component props."_
- Default story _"Sets all the default values for all component props. Giving an example of how it looks without configuration"_

This PR does include some changes to the `static/` arguments json file that add new attributes. For this part I'm including a changeset.

### Integration

No

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


lib/tasks/docs.rake
```diff
@@ -507,7 +509,7 @@ namespace :docs do
   def lookbook_url(component)
     path = component.name.underscore.gsub("_component", "")
 
-    "https://primer.style/view-components/lookbook/inspect/#{path}/default/"
+    "https://primer.style/view-components/lookbook/inspect/#{path}/playground/"
```
langermank, line 512: I realize I specifically requested this, but I don't know what I was thinking. We actually do want the default for docs 🤦‍♀️ 

```suggestion
    "https://primer.style/view-components/lookbook/inspect/#{path}/default/"
```

langermank (approved):
Just one silly comment 😆 this is great!

PR #1528, opened 2022-10-20
[WIP] migrating Primer::ImageCrop to Primer::Alpha::ImageCrop

mxriverlynn:
⚠️ WIP ⚠️ 

### Description

this PR migrates `Primer::ImageCrop` to `Primer::Alpha::ImageCrop`, and marks the original as deprecated

### Integration

migrating files that use `Primer::ImageCrop` but not immediately. deprecations will be raised when a file that uses it is edited

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


previews/primer/image_crop_preview.rb
```diff
@@ -7,21 +7,21 @@ class ImageCropPreview < ViewComponent::Preview
     #
```
jonrohan, line 7: It's probably my fault not updating the migration script properly, but this whole file should move to `previews/primer/alpha` also

jonrohan (approved):

PR #1531, opened 2022-10-21
rubocop/config/default: Unset `DisabledByDefault`

issyl0:
### Description

- The `DisabledByDefault` config option made it so that "all cops in the default configuration are disabled, and only cops in user configuration are enabled", which makes cops "opt-in rather than opt-out" (source: https://github.com/rubocop/rubocop/blob/d1270c468cdb9a211229e36aef8d568ae3bfe607/config/default.yml#L91-L102).
- It's not immediately obvious that this gem has this config option.  This means that people may write new, custom cops in downstream projects, with no configuration in .rubocop.yml, and then they never run because all cops are disabled unless explicitly enabled or otherwise configured with `Include`/`Exclude`.
- RuboCop does not warn users that the config inheritance has set `DisabledByDefault` somewhere up the line, leading to users mistakenly not enabling cops they may have intended to.
- Other of GitHub's open source gems, like `rubocop-github`[1] and `rubocop-rails-accessibility`[2], have moved away from `DisabledByDefault`. This is the last one (at least that I can find in GitHub's main project's RuboCop gem inheritance tree).

[1] - https://github.com/github/rubocop-github/pull/119
[2] - https://github.com/github/rubocop-rails-accessibility/pull/7

### Integration

Nope.

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/.changeset/famous-teachers-lay.md
+++ b/.changeset/famous-teachers-lay.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+rubocop/config/default: Unset `DisabledByDefault`
--- a/lib/rubocop/config/default.yml
+++ b/lib/rubocop/config/default.yml
@@ -1,9 +1,6 @@
 require:
   - rubocop/cop/primer
 
-AllCops:
-  DisabledByDefault: true
-
 Primer/SystemArgumentInsteadOfClass:
   Enabled: true
 


jonrohan (approved):
Thanks for the clear explanation. This should be in the next release 105, probably out next week

BlakeWilliams (approved):

camertron:
Hey @issyl0, unfortunately we need to revert this PR because it's causing an [issue](https://github.com/primer/view_components/actions/runs/3300796576/jobs/5445655857) with our CI builds at the moment.

The problem is erblint only looks at snippets of Ruby code, eg `<%= render(MyComponent.new) %>`. In this example, applying regular 'ol Rubocop rules will say the line begins with indentation characters because it sees `[space]render(MyComponent.new)[space]`. That's why our config originally disables all cops and enables only a select few. We should probably have a separate erblint and Rubocop config.

camertron:
Ok, it looks like disabling cops can be done directly inside .erb-lint.yml 🎉 Here's the PR: https://github.com/primer/view_components/pull/1535

issyl0:
Thanks and sorry for the friction here @camertron, I didn't know you used `DisabledByDefault` in this way as a considered choice. Doing it in this project's `erb-lint.yml` for your tests makes sense to me, so it doesn't pollute the gem config that is shipped to users.

PR #1533, opened 2022-10-21
Banner component

camertron:
### Description

This PR introduces the `Banner` component, which will eventually form the basis for an improved `Flash` component. I'd like to ship this before shipping the [`Flash` component upgrade](https://github.com/primer/view_components/pull/1494). The idea here is that `Flash` is a `Banner` with some extra javascript for announcing the message to screen readers.

### Integration

No, prod uses the `Flash` component to display banners. `Flash` should only be used to display announced alert and status messages, so my plan is to replace all usages of `Flash` with `Banner` in dotcom, then land the `Flash` upgrade and start using it for flash-specific purposes in dotcom.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


jonrohan (approved):
👍🏻

PR #1538, opened 2022-10-25
Export component CSS selectors as json file for testing

jonrohan:
### Description

In support of eventually having https://github.com/github/primer/issues/1443 this PR introduces in the build a step that will export all the CSS selectors as a json file that can be read by our tests/linters/etc.

I opted to make a json file for each component so that we might be able to compare component css to component previews in an eventual test.

An [example output](https://unpkg.com/browse/@primer/view-components@0.0.0-20221025160746/app/components/primer/beta/banner.css.json) looks like this: 

```json
{
  "name": "primer/beta/banner",
  "selectors": [
    ".Banner",
    ".Banner .Banner-actions"
  ] 
}
```

### Integration

no 

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


app/components/primer/alpha/action_list.pcss
```diff
@@ -168,6 +168,97 @@
...
+  @keyframes checkmarkOut {
+    from {
+      clip-path: inset(0 0 0 0);
+    }
+
+    to {
+      clip-path: inset(16px 0 0 0);
+    }
+  }
+
```
jonrohan, line 261: This code came from `action-list-selection` to combine into one file. However I noticed this file wasn't actually being imported in `primer.pcss` not sure if this was an oversight or intended to not be included. @camertron do you know?

camertron: Ooo interesting! I'm not sure, probably a mistake? @langermank do you know if we need this file?

camertron (approved):

PR #1539, opened 2022-10-25
Bug Fix: Fixing bad CSS in action-list.css and writing test to check for error

jonrohan:
### Description

I noticed in our output we have some invalid CSS which is the result in typos from migrating SCSS to PostCSS. You can see this error on the [latest release of action-list.css](https://unpkg.com/@primer/view-components@0.0.106/app/components/primer/alpha/action_list/action-list.css) file. There's a `&` in the output which isn't valid CSS.

- I fixed the error in the CSS files
- Added a new test to check for things that might end up in bad output
- Added a stylelint rule to catch issues

### Integration

no

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/.changeset/little-bees-doubt.md
+++ b/.changeset/little-bees-doubt.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+Bug Fix: Fixing bad CSS in action-list.css and writing test to check for error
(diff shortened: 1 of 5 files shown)


camertron (approved):
Awesome, thank you. How did this happen? ActionList seemed to work ok when I was testing.

PR #1540, opened 2022-10-25
Write test to make sure asset naming matches component

jonrohan:
### Description

This PR introduces a test to make sure that assets are named to match the component ruby file. This follows convention [setup for view components](https://viewcomponent.org/guide/javascript_and_css.html) and also allows us to write tests around this convention to match assest with their component.

```
app/components
  ├── component.rb
  ├── component.pcss
  └── component.html.erb
```

### Integration

no

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated previews


camertron (changes requested):
Hmm I'm not sure I'm in favor of this change. I learned recently that web components are required to include a hyphen in their names, eg. `nav-list` instead of `navlist`. Up until now it hasn't been a problem, but I was bitten last week by `banner` and `flash`. Apparently it's common to prefix single-word web component names with `x-`, so the custom element names are `x-banner` and `x-flash`. We have a prettier rule that forces the element name to match the file name, so I had to use x_banner.ts. I really don't want to have to name the component x_banner.rb too 😓

jonrohan:
> Hmm I'm not sure I'm in favor of this change. I learned recently that web components are required to include a hyphen in their names, eg. `nav-list` instead of `navlist`. Up until now it hasn't been a problem, but I was bitten last week by `banner` and `flash`. Apparently it's common to prefix single-word web component names with `x-`, so the custom element names are `x-banner` and `x-flash`. We have a prettier rule that forces the element name to match the file name, so I had to use x_banner.ts. I really don't want to have to name the component x_banner.rb and x_banner.html.erb too 😓

I can drop the `ts` file requirement. I do need the CSS files to match so I can build testing around class coverage

camertron (approved):
Ok I'm cool with that. We might also consider relaxing the eslint rule.

Issue #1551, opened 2022-10-28
Primary button scheme is not focusable


jonrohan:
https://github.com/github/github/pull/244059#discussion_r1007436087

The primary scheme on the Beta button doesn't have a focus ring

PR #1552, opened 2022-10-28
Disable focus-visible-pseudo-class plugin to fix primary Button

jonrohan:
### Description

This fixes an issue with the primary scheme of the `Primer::Beta::Button` component. The button wasn't accepting a focus ring. 

I discovered the issue was that our PostCSS plugin `focus-visible-pseudo-class` was adding a polyfill class that was not being applied to our elements. This caused the button to never be focused. https://preset-env.cssdb.org/features/#focus-visible-pseudo-class

Closes https://github.com/primer/view_components/issues/1551

### Integration

No

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/.changeset/metal-chicken-applaud.md
+++ b/.changeset/metal-chicken-applaud.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+Disable focus-visible-pseudo-class plugin to fix primary Button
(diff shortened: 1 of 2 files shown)


camertron (approved):
Wow, nice find! 🎉

PR #1553, opened 2022-10-28
Run accessibility tests in parallel

camertron:
### Description

This PR runs our accessibility tests in parallel to speed them up. It uses our org's newly-available 8-core runners and spins up 4 test workers (plus 4 puma servers).

A recent build of this branch completed running our accessibility tests in 2 minutes and 34 seconds. A recent build of main did the same in 6 minutes and 18 seconds, meaning **this PR reduces a11y run time by about 3 minutes and 34 seconds**. I have observed that our accessibility tests now take about as long to run as our system tests.

NOTE: a change will need to be made to the view_component test suite so it continues to run our accessibility tests in a single process, as its org does not have access to larger runners.

### Integration

Nope! These changes only affect tests.

### Merge checklist

- [x] Added/updated tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ concurrency:
 jobs:
   accessibility:
     name: Accessibility
-    runs-on: ubuntu-latest-16-cores
+    runs-on: ubuntu-latest-8-cores
     steps:
     - uses: actions/checkout@v3
     - uses: ruby/setup-ruby@v1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ concurrency:
 jobs:
   accessibility:
     name: Accessibility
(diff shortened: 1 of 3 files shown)


camertron:
I'm blocked on this until we can get larger hosted runners in GitHub actions. Default runners have 2 CPUs, so you'd think we'd be able to spin up 2 testing processes, one per CPU. Unfortunately our a11y tests spin up two Ruby process and two Cuprite/Chrome processes, which can't effectively run in parallel. The main test process also needs resources, so to really fly we need at least 4 CPUs. Fortunately larger runners are in beta right now. I've signed up for the beta and requested expedited access for the primer org [here](https://github.com/github/c2c-actions/issues/5500).

jonrohan (approved):
Fantastic! 
Screenshot 2022-11-02 at 4 00 55 PM

PR #1555, opened 2022-10-31
Removing snapshot testing related code

jonrohan:
### Description

This removes the snapshot code I added. I found it was too unpredictable and sometimes would generate new snapshots after changes to the README.md file.

The closest I got to a fix for this is in https://github.com/primer/view_components/pull/1455 but it was still having issues

image

So I'm opting to close and remove this until we can integrate something better. https://github.com/primer/react/pull/2491

### Integration

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


langermank (approved):
Oh bummer! Hope we can find a better solution later :D

Issue #1440, opened 2022-10-04
ToggleSwitchElement: Uncaught DOMException


safeforge:
I'm using the library `@primer/view-components` as an npm package and compiling it into the rails 7 application with `esbuild`. After the upgrade from `0.0.92`, I receive the current error in the javascript console. 

```console
toggle_switch.js:141 Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': this constructor has already been used with this registry
    at http://localhost:3000/assets/application-5fdd23dba4c210059bbefff8367af8dc8ea85c1da52231cab49e88648fb0561b.js:6076:27
    at http://localhost:3000/assets/application-5fdd23dba4c210059bbefff8367af8dc8ea85c1da52231cab49e88648fb0561b.js:8129:3
(anonymous) @ toggle_switch.js:141
(anonymous) @ index.ts:140
```

I tried to investigate and I discovered if I remove the `export` from `alpha/toggle_switch.js` everything works as expected. 

```js
ToggleSwitchElement = __decorate([
    controller
], ToggleSwitchElement);
//export { ToggleSwitchElement };
if (!window.customElements.get('toggle-switch')) {
    window.ToggleSwitchElement = ToggleSwitchElement;
    window.customElements.define('toggle-switch', ToggleSwitchElement);
}
```

@camertron since you're the author of #1321 can you tell me if it's a library bug and point me in the right direction on how to fix it?

Thank you for your support.

lesliecdubs:
Hi @safeforge, sorry to hear you're running into this issue! 

I'm adding this to our team project board so we can triage and get back to you if @camertron hasn't weighed in by then. Thanks!

PR #1560, opened 2022-11-01
Don't export web components

safeforge:
As mentioned at #1450 exporting a component can cause errors when used as an npm package.  In the latest release two components `SegmentedControlElement` and  `XBannerElement` perform the export and I'm receiving the same error reported in #1440.

The code contains a fix with the same approach used by @camertron 

/cc @jonrohan


keithamus (approved):
This looks reasonable to me.

We probably want to export the components to allow for extensions, but we can resolve that issue later, for now let's get this error resolved.

PR #1575, opened 2022-11-04
Allow form field names to end in "?"

neall:
### Description

Rails will allow input names that end in a question mark and call the appropriate method on the model to (for example) decide if a checkbox should start off as "checked". This change ensures we can still find a caption template associated with a field named with a question mark, but not have to embed a question mark in the template file name.

### Integration

No changes in production. (Prod doesn't use this feature yet.)

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews

### DIFF
--- a/.changeset/old-beds-mate.md
+++ b/.changeset/old-beds-mate.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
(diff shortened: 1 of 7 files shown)


camertron (approved):
🎉

PR #1578, opened 2022-11-07
Adding an assets panel to the lookbook demo to show assets for the component

jonrohan:
### Description

image

This PR adds an "Assets" [custom panel](https://lookbook.build/guide/extend/panels/) to the demo app that was mostly copied from the lookbook example app. 

[Preview URL](http://primer-view-components-preview-1578.eastus.azurecontainer.io/view-components/lookbook/inspect/primer/beta/button/playground)

### Integration

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -17,6 +17,6 @@ module.exports = {
         'focus-visible-pseudo-class': false
       }
     }),
-    require('cssnano'),
+    process.env.NODE_ENV === 'development' ? null : require('cssnano')
   ],
 };
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -17,6 +17,6 @@ module.exports = {
         'focus-visible-pseudo-class': false
       }
     }),
-    process.env.CI === '1' ? require('cssnano') : null
+    process.env.CI === 'true' ? require('cssnano') : null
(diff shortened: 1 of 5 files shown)


postcss.config.js
```diff
@@ -17,6 +17,6 @@ module.exports = {
         'focus-visible-pseudo-class': false
       }
     }),
-    require('cssnano'),
+    process.env.CI === 'true' ? require('cssnano') : null
```
jonrohan, line 20: I'm trying to get the CSS to not be compressed output for the new panel, my thinking here is to set to `null` in the proper circumstances.

langermank (approved):
Love this!

postcss.config.js
camertron: What if we just read in the un-compiled .pcss file instead of the compiled .css file?

camertron (approved):
Super cool!

postcss.config.js
jonrohan: Was thinking if we're sharing the compiled CSS it would be easier to copy/paste to a prototyping tool like Codepen.

PR #1579, opened 2022-11-07
Move Banner to alpha namespace

camertron:
### Description

The `Banner` component currently lives in the `Beta` namespace because it will eventually form the basis of the `Flash` component, which is also in `Beta`. However, since `Banner` uses new markup and new styles, it's more correct for it to be in the `Alpha` namespace. It will need to be evaluated against the beta criteria before being elevated to that status.

### Integration

Yes. There are only a few instances of `Primer::Beta::Banner` in production though, so the migration should be easy.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


app/components/primer/alpha/banner.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Primer
-  module Beta
+  module Alpha
     # Use `Banner` to highlight important information.
     class Banner < Primer::Component
       status :beta
```
jonrohan, line 7:
```suggestion
      status :alpha
```

jonrohan (commented):
Noticed one thing

app/components/primer/alpha/banner.rb
camertron: Woops, thanks!

jonrohan (approved):

PR #1584, opened 2022-11-08
Move `blankslate` styles to PVC

simurai:
### Description

This adds the `blankslate` styles from [PCSS](https://github.com/primer/css/blob/5a0b9b2939c1428430d249aeeb9adb0ba8bc18ce/src/blankslate/blankslate.scss).

There should be no visual changes. See this commit https://github.com/primer/view_components/commit/a263494b9ed92c62eec9f6b9df3834ea853b5cdf for the diff of migrating from `.scss` to `.pcss`.

Closes https://github.com/github/primer/issues/1497

### Integration

Yes, [this line](https://github.com/github/github/blob/e009ce8f20d5700a7f4a581e3c7953951469bf9c/app/assets/stylesheets/bundles/primer/index.scss#L121) should be updated on dotcom:

```diff
- @import '@primer/css/blankslate/blankslate.scss';
+ @import '@primer/view-components/app/assets/styles/blankslate.scss';
```

### Merge checklist

- [ ] ~~Added/updated tests~~
- [ ] ~~Added/updated previews~~
- [x] Visual regression test

Before | After
--- | ---
Screen Shot 2022-11-08 at 15 18 09 | Screen Shot 2022-11-08 at 15 18 18

If you can't spot a difference 👉  good!

### DIFF
--- a/app/components/primer/primer.pcss
+++ b/app/components/primer/primer.pcss
@@ -3,3 +3,4 @@
 @import "./alpha/banner.pcss";
 @import './alpha/segmented_control.pcss';
 @import "./beta/button.pcss";
+@import "./beta/blankslate.pcss";
--- a/app/components/primer/beta/blankslate.pcss
+++ b/app/components/primer/beta/blankslate.pcss
@@ -1,0 +1,89 @@
+// stylelint-disable selector-max-type
+.blankslate {
+  position: relative;
+  padding: $spacer-5;
+  text-align: center;
+
+  p {
+    color: var(--color-fg-muted);
+  }
+
+  code {
+    // stylelint-disable-next-line primer/spacing
+    padding: 2px 5px 3px;
+    font-size: $h5-size;
+    background: var(--color-canvas-default);
+    border: $border-width $border-style var(--color-border-muted);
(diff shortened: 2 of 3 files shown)


app/components/primer/primer.pcss
```diff
@@ -3,3 +3,4 @@
 @import "./alpha/banner.pcss";
 @import './alpha/segmented_control.pcss';
 @import "./beta/button.pcss";
+@import "./beta/blankslate.pcss";
```
simurai, line 6: If we import it here (to be used in Lookbook), I guess we'll have to expand the import on dotcom. Like https://github.com/github/github/pull/245966.

Or an alternative would be to not import `blankslate` here, but update the Lookbook import to something like:

```diff
- *= require @primer/css/dist/blankslate.css
+ *= require ../../../../../app/components/primer/beta/blankslate.css
```

But I haven't figured out the exact local path.

simurai: nvm.. We decided to not expand the PVC imports on dotcom. It might cause regression changing the source order. But we'll have to do that eventually anyways.

jonrohan (approved):

PR #1588, opened 2022-11-09
Check accessibility of previews with `axe-core` npm package.

koddsson:
### Description

Use `axe-core` from npm in Cuprite System tests. I also made sure to update to `axe-core@4.5`.

### Integration

No

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews

### References

Closes https://github.com/github/platform-ux/issues/1122


keithamus (approved):
Oh yeah! This looks great!

jonrohan:
Related, https://github.com/primer/view_components/pull/1589

If we switch to playwright, then the https://www.npmjs.com/package/@axe-core/playwright axe-core playwright package might be easier to use?

koddsson:
> If we switch to playwright, then the https://www.npmjs.com/package/@axe-core/playwright axe-core playwright package might be easier to use?

Yeah! I'm sure `@axe-core/playwright` would be pretty much plug and play into a JS playwright setup. This PR plugs better into our existing test infrastructure. I think using playwright directly instead of Capybara is the way to go eventually but I'd prefer to have it implemented in ruby with the existing tests.

test/system/test_case.rb
```diff
@@ -33,34 +31,75 @@ def visit_preview(preview_name, params = {})
...
 
-      assert_accessible(page)
+      assert_accessible
+    end
+
+    def format_accessibility_errors(violations)
+      results = violations.map.with_index do |summary, index|
+        summary["nodes"].map do |node|
+          violations = node["any"] + node["all"]
+          <<~OUTPUT
```
camertron, line 41: `<<~`-style heredocs strip leading whitespace, so I'm not sure why things are indented so far. I'd probably reach for something like ERB to do this. Happy to pair if you like.

previews/primer/alpha/tooltip_preview.rb
```diff
@@ -7,7 +7,7 @@ class TooltipPreview < ViewComponent::Preview
       # @param type [Symbol] select [["Description", description], ["Label", label]]
       # @param direction select [s, n, e, w, ne, nw, se, sw]
       # @param tooltip_text text
-      def playground(type: :description, direction: :s, tooltip_text: "Tooltip text")
+      def playground(type: :description, direction: :s, tooltip_text: "You can press a button")
```
camertron, line 10: Why are all these copy changes necessary?

camertron (changes requested):

previews/primer/alpha/tooltip_preview.rb
koddsson: I forget what the rule name was but it's essentially that the screenreader text needs to include the text of the element so that there isn't a huge mismatch in the information conveyed to users.

test/system/test_case.rb
koddsson: I'd love to pair on this but I think we're on pretty incompatible timezones 😬 

I didn't realize the whitespace was stripped from this syntax, I think I just copied it from somewhere. That's good to know :)

camertron: Just pushed up some formatting fixes, we should be good to land this now :)

camertron (approved):

jonrohan (approved):

PR #1589, opened 2022-11-09
Using playwright to generate snapshots

jonrohan:
### Description

Describe your changes here.

Closes # (type the issue number after # if applicable; otherwise remove this line)

### Integration

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


test/playwright/previews.test.ts
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable no-unused-vars */
+import {test, expect} from '@playwright/test'
+import {componentPreviews} from './helpers'
+
+test.beforeEach(async ({page}, testInfo) => {
+  testInfo.snapshotSuffix = ''
```
joshblack, line 7: Have you had any trouble with the snapshots on macOS and ubuntu with this change? Would definitely simplify stuff if we didn't have to use docker 👀

playwright.config.ts
```diff
@@ -0,0 +1,48 @@
...
+  testMatch: '**/*.test.ts',
+  /* Maximum time one test can run for. */
+  timeout: 30 * 1000,
+
+  // https://playwright.dev/docs/api/class-testconfig#test-config-output-dir
+  outputDir: path.join(__dirname, '.playwright', 'results'),
+  snapshotDir: path.join(__dirname, '.playwright', 'screenshots'),
+
+  /* Run tests in files in parallel */
+  fullyParallel: true,
```
joshblack, line 19: Such a good option 🔥 Did not realize that it's parallel on test files by default

joshblack (approved):
So good 🔥

test/playwright/previews.test.ts
jonrohan: I wasn't worried about it too much, in a codespace it's the same environment as CI. And if someone did run it on macos and checked it in, the ci would override them with new screenshots since we check in any changes the ci job detects.

PR #1595, opened 2022-11-10
Add lazy loader to action_controller_base

soberstadt:
### Description

Instances of `ActionController::API` subclasses are currently incompatible with the primer engine. 

image

### Integration

No breaking changes.

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


camertron:
Hey @soberstadt, can you provide a bit more info around why this change is necessary? Things seem to be working for us as-is.

soberstadt:
Do you have an controllers that directly extend `ActionController::API`? For example in my application we're using doorkeeper which exposes an API who's controller extends `ActionController::API`. I'll try to put together a demo app to reproduce my error.

soberstadt:
Demo app: https://github.com/soberstadt/view_components1595
I wasn't able to get it working as a single page app.

image

camertron:
Hey @soberstadt, apologies for the delay here. I'm still not sure I understand the change, but if it works for you and doesn't adversely affect github.com, I'm ok merging it.

Could you add a changeset when you get a chance? Run `npx changeset` and follow the prompts (we always bump the patch version btw).

camertron (approved):
Thank you!

PR #1574, opened 2022-11-03
Component-Specific CSS Selector Tests

mxriverlynn:
### Description

This PR adds component-specific CSS selector tests to PVC. It looks at the css created for a specific component, and compares it to the CSS selectors that are produced by the various lookbook previews for the component. selectors that are not found in the immediate previews (not counting preview params) will fail the component-specific selector CI test and list the components that need additional previews.

some custom selectors are not applicable to this test and can be ignored by modifying the `IGNORED_SELECTORS` const in `test/css/component_specific_selectors_test.rb`. these often include selectors based on state, such as `:hover` or a specific `aria` tag.

### Current Failures

This PR contains a few known failures in the CI step for testing selectors. This is caused by existing components with many selectors that either need to be ignored or for which previews need to be built. Because of the current volatility in these selectors and failures, the CI step for testing selectors is left as optional. This should allow the PR to land as-is, and let future PRs add the needed ignores and previews.

These components are currently known to be failing:
* `Primer::Alpha::ActionList`
* `Primer::Beta::BlankSlate`
* `Primer::Beta::Button`
* `Primer::Beta::Label`

Once these components are passing successfully, it's recommended the CI step for testing custom selectors be marked as required.

### PR Task List:

- [x] adds tests that ensure all selectors defined in a component's CSS are used in previews
- [x] allow component-specific selector ignores, not just global patterns
- [x] group component options in lookbook previews, to ensure preview coverage of options
- [x] setup component-specific css selector tests as a separate, optional CI step

### Component Preview Groups

There was a need to add many new previews that only differ by one param. Rather than exploding the list of items in the left hand navigation, preview groups are being added to the components where it makes sense. 

For example, this is the new preview for `Primer::Alpha::SegmentedControl`, looking at the "Full Width" group:
image
The full list of components that have grouped previews, added in this PR:
* `Primer::Alpha::Banner`
* `Primer::Alpha::SegmentedControl`
* `Primer::Beta::Counter`
* `Primer::Beta::ProgressBar`

### Integration

> Does this change require any updates to code in production?

no. this primarily applies to CI builds of PVC while incidentally improving lookbook previews for some components

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation
- [x] Added/updated previews

### Run Component-Specific Selector tests

```
bundle exec rake test:component_css
```

### Example Output from Failed Run:
```
❯ bundle exec rake test:component_css
Run options: --seed 13793

# Running:

...........................................................F

Failure:
[...]

PR #1602, opened 2022-11-14
Update css selector test

simurai:
### Description

This is a follow-up to https://github.com/primer/view_components/pull/1574 and adds updates to:

- [x] Label
- [x] Blankslate

The selectors added to `IGNORED_SELECTORS` are mostly old/deprecated classes that we still need to migrate on dotcom before we can remove them. 

### Integration

No

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews

### DIFF
--- a/test/css/component_specific_selectors_test.rb
+++ b/test/css/component_specific_selectors_test.rb
@@ -16,10 +16,18 @@ class ComponentSpecificSelectorsTest < Minitest::Test
   IGNORED_SELECTORS = {
     :global => [/^\d/, ":"],
     Primer::Alpha::ActionList => [/^to/],
-    Primer::Alpha::Banner => [".Banner .Banner-close"],
-    Primer::Alpha::SegmentedControl => [".Button-withTooltip"],
-    Primer::Beta::Button => ["summary.Button"],
-    Primer::Beta::Counter => ["Counter .octicon"],
+    Primer::Alpha::Banner => [
+      ".Banner .Banner-close"
+    ],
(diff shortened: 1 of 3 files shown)


test/css/component_specific_selectors_test.rb
```diff
@@ -16,10 +16,34 @@ class ComponentSpecificSelectorsTest < Minitest::Test
   IGNORED_SELECTORS = {
     :global => [/^\d/, ":"],
```
jonrohan, line 17: I'm wondering if we should add a `[` character to ignored selectors? That would remove any that have attribute selectors on them like `[hidden]`. I'm just not sure it's feasible to test against those because presumably they're all event based like clicks

```suggestion
    :global => [/^\d/, ":", "["],
```

mxriverlynn: +1

if we want to be a bit more specific, `/\[.*\]/` will look for the full attribute selector, like `[aria-label="foo"]`, but i don't know if it makes much difference (or any at all)

either one of these options should also take care of the `.ActionListItem[hidden]` issue noted above

mxriverlynn (approved):
add the `[` ignore and this looks good to me 👍🏻

simurai:
> @mxriverlynn add the `[` ignore and this looks good to me 👍🏻

Thanks! 🙇 Yes, that seems to ignore all the attributes in selectors.

Today I might look into adding more previews for `flash` and `dropdown` and ignore the rest (`ActionList/Button`).

PR #1606, opened 2022-11-15
Update previews and get CSS coverage to pass

simurai:
### Description

This is a follow-up to https://github.com/primer/view_components/pull/1602 and adds some more previews and ignores some selectors for:

- [x] `flash`
- [x] `dropdown`

To make the "CSS coverage" pass I also added the remaining selectors to `IGNORED_SELECTORS` for:

- [x] `ActionList`
- [x] `Button`

We can take a closer look once we move the styles from Primer CSS. But might be good to have `main` not fail for too long.

### Integration

No

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews

### DIFF
--- a/test/css/component_specific_selectors_test.rb
+++ b/test/css/component_specific_selectors_test.rb
@@ -14,8 +14,18 @@ class ComponentSpecificSelectorsTest < Minitest::Test
   include Primer::RenderPreview
 
   IGNORED_SELECTORS = {
-    :global => [/^\d/, ":", /\[.*\]/],
-    Primer::Alpha::ActionList => [/^to/],
+    :global => [/^\d/, ":", /\[.*\]/, /^to/, /^from/],
+    Primer::Alpha::ActionList => [
+      ".ActionListWrap--inset",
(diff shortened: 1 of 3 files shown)


test/css/component_specific_selectors_test.rb
```diff
@@ -14,16 +14,34 @@ class ComponentSpecificSelectorsTest < Minitest::Test
   include Primer::RenderPreview
 
   IGNORED_SELECTORS = {
-    :global => [/^\d/, ":", /\[.*\]/],
-    Primer::Alpha::ActionList => [/^to/],
```
simurai, line 18: I moved `/^to/` back to `:global` and also added `/^from/`. I think they are used for keyframe animations like:

```css
@keyframes checkmarkIn {
  from {
    clip-path: inset(16px 0 0 0);
  }

  to {
    clip-path: inset(0 0 0 0);
  }
}
```

and probably used for other components at some point.

jonrohan (approved):

PR #1607, opened 2022-11-16
Move `Avatar` + `AvatarStack` styles to PVC

simurai:
### Description

This is part of [#1342](https://github.com/github/primer/issues/1342) and adds the [`Avatar`](https://github.com/primer/css/blob/0370244f49322863534c62b5b5589128c551ccc7/src/avatars/avatar.scss) + [`AvatarStack`](https://github.com/primer/css/blob/0370244f49322863534c62b5b5589128c551ccc7/src/avatars/avatar-stack.scss) styles from PCSS. There should be no visual changes.

### Integration

Yes, the following lines can be removed on [dotcom](https://github.com/github/github/blob/e009ce8f20d5700a7f4a581e3c7953951469bf9c/app/assets/stylesheets/bundles/primer/index.scss#L114):

```diff
- @import '@primer/css/avatars/avatar.scss';
- @import '@primer/css/avatars/avatar-stack.scss';
```

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews
- [x] Visual regression test

Before | After
--- | ---
Screen Shot 2022-11-16 at 10 41 15 | Screen Shot 2022-11-16 at 10 41 21
Screen Shot 2022-11-16 at 10 42 46 | Screen Shot 2022-11-16 at 10 42 54


test/css/component_specific_selectors_test.rb
```diff
@@ -45,6 +45,21 @@ class ComponentSpecificSelectorsTest < Minitest::Test
...
+      ".avatar-link",
+      ".avatar-group-item",
+      ".avatar-1",
+      ".avatar-2",
+      ".avatar-3",
+      ".avatar-4",
+      ".avatar-5",
+      ".avatar-6",
+      ".avatar-7",
+      ".avatar-8"
```
simurai, line 58: These sizes are not use in PVC and instead the `width/height` is inlined directly. I guess we could:

1. Use these classes for sizing in PVC
2. Move the styles to dotcom

jonrohan: Hmm, looking at the react component, [they accept size in pixels](https://primer.style/react/Avatar), so I think we should move these to dotcom.

simurai: Yeah. I guess we should also move `.avatar-link` and `.avatar-group-item` to dotcom. Or even replace them with utility classes since they are sparingly used.

Anyways, I think I'll leave that for another "cleanup unused CSS" round.

jonrohan (approved):

PR #1611, opened 2022-11-17
Don't export XBannerElement

safeforge:
@camertron for some reason during the promotion from beta to alpha #1579 the export is present again, we should avoid registering the component twice as mentioned in #1560 and #1440.


jonrohan (approved):
Thanks

PR #1618, opened 2022-11-18
Move `Box` styles to PVC

simurai:
### Description

This is part of [#1342](https://github.com/github/primer/issues/1342) and adds the [`Box`](https://github.com/primer/css/blob/c8100be771634093fe3a4d36481fc43c0d966015/src/links/link.scss) styles from PCSS. There should be no visual changes.

### Integration

Yes, the [following line](https://github.com/github/github/blob/68ac277a9c271120ebf606b6fd457b2ed8b905e5/app/assets/stylesheets/bundles/primer/index.scss#L34) can be removed on dotcom:

```diff
- @import '@primer/css/box/box.scss';
```

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews
- [x] Visual regression test

Before | After
--- | ---
Screen Shot 2022-11-18 at 17 45 32 | Screen Shot 2022-11-18 at 17 45 14


test/css/component_specific_selectors_test.rb
```diff
@@ -34,6 +34,21 @@ class ComponentSpecificSelectorsTest < Minitest::Test
...
+      ".Box-row.unread",
+      ".Box-row.navigation-focus",
+      ".Box-row--focus-gray",
+      ".Box-row--focus-blue",
+      ".Box-row-link",
+      ".Box-row--drag-button",
+      ".Box--scrollable",
+      ".Box--blue",
+      ".Box--danger",
+      ".Box-header--blue"
```
simurai, line 50: Since there are quite a few un-used `.Box` styles, an alternative would be:

1. Rename `.Box` classes to `.BorderBox`. To match the `Primer::Beta::BorderBox` name.
2. Then only keep the styles that are needed for PVC.
3. On dotcom we can keep importing the `.Box` styles from PCSS until all use cases are replaced either by `BorderBox` or the upcoming `DataTable`.

I might do that on Monday.

simurai: Ok, nvm... seems that `Primer::Beta::BorderBox` is already too dependent on some of the undocumented styles of `.Box`. Like `Box-btn-octicon` [being used inside `Primer::Beta::BorderBox`](https://github.com/search?q=repo%3Agithub%2Fgithub+Primer%3A%3ABeta%3A%3ABorderBox+Box-btn-octicon&type=code). I guess we _could_ add 

```diff
+ .BorderBox,
  .Box {
    ...
  }
```

to PCSS. But that also feels backwards and keeps things hard to maintain. So I think we should do:

1. Keep this PR "as is".
2. Then at some point, go through all the ignored `.Box` classes and decide if they should become part of the `Primer::Beta::BorderBox` API. For example [`.Box--blue` is only used once on dotcom](https://github.com/search?q=repo%3Agithub%2Fgithub+Box--blue&type=code). So we probably just want to use a blue border for that case and remove the `.Box--blue` class in PVC.

PR #1626, opened 2022-11-21
Move `navigation` styles to PVC

simurai:
### Description

This is part of [#1342](https://github.com/github/primer/issues/1342) and adds the [`navigation`](https://github.com/primer/css/tree/f3acb2b116f10955e1ba4f1a35aa874eed019ceb/src/navigation) styles from PCSS. There should be no visual changes.

### Integration

> Does this change require any updates to code in production?

Yes, the [following lines](https://github.com/github/github/blob/68ac277a9c271120ebf606b6fd457b2ed8b905e5/app/assets/stylesheets/bundles/primer/index.scss#L68) can be removed on dotcom:

```diff
  // @primer/css/navigation/
- @import '@primer/css/navigation/menu.scss';
- @import '@primer/css/navigation/tabnav.scss';
  @import '@primer/css/navigation/filter-list.scss';
  @import '@primer/css/navigation/sidenav.scss';
  @import '@primer/css/navigation/subnav.scss';
- @import '@primer/css/navigation/underline-nav.scss';
```

### Merge checklist

- [x] Added/updated tests
- [ ] ~~Added/updated documentation~~
- [x] Added/updated previews
- [x] Visual regression test

Before | After
--- | ---
Screen Shot 2022-11-21 at 15 01 49 | Screen Shot 2022-11-21 at 15 01 56
Screen Shot 2022-11-21 at 16 52 10 | Screen Shot 2022-11-21 at 16 52 18

PR #1627, opened 2022-11-21
Move `TabNav` styles to PVC

mxriverlynn:
⚠️ WIP ⚠️ 

### Description

Describe your changes here.

Closes # (type the issue number after # if applicable; otherwise remove this line)

### Integration

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


app/components/primer/css/variables.pcss
```diff
@@ -0,0 +1,61 @@
+:root {
```
langermank, line 1: Hey @mxriverlynn! I'm not sure we want to move SCSS variables to PVC. I think they will mostly be reserved for utility CSS at this point, and the CSS that we move over from PCSS should use the new set of design tokens that are already being used in `tab_nav`. If there's an issue or discussion you can point me towards that is suggesting we should move all our existing variables to PVC, please let me know!

simurai:
Sorry.. 😩 The `TabNav` styles already got moved as part of https://github.com/primer/view_components/pull/1626.

To avoid moving the same thing twice, should we add our `@handle` next to the task in https://github.com/github/primer/issues/1342 before we start? I added all [remaining PCSS imports](https://github.com/primer/view_components/blob/main/demo/app/assets/stylesheets/application.css) for Lookbook. Only the hard ones are left. 😆 The ones with `???` at the end I'm not even sure if we should move those. Maybe we need to have a discussion if dotcom should completely drop PCSS dependency or keep it for things like color and base styles.

app/components/primer/css/variables.pcss
simurai: Migrating the SCSS variables to the CSS variables from Primitives is definitely the hardest part. 😆 

There isn't always a clear 1:1 replacement and it depends on the use case. Especially for `margin/padding`. I started a [.scss to .pcss migration guide](https://gist.github.com/simurai/2287ce592d3b1bcb2f30be8b73fac877). It's not really complete and I've just been adding to it while going along. Also might not be 100% accurate. Let me know in case something is unclear.

package.json
```diff
@@ -77,6 +77,7 @@
     "postcss-cli": "^10.0.0",
     "postcss-import": "^14.1.0",
     "postcss-mixins": "^9.0.3",
+    "postcss-nested": "^6.0.0",
```
simurai, line 80: So far we haven been adding a `& ` anywhere a nested class is used:

```diff
  .some-class {
-   .octicon {
+   & .octicon {
      ...
    }
  }
```

but I assume `postcss-nested` would allow us to keep using the "SCSS nesting" without the `& `. @jonrohan also mentioned that it might will become a CSS standard. So using `postcss-nested` might be fine?

jonrohan: We're trying to write closer to CSS spec. That said they are talking about changing the specification for nesting and there's an experimental plugin we could use to try it. https://twitter.com/css_tools_/status/1587392707206561792?s=46&t=OsB8hQIe4u2xhsK3WD3iYw

https://twitter.com/css_tools_/status/1587392697857413120?s=46&t=OsB8hQIe4u2xhsK3WD3iYw

mxriverlynn:
closed as duplicate

PR #1630, opened 2022-11-22
Move `button` styles to PVC

simurai:
### Description

This is part of [#1342](https://github.com/github/primer/issues/1342) and adds the [`button`](https://github.com/primer/css/tree/f3acb2b116f10955e1ba4f1a35aa874eed019ceb/src/navigation) styles from PCSS. There should be no visual changes.

### Integration

Yes, the [following lines](https://github.com/github/github/blob/68ac277a9c271120ebf606b6fd457b2ed8b905e5/app/assets/stylesheets/bundles/primer/index.scss#L37) can be removed on dotcom:

```diff
- @import '@primer/css/buttons/button.scss';
- @import '@primer/css/buttons/button-group.scss';
- @import '@primer/css/buttons/misc.scss';
```

### Merge checklist

- [x] Added/updated tests
- [x] Visual regression test

Before 
Screen Shot 2022-11-22 at 17 25 22

After
Screen Shot 2022-11-22 at 17 25 16


test/css/component_specific_selectors_test.rb
```diff
@@ -79,6 +79,14 @@ class ComponentSpecificSelectorsTest < Minitest::Test
       ".Button--iconOnly.Button--small",
       ".Button--iconOnly.Button--large"
     ],
+    Primer::Beta::ButtonGroup => [
+      ".BtnGroup + .BtnGroup",
+      ".BtnGroup+.BtnGroup",
+      ".BtnGroup + .btn",
+      ".BtnGroup+.btn",
```
simurai, line 86: @mxriverlynn Not sure how to make the `CSS Coverage` check pass with these `+` sibling selectors. 

[It fails](https://github.com/primer/view_components/actions/runs/3536603923/jobs/5935785853#step:7:15) with:

```
.BtnGroup+.BtnGroup
.BtnGroup+.btn
```

even when added to `IGNORED_SELECTORS`? 🤔

mxriverlynn: the ignored selectors are all treated as regex, so you would need to escape it like `/\+/` ... i think that should capture anything with a `+` in the selector

unfortunately, we don't really have a way to parse through all the selector rules for anything more than just a very basic, single selector. there may be some way to do it, but i don't know how at the moment

simurai: @mxriverlynn Thanks. 🙇 Adding `/\+/` to global resolved it and yeah, might be good to have that anyways for future sibling selectors. 👍

jonrohan (approved):

PR #1632, opened 2022-11-22
Improve performance of deny_* methods by not raising in prod environments

camertron:
### Description

It was brought to our attention that dotcom spends a lot of time in the `deny_single_argument` method. This PR is an attempt to make it faster by skipping the check in production. We follow a similar approach in other parts of the codebase to avoid raising errors in production, which we should be doing for the various `deny_*` methods as well. So, we get more consistent behavior and more performance 🤗  Speaking of performance, here's the speedup:

```
$> RAILS_ENV=production bx rails r bench.rb 
             control    14.772k i/100ms
             control    149.042k (± 5.9%) i/s -    753.372k in   5.073306s

Pausing here -- run Ruby again to measure the next benchmark...

$> RAILS_ENV=production bx rails r bench.rb 
           optimized    19.347k i/100ms
           optimized    198.415k (± 3.4%) i/s -      1.006M in   5.076516s

Comparison:
           optimized:   198415.1 i/s
             control:   149042.0 i/s - 1.33x  (± 0.00) slower
```

I'm seeing an improvement in the 30-35% range.

### Integration

No.

### Merge checklist

- [x] Added/updated tests


jonrohan (approved):
Not blocking, but I was wondering if we could solidify this performance gain somehow in a test in `test/performance/`

camertron:
@jonrohan that's a great idea. Maybe we could add a test that ensures the prod version is faster than the non-prod version. Let me see what I can do.

PR #1638, opened 2022-11-25
Move `FormControl` styles to PVC

simurai:
### Description

This is part of [#1342](https://github.com/github/primer/issues/1342) and adds the [`FormControl`](https://github.com/primer/css/blob/aee4b6f571d88f391fcf98170857c4eed7b1ae82/src/forms/FormControl.scss) styles from PCSS. There should be no visual changes.

### Integration

Yes, the [following lines](https://github.com/github/github/blob/985ee396bf03faf4c2cd43d0d6fc1ef9248a46cf/app/assets/stylesheets/bundles/primer/index.scss#L53) can be removed on dotcom:

```diff
- @import '@primer/css/forms/FormControl.scss';
```

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews
- [x] Visual regression test

Not sure why the `Primer::Beta::AutoComplete` placeholder text is lighter in the [snapshots](https://github.com/primer/view_components/pull/1638/files?file-filters%5B%5D=.png&show-viewed-files=false). Locally they seem to look the same:

Before | After
--- | ---
Screen Shot 2022-11-28 at 15 53 11 | Screen Shot 2022-11-28 at 15 53 18


postcss.config.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
       features: {
         'nesting-rules': true,
         'focus-visible-pseudo-class': false,
+        'logical-properties-and-values': false,
```
simurai, line 18: Had to disable `logical-properties-and-values` to make properties like `padding-inline-start` work.

When set to `true`, the output adds `[dir="ltr"]` + `[dir="rtl"]` to the selector, I assume as a fallback. But we currently don't use these attributes in Lookbook/dotcom.

test/css/component_specific_selectors_test.rb
```diff
@@ -39,6 +39,16 @@ class ComponentSpecificSelectorsTest < Minitest::Test
...
     ],
+    Primer::Alpha::TextField => [
+      ".FormControl-inlineValidation p",
+      ".FormControl-select",
+      ".FormControl-textarea",
+      ".FormControl-input",
+      ".FormControl-input-wrap",
+      ".FormControl-select-wrap",
+      ".FormControl-checkbox-wrap",
+      ".FormControl-radio-wrap"
```
simurai, line 50: Currently all `.FormControl` styles are part of `Primer::Alpha::TextField` and located at `app/components/primer/alpha/text_field.pcss`. But the `.FormControl` styles are not just for text fields and also include styles for select, radio, checkbox etc. so not sure if that should be split up?

Or we could also move `.FormControl` to a place where we have styles that get shared across multiple components? Maybe [app/lib/primer/css/](https://github.com/primer/view_components/tree/main/app/lib/primer/css).

keithamus (approved):

PR #1643, opened 2022-11-28
Fix disabled ToggleSwitch behavior

camertron:
### Description

We received a bug report via Slack today showing that disabled `ToggleSwitch`es can still be toggled. We have tests to verify this behavior that are currently passing erroneously. This PR accomplishes the following:

1. Fixes the tests in question.
2. Disallows toggling disabled `ToggleSwitch`es.
3. Fixes a bug causing the wrong on/off state to be submitted on toggle. The form should send a 0 if the switch is being toggled from on -> off, and a 1 if it's being toggled from off -> on.

### Integration

No. The `ToggleSwitch` component isn't being used anywhere in prod.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews

Co-authored with @neall


jonrohan (approved):
LGTM

PR #1650, opened 2022-11-29
add relative-time component

keithamus:
### Description

This adds a new `RelativeTime` component, which replaces the `LocalTime` and `TimeAgo` components.

It also _changes_ the `LocalTime` and `TimeAgo` components to use the `relative-time` web component, as they currently use `local-time` and `time-ago` components, which themselves are deprecated as they are just configurations over the `relative-time` component.

### Integration

Currently no, but we want to deprecate `LocalTime` and `TimeAgo` and switch them over to `RelativeTime`.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


jonrohan (approved):
Looking great

PR #1672, opened 2022-12-05
forms immediate server side validation

neall:
### Description

This adds optional client-side functionality to Forms framework text fields so they can validate against the server as the user types.

immediate-validation

I had to touch a lot of lines in `form_control.html.erb`, so I wrapped that in some more tests. There's still a lot of logic in that file but I moved some of it into other places where it is easier to read.

In particular that file is where we render the error-message-related elements. We normally don't render them if there is no error on the field, but if we are using auto-check and updating the client we need to render those elements hidden.

### Integration

This won't require any changes to existing forms; nothing changes unless the developer specifies a URL for `auto_check_src` on their text field.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


test/lib/primer/forms/deep_thought.rb
```diff
@@ -0,0 +1,17 @@
+class DeepThought
```
camertron, line 1: Can we stick this maybe into a models/ directory?

test/lib/primer/forms/forms_test.rb
```diff
@@ -54,6 +55,51 @@ def test_the_input_is_described_by_the_caption
     assert_selector "input[aria-describedby='#{caption_id}']"
   end
 
+  def test_includes_activemodel_validation_messages
```
camertron, line 58: I moved these to test/lib/primer/forms/form_helper_test.rb, so they can be removed from here.

I think we also want to `require "deep_thought"` in test/lib/primer/forms/form_helper_test.rb and remove the `DeepThought` class.

camertron (changes requested):
Yasss! 🎉  Just a few minor things.

lesliecdubs:
Oh, this is neat! Was there an issue tied to this effort? Curious if we involved Primer accessibility design in the decisions about how the client-side validations appear in the form.

lib/primer/forms/primer_text_field.ts
```diff
@@ -0,0 +1,42 @@
...
+declare global {
+  interface Window {
+    PrimerTextFieldElement: typeof PrimerTextFieldElement
+  }
+}
+
+if (!window.customElements.get('primer-text-field')) {
+  Object.assign(window, {PrimerTextFieldElement})
+  window.customElements.define('primer-text-field', PrimerTextFieldElement)
+}
```
keithamus, line 42: This won't be needed because of the `@controller` decorator.

lib/primer/forms/primer_text_field.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  connectedCallback(): void {
+    this.inputElement.addEventListener('auto-check-success', () => { this.clearError() })
+
+    this.inputElement.addEventListener('auto-check-error', (event: any) => {
+      event.detail.response.text().then((error_message: string) => {
+        this.setError(error_message)
+      })
+    })
+  }
```
keithamus, line 18: For robustness, here it might be good to use an [AbortSignal so events can be correctly torn down](https://catalyst.rocks/guide/patterns/#registering-global-or-many-event-listeners):

```suggestion
  #abortController: AbortController | null
  
  connectedCallback(): void {
    this.#abortController?.abort()
    const signal = (this.#abortController = new AbortController())
    this.inputElement.addEventListener('auto-check-success', () => { this.clearError() }, { signal })

    this.inputElement.addEventListener('auto-check-error', (event: any) => {
      event.detail.response.text().then((error_message: string) => {
        this.setError(error_message)
      })
    }, { signal })
  }
  
  disconnectedCallback() {
    this.#abortController.abort()
  }
```

camertron:
@lesliecdubs good questions! There's no tracking issue to my knowledge. This work came from one of the settings page forms @neall's been working on. It so happens I also need it for the billing upgrade form.

> Curious if we involved Primer accessibility design in the decisions about how the client-side validations appear in the form.

We have not explicitly sought their input, however the markup and behavior adheres to the existing [form interface guidelines](https://primer.style/design/ui-patterns/forms). I believe they have been a11y reviewed, although I may be wrong about that.

In any case, we do need to have someone from a11y look at this. We probably need to announce the validation message somehow.

PR #1675, opened 2022-12-06
move `Primer::LocalTime` to `Primer::Alpha::LocalTime`

mxriverlynn:
### Description

this PR moved `Primer::LocalTime` to `Primer::Alpha::LocalTime`, and deprecates the original. 

it also makes several adjustments to the `component_status_migrator` thor script:
* handle inconsistency between preview file name and component file name / class name
* move all the misc files that come with a component (javascript, css, etc)
* correct the `primer.ts` file imports

### Integration

no immediate changes are required, but component deprecations will need to be taken care of

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


keithamus:
We may want to move this all the way to `Deprecated` (or remove it?) as `RelativeTime` will replace this.

PR #1688, opened 2022-12-07
Update tooltip to run setup on connectedCallback instead of relying on attributeChangedCallback

manuelpuyol:
### Description

It looks like when we navigate using a Turbo Frame in dotcom, due to some Turbo internals,`attributeChangedCallback` **runs before the element is added to the DOM tree**. This causes `control` to not be found (since it's also not in the tree) and the setup is skipped. In this case, `control` will not have the correct `aria-` attributes.

With this change, I'm guaranteeing that the setup **always** runs when the `tool-tip` connects to the page. 

### Integration

Nope, this should be a no-op in terms of functionality

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/.changeset/new-dancers-train.md
+++ b/.changeset/new-dancers-train.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Fix tooltips not adding aria attributes when Turbo navigating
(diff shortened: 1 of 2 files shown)


camertron (approved):

khiga8 (approved):
Nice fix and cleanup!

In the past, we've run into issues with `#updatePosition` method being called on page load which led to performance issues on pages with hundreds of tooltips (since calculating position is expensive). It doesn't seem like this PR introduces that issue so this seems fine to me.

PR #1691, opened 2022-12-08
Start migration to new slots API

joelhawksley:
### Description

This PR starts our migration to use the new `with_` slots API as per https://viewcomponent.org/CHANGELOG.html#2640.

I've added components that don't use slots into the new behavior along with TimelineItem, Subhead, and Layout.


camertron (approved):
Awesome, thank you 🎉

jonrohan (approved):

PR #1693, opened 2022-12-08
Use URL helpers in previews that interact with the server-side

camertron:
### Description

Some of our previews don't work in the prod Lookbook because the URLs are hard-coded and only work in dev.

### Integration

No.

### Merge checklist

- [ ] ~Added/updated tests~
- [x] Added/updated previews

### DIFF
--- a/.changeset/rude-toes-play.md
+++ b/.changeset/rude-toes-play.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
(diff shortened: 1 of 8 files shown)


jonrohan:
Nice! I think we have similar problems with the autocomplete component

jonrohan:
It doesn't seem to work on http://primer-view-components-preview-1693.eastus.azurecontainer.io/

jonrohan (approved):

PR #1695, opened 2022-12-09
Remove `forms.css` import

simurai:
### Description

Follow up to https://github.com/primer/view_components/pull/1638. This removes the `forms.css` import.

### Reasoning

Currently the `FormControl` styles are loaded twice in Lookbook:

Screen Shot 2022-12-09 at 17 10 15

1. Part of [`Primer::Alpha::TextField`](https://github.com/primer/view_components/blob/ee7f476dd13b58e1c38538950780493b97142c8b/app/components/primer/alpha/text_field.pcss)
2. Part of a PVC > PCSS > PVC round trip through the [PCSS forms bundle](https://github.com/primer/css/blob/7f48bcf07c3b57299daf4530534477c756647bbb/src/forms/index.scss#L9)

The problem is that if we remove some property in PVC, e.g. the `height`, it will still be there from the second import. This could be pretty confusing. "I just removed it, why is it still there?" 🤔

So this PR 🔥 removes the `forms.css` import that contains the 2nd `FormControl` styles.

### Consequences

Merging this PR will have the effect that the `Primer::Alpha::AutoComplete` in Lookbook will be un-styled. It's not great, but since that component is deprecated and replaced by the newer `Primer::Beta::AutoComplete`, maybe ok. 🤷 It might even help in preventing people from using the old component. 😉 

Component | Before | After
--- | --- | ---
`Primer::Alpha::AutoComplete` | Screen Shot 2022-12-09 at 17 07 14 | Screen Shot 2022-12-09 at 17 08 41
`Primer::Beta::AutoComplete` | Screen Shot 2022-12-09 at 17 10 15 | Screen Shot 2022-12-09 at 17 11 26

### Alternatives

There are a couple options:

1. Add a new `forms-rails` bundle to PCSS and use it for PVC. 👉  [Draft PR](https://github.com/primer/css/pull/2344). **Downside**: Feels a bit overkill and yet another thing to clean up later.
2. Remove the `FormControl.scss` import from the PCSS `forms/` bundle without making a major version. It's pretty new and was never documented. **Downside**: Someone might already uses the `FormControl` styles.
3. **This PR**: In PVC we could just remove the [`forms.css`](https://github.com/primer/view_components/blob/ee7f476dd13b58e1c38538950780493b97142c8b/demo/app/assets/stylesheets/application.css#L8) import. It's only needed for the deprecated `Primer::Alpha::AutoComplete`. **Downside**: `Primer::Alpha::AutoComplete` will look un-styled in Lookbook.
4. Do nothing (for now). **Downside**: If we remove something in PVC's `FormControl`, e.g. `margin` it will still be present from PCSS's `FormControl`. It's not a problem on dotocm (separate imports), just when debugging in Lookbook we'd have to keep it in mind.

None seem great. But personally I think **option 3** (this PR) is the best compromise.

### Integration

No, it only affects Lookbook

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


jonrohan (changes requested):
Seems like the color changed in the visual difference?

camertron (approved):
I'm cool with this, I think your reasoning is sound. @jonrohan could the color changes be related to the fact that the alpha version of the auto complete component is now basically unstyled?

jonrohan (approved):
Ah, I didn't read the comment body well enough on my phone

simurai:
> @jonrohan Seems like the color changed in the visual difference?
> Ah, I didn't read the comment body well enough on my phone

Yes, unfortunately it's a compromise and the old `Primer::Alpha::AutoComplete` will look un-styled in Lookbook. But I think out of the 4 possible compromises, still the best option. 🤷

PR #1700, opened 2022-12-09
Fix a test that fails in non-UTC TZs

neall:
### Description

This test was passing in CI (which must be in UTC), but when run on developer machines in other time zones it would fail.

To fix it, we changed the test from making a Time object and converting to UTC to have it make a Time object in UTC from the beginning.

### Merge checklist

- [x] Added/updated tests


keithamus (approved):
Ahh cool!

PR #1703, opened 2022-12-12
Beta components use new Slots API

joelhawksley:
### Description

I've updated most of the Beta folder to use the new slots API. I took the liberty of standardizing some syntax along the way.

### DIFF
--- a/.changeset/thick-numbers-beam.md
+++ b/.changeset/thick-numbers-beam.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+Beta components use new `with_*` Slots API.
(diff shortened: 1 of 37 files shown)


ollie-iterators:
I believe that the file "progress_bar_test.rb" needs to be changed to use the new Slots API

joelhawksley:
> I believe that the file "progress_bar_test.rb" needs to be changed to use the new Slots API

Correct. I'll do more of these in another PR ❤️

ollie-iterators:
I believe that the file "progress_bar_test.rb" is the only file in the beta test file that needs to be updated.

camertron (approved):

PR #1707, opened 2022-12-13
Using the primer/css actionlist import in lookbook

jonrohan:
### Description

I'm not sure I ever had this imported properly, but this fixes an issue with the lookbook autocomplete styling.

image
https://primer.style/view-components/lookbook/inspect/primer/beta/auto_complete_preview/playground/

### Integration

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


simurai (approved):
I think this is fine to temporarily fix `AutoComplete`.

Downside is that Lookbook will have some of the `ActionList` styles twice:

Screen Shot 2022-12-14 at 11 26 16

So maybe something to keep in mind when making changes to `ActionList` in PVC and testing in Lookbook.

PR #1709, opened 2022-12-14
Fix doc examples so server-side components like AutoComplete and ToggleSwitch work as expected

camertron:
### Description

All our doc examples that interact with the server are broken because they use invalid routes. This PR updates them. Note that this change will not appear to work in preview environments - it will only work in production.

### Integration

No

### Merge checklist

- [ ] ~Added/updated tests~


jonrohan (approved):
Seems good 👍🏻

PR #1715, opened 2022-12-16
ToggleSwitch form

neall:
### Description

This creates a `Primer::Forms::ToggleSwitchForm` that will just have one auto-self-submitting ToggleSwitch.

image

Developers can render a direct instance of this class in a template or they can inherit from it in order to bake in some of the parameters. See the comments in [lib/primer/forms/toggle_switch_form.rb](https://github.com/primer/view_components/compare/neall/toggle-switch-form?expand=1#diff-cd2352bf969210bb5993921f05b43828b8d02aa88d409a550b05eff530877048) for examples.

### Integration

Nothing is using this yet.

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/.changeset/lovely-boats-hang.md
+++ b/.changeset/lovely-boats-hang.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Add Primer::Forms::ToggleSwitchForm
--- a/app/forms/example_toggle_switch_form.rb
+++ b/app/forms/example_toggle_switch_form.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class ExampleToggleSwitchForm < Primer::Forms::ToggleSwitchForm
+  set_name :example_field
+  set_label "Example"
+end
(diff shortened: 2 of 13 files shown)


ollie-iterators:
Shouldn't the module for this component be something like Alpha or Beta? I don't think it should be Form.

neall:
@ollie-iterators The `Primer::Forms::ToggleSwitch` class is only for internal use by the Forms framework and is not the same as the `Primer::Alpha::ToggleSwitch` component which is where you expect it to be.

The fact that `Primer::Forms::ToggleSwitch` and many other parts of the Forms framework are implemented as subclasses of `Primer::BaseComponent` is a bit confusing. They are left undocumented because they're meant to be internal, but that doesn't help in the context of PRs like this.

The Forms framework in general is not namespaced with the individual components. That's why the new `Primer::Forms::ToggleSwitchForm` class we create in this PR is at that namespace: it's a sibling of the `Primer::Forms::Base` class that it inherits from. I am hesitant to suggest changing that.

camertron (approved):
🎉

PR #1717, opened 2022-12-19
Make tool_tip Trusted Types compatible

KyFaSt:
### Description

This change makes primer view_components compatible with the [CSP directive Trusted Types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/trusted-types). This CSP directive allows developers to mark a value as a Trusted Type, usually this would be done in conjunction with running some type of sanitizer like DOMPurify to ensure the value doesn't contain any unsafe elements. Fortunately, view_components doesn't have major violations, just this one but unfortunately the change in this PR does not buy any security benefits, it just adheres to the Trusted Types API -- not passing bare strings directly to potentially dangerous injection sinks. Currently this implementation is the best way to make this library compatible with trusted types.

### Integration

No, this should stay functionally equivalent.

### Merge checklist

I omitted these because I don't think this change should affect tests or documentation

~- [ ] Added/updated tests~


camertron (commented):
Hey @KyFaSt, thanks for this! Would you mind adding a changeset? Just run `npx changeset` and follow the prompts (we only bump the patch version for PVC right now).

jonrohan:
@keithamus @khiga8 I'm curious if there was any reason we did innerHTML here before?

keithamus (approved):
`innerHTML` is just a bit neater. This change is 👍 though

PR #1728, opened 2022-12-22
Change `ActionList::Item` tag to `button`

langermank:
### Description

To be honest, I'm not entirely sure why it was a `span`. This could be a huge mistake. I'm hoping someone else will know! 🙌 

If this becomes a `button`, it makes it much easier to use for triggering our new `Dialog` component (which is an extremely common use case.)

### Integration

No

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/app/components/primer/alpha/action_list/item.rb
+++ b/app/components/primer/alpha/action_list/item.rb
@@ -48,7 +48,7 @@ module Primer
             Primer::BaseComponent.new(tag: :svg, width: "16", height: "16", **system_arguments)
           },
           content: lambda { |**system_arguments|
-            Primer::BaseComponent.new(tag: :span, **system_arguments)
+            Primer::BaseComponent.new(tag: :button, **system_arguments)
           }
         }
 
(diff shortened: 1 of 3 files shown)


app/components/primer/alpha/action_list/item.rb
```diff
@@ -48,7 +48,7 @@ class Item < Primer::Component
             Primer::BaseComponent.new(tag: :svg, width: "16", height: "16", **system_arguments)
           },
           content: lambda { |**system_arguments|
-            Primer::BaseComponent.new(tag: :span, **system_arguments)
+            Primer::BaseComponent.new(tag: :button, **system_arguments)
```
camertron, line 51: Is this change necessary? The user could easily provide a button of their own here.

camertron (commented):
Nice! I think this is fine given it's how the old `NavigationList` works.

camertron (approved):

PR #1731, opened 2022-12-24
Don't export PrimerMultiInputElement

safeforge:
Exporting a component can cause errors when used as an npm package. Please see #1560

/cc @jonrohan

### DIFF
--- a/.changeset/dull-rabbits-glow.md
+++ b/.changeset/dull-rabbits-glow.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Avoid double-registering of exported components
--- a/lib/primer/forms/primer_multi_input.ts
+++ b/lib/primer/forms/primer_multi_input.ts
@@ -2,7 +2,7 @@
 import {controller, targets} from '@github/catalyst'
 
 @controller
-export class PrimerMultiInputElement extends HTMLElement {
+class PrimerMultiInputElement extends HTMLElement {
   @targets fields: HTMLInputElement[]
 
   activateField(name: string) {


jonrohan:
Is there some kinda linting we could introduce to avoid this in the future?

camertron:
Just popping in here to say that this time I actually need to reference the `PrimerMultiInputElement` class in another file, so it has to be exported. Last time it didn't matter because I didn't need to use `ToggleSwitchElement` anywhere. I didn't dig into why the error reported by @safeforge in #1440 occurs, so I don't understand it at all. @keithamus mentioned [here](https://github.com/primer/view_components/pull/1560#pullrequestreview-1163139912) that it should be possible to export custom elements, perhaps he can chime in here. It probably has something to do with how Catalyst manages custom elements.

safeforge:
Thank you for explaining. I had the chance to investigate another way to use the library as an npm and seems to work properly. Instead of using `import '@primer/view-components';` I tried with full path and I'm able to compile and use all components. 

```patch
diff --git a/app/javascript/application.js b/app/javascript/application.js
index 524536c..c4f2aa4 100644
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,3 +1,3 @@
 import '@hotwired/turbo-rails';
-import '@primer/view-components';
+import '@primer/view-components/app/assets/javascripts/primer_view_components.js';
 import './stimulus.js';
```

Removing the `module`  from `package.json` is possible to import `@primer/view-components` without errors and I'm wondering if it's necessary to keep both `main` and `module`

```patch
diff --git a/package.json b/package.json
index 5e8086f7..4ad29a34 100644
--- a/package.json
+++ b/package.json
@@ -3,7 +3,6 @@
   "version": "0.0.116",
   "description": "ViewComponents for the Primer Design System",
   "main": "app/assets/javascripts/primer_view_components.js",
-  "module": "app/components/primer/primer.js",
   "types": "app/components/primer/primer.d.ts",
   "repository": "primer/view_components",
   "keywords": [
```

safeforge:
Thank you for your support, I'm going to close the issue since importing the full path seems to work properly.

writercoder:
I've opened a PR with an update to the install documentation as I spent a while chasing this one down.

PR #1732, opened 2022-12-27
Update markdownlinting packages and configs

khiga8:
### Description

- Introduce `markdownlint-github` and configure rules.
- Swap `markdownlint-cli` to `markdownlint-cli2` which is **compatible** with the [markdownlint VSCode extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint). With `markdownlint-cli`, the extension won't respect the configs that are set.
- Add [markdownlint VSCode extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) to `.devcontainer.json` so every new codespace comes with it installed.

### Integration

No

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


khiga8:
This PR will introduce the `no-generic-link-text` but I realized that it will falsely flag references to the `[Link]` component. `markdownlint` does support disable comments but due to how markdown files are generated in PVC, this doesn't seem like a viable path since the disable would just get overwritten.

Maybe we can have all component links have component as part of the name like `Read more at [Blankslate component]` or `Learn more at [Link component]`... but also I'm not really sure about this... I don't want to make an executive decision on this now.

The option I'm leaning towards is to just let exceptions be defined for the rule in the markdown config, so `link` will be specified as an exception. I opened a PR here which I'd appreciate a review on: [Staff-only: markdownlint-github PR](https://github.com/github/markdownlint-github/pull/27).

Let me know if you have additional thoughts: @primer/rails-reviewers.

.markdownlint-cli2.cjs
```diff
@@ -0,0 +1,13 @@
+const options = require('@github/markdownlint-github').init({
+  "line-length": false,
+  "ol-prefix": false,
+  "no-trailing-punctuation": false,
+  "no-inline-html": false,
+  "first-line-heading": false,
+  // Rules that we want to enable
+  "no-duplicate-heading": false
```
khiga8, line 8: I think we'll want to eventually enable this `no-duplicate-heading` rule. 

With Primer docs, anchor links are generated from the headings. When headings are duplicated on a page, the generated anchor links will always go to the first instance of the heading. 

One example is [#autocorrectable heading](https://primer.style/view-components/linting#autocorrectable) which appears twice on different sections of the page, and share the same anchor link.

jonrohan:
Nice! We have our lint system setup so it won't run unless there's a change to a markdown file. You should test this new config by adding an error somewhere.

khiga8:
> Nice! We have our lint system setup so it won't run unless there's a change to a markdown file. You should test this new config by adding an error somewhere.

Here's [failed run](https://github.com/primer/view_components/actions/runs/3788886862/jobs/6442115543) after introducing an error.

Does https://github.com/primer/view_components/pull/1732#issuecomment-1366020678 sound reasonable to you?

jonrohan (approved):
> Does [#1732 (comment)](https://github.com/primer/view_components/pull/1732#issuecomment-1366020678) sound reasonable to you?

Yep sounds good to me 👍🏻

khiga8:
I am not authorized to merge. This is ready! Please merge freely.

jonrohan (approved):

PR #1748, opened 2023-01-03
Use published version of kuby-core

camertron:
### Description

Dependabot [reminded me](https://github.com/primer/view_components/pull/1740) that kuby-core is pinned to a specific commit hash in our bundle. I have since upstreamed all the necessary code we needed for PVC, so we should be able to switch to a published version of the gem now.

### Integration

No.

### Merge checklist

~- [ ] Added/updated tests~


jonrohan (approved):
Seems good to go 👍🏻

camertron:
I had to upgrade cert-manager, but now everything seems a-ok.

PR #1750, opened 2023-01-05
Adding :auto to size options to Dialog component

jonrohan:
### Description

I think it's possible we're missing more than just auto? I saw a few other sizes in the css that didn't seem to be in the component. eg. [Overlay--size-full](https://github.com/primer/view_components/blob/a8b8e95ab1116548a116124aa2dc25eb98d19768/app/components/primer/alpha/dialog.pcss#L32)

[Slack thread](https://github.slack.com/archives/CSGAVNZ19/p1672933074645839)

### Integration

No

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/.changeset/selfish-countries-vanish.md
+++ b/.changeset/selfish-countries-vanish.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+Adding `:auto` to size option to Dialog component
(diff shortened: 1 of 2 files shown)


keithamus (approved):
LGTM

keithamus:
Actually @langermank might have some commentary here. I feel like there's a reason why this isn't already in here.

langermank (approved):
Lets not add the `full` as a param since its reserved for small screens (responsive)

PR #1753, opened 2023-01-06
upgrade button component in dialog

keithamus:
### Description

[Per slack (staff only link)](https://github.slack.com/archives/C02PTFZQ5LZ/p1672866608190239?thread_ts=1672866608.190239&cid=C02PTFZQ5LZ)

> It looks like the `Primer::Alpha::Dialog` uses the method with_show_button which generates the deprecated `Primer::ButtonComponent` and it has different CSS classes than `Primer::Beta::Button`
Will the `Primer::Alpha::Dialog` component be updated to show the `Primer::Beta::Button button`?

### Integration

We will need to check Dialog show_button slots for incompatible system arguments from the old button 

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


jonrohan (approved):
One thing we'll need to search for here is all the uses of the Dialog component and check that the `with_show_button` slot isn't passing arguments that are deprecated. https://primer.style/view-components/guides/primer_button_component

PR #1754, opened 2023-01-06
Use a mouseup mousedown event to guard accidental clicks dismissing dialog

keithamus:
### Description

From https://github.com/github/primer/issues/1628 (staff only)

> [we] noticed that if a user selects text, drags the cursor outside of the modal and then releases the mouse button, the dialog closes as if they had clicked outside of the dialog.

This fixes the problem by  using a `mousedown` event listener to track that a click starts out on the backdrop, and then registering a onetime mouseup listener to ensure the mouse is still over the backdrop. 

With this pattern we can track click events that happen only where the backdrop was the element initially part of the mousedown and part of the mouseup, this prevents:

- Dismissing the dialog if the click started inside the dialog but ended up on the backdrop
- Dismissing the dialog if the click started on the backdrop but the user intends to "undo" their click and moves the mouse inside the dialog.

Closes https://github.com/github/primer/issues/1628

### Integration

### Merge checklist

- [ ] Added/updated tests
- [x] Added/updated previews

### DIFF
--- a/.changeset/witty-mails-attend.md
+++ b/.changeset/witty-mails-attend.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Fix bug where clicking inside dialog, but finishing click outside, dismisses dialog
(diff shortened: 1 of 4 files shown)


jonrohan (approved):
Just preliminary looking at this on lookbook confirmed the problem is fixed. Is there any tradeoffs to using this vs the previous way?

keithamus:
It might be worth making sure that touch devices still work, but other than that I cannot think of any trade offs with this solution.

jonrohan:
Checked on ios safari works 👍🏻

PR #1764, opened 2023-01-18
Use a bash script to check for changesets

camertron:
### Description

I'm seeing the Triage / Check for changeset job fail from time to time. The brettcannon/check-for-changed-files action we're using makes requests to the GitHub API to find the list of changed files and occasionally runs into a rate limit. I thought we could probably achieve the same thing with some regular 'ol git commands. How hard could it be? As it turns out, quite difficult. The script I added does the following:

1. Fetches `main` (or whatever the base branch is)
2. Fetches HEAD commits until git can find a merge base between the base branch and HEAD.
3. Examines the diff and `grep`s for .md files in the .changeset directory.

### Integration

No

### Merge checklist

~- [ ] Added/updated tests~

### DIFF
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -29,11 +29,12 @@ jobs:
   check-for-changeset:
     name: Check for changeset
     runs-on: ubuntu-latest
+    env:
+      SKIP_LABEL: "skip changelog"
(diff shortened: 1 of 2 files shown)


jonrohan (approved):
Nice improvement 🚀

PR #1773, opened 2023-01-24
Add a max-width to `Banner`

simurai:
### Description

This adds a `Banner--centered` class when the banner is `full` with and the `content` or `description` is over `120` characters long. It makes sure that the text starts to wrap onto multiple lines and doesn't get too long.

👀  [Lookbook preview](http://primer-view-components-preview-1773.eastus.azurecontainer.io/view-components/lookbook/inspect/primer/alpha/banner/full_width)

#### Before

Screen Shot 2023-03-10 at 14 35 25

#### After

Screen Shot 2023-03-10 at 14 34 30

Closes https://github.com/github/primer/issues/1697

### Integration

No

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


langermank:
@simurai based on what you show in your Loom, I really like the solution described here https://github.com/github/primer/issues/1524. I'm curious how feasible it is though, since it depends on the updated header design.

simurai:
> I really like the solution described here [github/primer#1524](https://github.com/github/primer/issues/1524). I'm curious how feasible it is though, since it depends on the updated header design.

Yeah, maybe it needs a few more explorations, to see how it feels. Also with multiple Banner stacked on top of each other.

.changeset/kind-humans-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+Add a max width to Banner content
```
simurai, line 5: Just a test to trigger the banner.

simurai (commented):
Just a test to trigger the banner.

simurai (commented):
Another test

app/components/primer/alpha/banner.rb
```diff
@@ -104,7 +104,8 @@ def initialize(full: false, full_when_narrow: false, dismissible: false, descrip
           LEGACY_SCHEME_MAPPINGS[@scheme],
           "Banner--full": full,
           "flash-full": full, # legacy
-          "Banner--full-whenNarrow": full_when_narrow
+          "Banner--full-whenNarrow": full_when_narrow,
+          "Banner--centered": full && description && description.length > 120 # center with max-width for long descriptions
```
simurai, line 108: I couldn't figure out how to do the same for `content`. Doing:

```
full && content && content.length > 120
```

results in the content not rendering? Is it because content can contain HTML? 🤔

camertron: @simurai I think that's probably because `content` isn't available until render time. Calling it in the constructor is probably causing a `nil` or blank string to be cached for the content, meaning any subsequent calls to `#content` will return `nil` too. Can you add this class in `before_render` instead?

simurai: > Can you add this class in `before_render` instead?

Yes, that did the trick. Thanks. 🙇

app/components/primer/alpha/banner.rb
```diff
@@ -121,6 +125,14 @@ def initialize(full: false, full_when_narrow: false, dismissible: false, descrip
 
       private
 
+      def before_render
+        @system_arguments[:classes] = class_names(
+          @system_arguments[:classes],
+          # Center with max-width for a lot of content or long descriptions
+          "Banner--centered": @full && content && content.length > 120 || @full && @description && @description.length > 120
```
jonrohan, line 132: One concern with checking length vs explicitly setting it, is if someone adds breakpoints on their own to the banner. We all anything to be put into the content of the banner, what if it's something like this:

```erb
      <%= render(Primer::Alpha::Banner.new) do |component| %>
        <% component.with_action_button(size: :small) { "Take action" } %>
        This is a banner with an action button!<br/>
        We sometimes have banners with a lot of text.<br/>
        But in this case I'm breaking the text to make sure the line lengths are small.<br/>
        This can be easily done with the `<br>` tag.
      <% end %>
```

simurai: Hmm.. Manually adding line breaks with `<br>` would behave like:

### Before 
Screen Shot 2023-03-13 at 12 24 31

### After
Screen Shot 2023-03-13 at 12 24 42

So yeah, it would make the content centered unrelated to the line length. Not sure if that's ok.

I think in general, it would be nice to avoid adding line breaks manually. Otherwise on narrow screens it could look weird once the the short lines start wrapping and create these gaps in between:

Screen Shot 2023-03-13 at 12 16 38

I thought about adding a `centered` option that you can set manually. But that comes with the risk of people using it also for short lines.

simurai:
Gonna close this as `wontfix`. As a workaround, using `<br>` should be ok too, see https://github.com/primer/view_components/pull/1773#discussion_r1132639345.

We can re-open it if it comes up more often.

PR #1776, opened 2023-01-24
Adding aria labeled by and described by check to SegmentedControl component

jonrohan:
### Description

This adds examples of including Subhead and Label components with the control to demonstrate proper labeling. I also added a new check that will make sure a component has at least one of `aria-label, aria-labeledby, aria-describedby`

Closes # https://github.com/github/primer/issues/1149

### Integration

We should check the existing uses of SegmentedControl to make sure they are using an aria-label

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/app/components/primer/component.rb
+++ b/app/components/primer/component.rb
@@ -55,6 +55,13 @@ module Primer
       raise ArgumentError, "`aria-label` is required." if aria_label.nil? && !Rails.env.production?
     end
 
+    def validate_aria_labeled
+      return if Rails.env.production?
+
+      aria_label = aria("label", @system_arguments) || aria("labelledby", @system_arguments) || aria("describedby", @system_arguments)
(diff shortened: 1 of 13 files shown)


app/components/primer/component.rb
```diff
@@ -55,6 +55,13 @@ def validate_aria_label
       raise ArgumentError, "`aria-label` is required." if aria_label.nil? && !Rails.env.production?
     end
 
+    def validate_aria_labeled
+      return if Rails.env.production?
+
+      aria_label = aria("label", @system_arguments) || aria("labelledby", @system_arguments) || aria("describedby", @system_arguments)
```
mperrotti, line 61: A label is always required, so we should check for `aria-label` or `aria-labelledby`. I see we're already checking for `aria-label` on line 55, so maybe we just move the label/labelledby checks to that block?

`aria-describedby` is only required when a caption has been passed.

camertron (approved):

PR #1786, opened 2023-01-30
Use lookbook 1.5.1

joelhawksley:
### Description

This PR updates our Lookbook dependency to 1.5.1 which includes a JSON endpoint we can use for Axe analysis.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -215,7 +215,7 @@ GEM
     tzinfo (2.0.6)
       concurrent-ruby (~> 1.0)
     unicode-display_width (2.4.2)
-    view_component (2.82.0)
+    view_component (3.0.0.rc1)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -212,10 +212,10 @@ GEM
       unicode-display_width (>= 1.1.1, < 3)
     thor (1.2.1)
     timecop (0.9.6)
-    tzinfo (2.0.5)
+    tzinfo (2.0.6)
       concurrent-ruby (~> 1.0)
     unicode-display_width (2.4.2)
-    view_component (3.0.0.rc1)
+    view_component (2.82.0)
```
jonrohan, line 218: Does this matter?

joelhawksley: It does! good catch.

jonrohan (approved):

PR #1805, opened 2023-02-02
Remove duplicate `aria-label` in `CloseButton`

paulcsmith:
### Description

The `CloseButton` had an issue where using `aria: { label: "foo" }` would result in multiple `aria-label` attributes in the generated HTML.

This fixes the issue by using the `aria` helper method that will check for nested hash keys or a String aria-label

### Integration

No. This should work as-is

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/app/components/primer/beta/close_button.rb
+++ b/app/components/primer/beta/close_button.rb
@@ -28,7 +28,7 @@ module Primer
           "close-button",
           system_arguments[:classes]
         )
-        @system_arguments[:"aria-label"] ||= "Close"
+        @system_arguments[:"aria-label"] = aria("label", system_arguments) || "Close"
       end
 
       def call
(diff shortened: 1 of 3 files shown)


app/components/primer/beta/close_button.rb
```diff
@@ -28,7 +28,7 @@ def initialize(type: DEFAULT_TYPE, **system_arguments)
           "close-button",
           system_arguments[:classes]
         )
-        @system_arguments[:"aria-label"] ||= "Close"
+        @system_arguments[:"aria-label"] = aria("label", system_arguments) || "Close"
```
paulcsmith, line 31: The `aria` method gets the attribute from either a String or a nested hash:

https://github.com/primer/view_components/blob/79d2b02d264410e82df3711d05939cfb15b9efef/app/components/primer/component.rb#L49-L51

jonrohan (approved):
Nice catch

PR #1806, opened 2023-02-02
Add group_item option to Beta::Button

paulcsmith:
### Description

Adds a `group_item` option to `Primer::Beta::Button`.

**Background:**

The old `ButtonComponent` had a `group_item` option so that the button could be used with a button group. Recently this component was deprecated so we switched to using `Beta::Button`. We didn't realize that the `group_item` option was no longer supported and there was no error generated when we switched components so the generated HTML was not as expected and caused a visual bug: 

Screenshot 2023-01-31 at 2 15 40 PM

**Workaround:**

We manually added `class: "BtnGroup-item" and it worked, but I think it would be more expected to be able to use the `group_item` option as before. What do you think?

### Integration

No. This adds/fixes (depending on how you look at it) an option to `Beta::Button`

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated previews


camertron (commented):
Hey @paulcsmith, thanks for the PR! I might suggest taking a look at the [`ButtonGroup`](https://primer.style/view-components/components/beta/buttongroup) component? Sounds like it's exactly what you need!

jonrohan:
Yeah, we left out that feature on the new buttons as it's a design we'd like to move away from. Ideally in this case your buttons would be 2 separate buttons with a gap between them.

We are working on a [SplitButton](https://github.com/primer/view_components/pull/1436) for cases like the merge box/dropdown where 2 buttons are connected but need separate actions.

paulcsmith:
@jonrohan Ah that makes sense! Thanks for the explanation!

@camertron Thanks for the heads up! I did look at that, but due to how we're using this button I couldn't get `ButtonGroup` to work. We are wrapping the button in a details/summary tag and I couldn't figure out a way to get that tag to render with the button inside of it. It seems to only render the slotted component and nothing else

```
<%= render Primer::ButtonGroup.new do |component| %>
  <details><!-- the detail tag does not render -->
    <%= component.with_button(tag: :summary) %>
  </details>
<% end %>
```

I'm open to giving that a try if you know of a way to get the details tag to render?

camertron:
@paulcsmith unfortunately there's no way to get that `<details>` tag to render 😢 We'd have to add explicit support for it, although it sounds like the right way to go here per @jonrohan's comment is to use two separate buttons as you're already doing.

paulcsmith:
Thanks @camertron! That makes sense to me 👍

PR #1812, opened 2023-02-03
Adding label styling to the `ActionList::Item` component

### LINKED ISSUE #1811: There should be a way to inject style directly to `ActionList::Item` labels
Hello 👋🏼

The implementation of the `ActionList::Item` has the ability for us to provide the label class which is great but if there are conflict between different styling classes (for example `font-size`) after the classes get injected, it causes the styling to not work as expected and a direct override of those styles might be desirable in certain scenarios.

For example, when I use this like so

```erb
<%= render(
  Primer::Alpha::ActionList.new(
    role: "list",
    show_dividers: false,
    aria: {
      label: "My Action List"
    })
  ) do |component| %>
  <% list_items.each_with_index do |item, i| %>
    <% component.with_item(
      label: branch[:name],
      label_classes: "branch-name", # this is a class
      content_arguments: {
        style: "justify-items:start",
      },
      ) do |item| %>
      <% item.with_leading_visual_icon(
        icon: my_icon,
        color: my_color
      ) %>
      <% item.with_trailing_visual_counter(
        count: 1
      ) %>
    <% end %>
  <% end %>
<% end  %>

```

You're unable to resize the label at all. The size of the label is quite an important thing and the only available sizes are `:medium`, `:large` and `:xlarge`. I guess another question arises as to why the `small` size is missing? Anyways, if we could add styling for ourselves, it would make this very easy to work with.

tjgurwara99:
### Description

This is a relatively small change to the API, which simply adds the styling change for the `ActionList::Item` label. This is sometimes needed when we needed when we need to override some styling on certain elements.

Closes #1811 (if the issue is accepted)

### Integration

No.

### Merge checklist

- [ ] Added/updated tests - I would require some help for this actually
- [x] Added/updated previews

### DIFF
--- a/app/components/primer/alpha/action_list/item.rb
+++ b/app/components/primer/alpha/action_list/item.rb
@@ -156,6 +156,7 @@ module Primer
           active: false,
           on_click: nil,
           id: self.class.generate_id,
+          label_style: nil,
           **system_arguments
         )
           @list = list
@@ -192,7 +193,8 @@ module Primer
               label_classes,
               "ActionListItem-label",
               "ActionListItem-label--truncate" => @truncate_label
-            )
+            ),
+            style: label_style
           }
 
           @content_arguments[:id] = @id
--- a/previews/primer/alpha/action_list_preview.rb
+++ b/previews/primer/alpha/action_list_preview.rb
@@ -362,6 +362,18 @@ module Primer
           component.with_item(label: "Active item", href: "/", active: true)
         end
       end
+
+      # @label Item [label styled]
+      def item_label_styled
+        render(Primer::Alpha::ActionList.new(
+                 aria: { label: "List heading" }
+               )) do |component|
+          component.with_heading(title: "Action List")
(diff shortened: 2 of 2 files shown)


langermank:
Hey @tjgurwara99 thanks for the PR! Can you point me to a design that requires the ability to restyle the label? We have a specific style in place which aligns with our typography scale and also the layout of action list items (alignment with icons, description etc.) and ideally designs using ActionList follow those guidelines. If for some reason we need to allow custom styling of the label, we would prefer to use system arguments to perform those overrides.

tjgurwara99:
Hey @langermank 👋🏼 The design that I'm currently working on is not publicly available for now. I can send it to you over if you'd like. After discussing it over at the [issue page](https://github.com/primer/view_components/issues/1811) linked above the, it seems the issue I was having was solved by adding another label class `f6` so I can close this PR now. I still think it might be a good idea to have custom styling options but I appreciate that we want to limit certain things by design. Anyways, I'm closing this now since my issue has been resolved but I'm happy to reopen it if we think that it could benefit the community having these separate style arguments.

PR #1813, opened 2023-02-06
Rewrite selector test

camertron:
### Description

PVC contains this really cool test that renders all our previews and makes sure at least one of our CSS rules matches each DOM node. Last week I added a couple of previews that caused the test to start failing. I didn't understand why, so I dug into the implementation.

#### Background

Although the test code uses the word "selector," it's actually looking at CSS class names. It extracts class names from DOM nodes and tries to look them up in a big list of all the CSS selectors across all our .pcss files. For simple cases such an approach works well; however for more complicated scenarios where multiple classes are involved, etc, it has a hard time.

For example, the CSS for our form inputs contains the selector `.FormControl-radio-group-wrap fieldset`. The previews that render the `RadioButtonGroup` input generate a `<div class="FormControl-radio-group-wrap">` around a `<fieldset>` as expected by the selector. However the component selector test reports the `.FormControl-radio-group-wrap` as unused because it doesn't exist by itself in the list of selectors.

#### The Problem

I believe the fundamental issue with the component selector test is that it compares CSS classes to selectors when they are really different things. In my opinion the test should instead try to match the selectors to DOM nodes, which contain surrounding markup that may be important to the selector.

#### The Solution

I found it really difficult to use our existing integration test setup for this, so I rewrote it as a system test that renders previews in a browser. It walks the DOM and uses JavaScript's [`Element.matches`](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches) method to make sure that at least one of our CSS selectors matches it.

### Integration

No

### Merge checklist

- [x] Added/updated tests


test/css/component_selector_use_test.rb
```diff
@@ -49,60 +27,108 @@
 #
 # ensure all of the classes used by components are valid, checking against the
 # available selectors in component-specific CSS
-class ComponentSelectorUseTest < Minitest::Test
-  include Primer::ComponentTestHelpers
+class ComponentSelectorUseTest < System::TestCase
+  parallelize workers: 4
```
jonrohan, line 31: nice ⚡

jonrohan (approved):
Excellent improvements 🚀

PR #1814, opened 2023-02-07
Form components

camertron:
### Description

This PR exposes the following form inputs as Primer components:

* Check box
* Check box group
* Form button
* Multi input
* Radio button
* Radio button group
* Select list
* Submit button
* Text area

While these components are now available in PVC (complete with previews etc) they will remain unlisted in the side nav, as consumers of PVC should use the forms framework in most cases.

This PR also contains a few small fixes to form inputs discovered during the preview writing process.

### Integration

No.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


previews/primer/alpha/check_box_preview.rb
```diff
@@ -0,0 +1,70 @@
...
+      # @param full_width toggle
+      # @param disabled toggle
+      def playground(
+        name: "my-check-box",
+        id: nil,
+        value: "picard",
+        label: "Jean-Luc Picard",
+        caption: "Make it so",
+        visually_hide_label: false,
+        full_width: false,
```
langermank, line 24:
```suggestion

```

previews/primer/alpha/check_box_preview.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      # @label Visually hidden label
+      def visually_hide_label
+        render(Primer::Alpha::CheckBox.new(visually_hide_label: true, name: "my-check-box2", label: "Jean-Luc Picard"))
+      end
+
+      # @label Full width
+      def full_width
+        render(Primer::Alpha::CheckBox.new(full_width: true, name: "my-check-box3", label: "Jean-Luc Picard"))
+      end
```
langermank, line 60:
```suggestion

```

Checkbox shouldn't have a full width param

previews/primer/alpha/radio_button_preview.rb
```diff
@@ -0,0 +1,70 @@
...
+      # @param full_width toggle
+      # @param disabled toggle
+      def playground(
+        name: "my-radio-button",
+        id: nil,
+        value: "bsg",
+        label: "Battlestar Galactica",
+        caption: "A pretty good show",
+        visually_hide_label: false,
+        full_width: false,
```
langermank, line 24: Also remove full width from radio (both checkbox and radio dont need this param)

app/components/primer/alpha/select_list.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Primer
```
langermank, line 3: Can this just be called "Select" like PRC?

previews/primer/alpha/text_area_preview.rb
```diff
@@ -0,0 +1,90 @@
...
+      # @param invalid toggle
+      # @param validation_message text
+      def playground(
+        name: "my-text-area",
+        id: "my-text-area",
+        label: "Tell me about yourself",
+        caption: "You can trust me, I'm a website",
+        required: false,
+        visually_hide_label: false,
+        size: Primer::Forms::Dsl::Input::DEFAULT_SIZE.to_s,
```
langermank, line 27: TextArea shouldn't have a size param, but it can have full width

app/components/primer/alpha/select_list.rb
langermank: If it needs to be two words, maybe `SelectField` to match `TextField`

langermank (commented):
Just a few things from my first pass!

app/components/primer/alpha/select_list.rb
camertron: It doesn't need to be two words, I can change it.

jonrohan (approved):
Not a blocker, but one thing I've been wanting is React has a [`FormControl` component](https://primer.style/react/FormControl) which is used a lot in their examples of other components. Not having it in PVC has made it more difficult to create the same a11y docs for newer components in PVC. Take that as a while you're here comment rather than a requirement for merging. I'm fine if you want to follow up with that, or not follow up at all.

Still this is great. 🚀

camertron:
@jonrohan oh interesting! The forms framework actually does have a `FormControl` component we could expose. Let me see how difficult that would be to accomplish.

camertron:
Hmm ok unfortunately `FormControl` is pretty specific to the forms framework. We would have to create our own PVC version of it. I can look into that. What functionality do you need specifically? Label, caption, and validation message?

jonrohan:
> What functionality do you need specifically? Label, caption, and validation message?

Yeah I think it takes care of the label + element + caption id matching and a11y. It's possible that it does some basic spacing and positioning.

PR #1817, opened 2023-02-07
Forms framework docs

camertron:
### Description

Adds documentation in the form of Lookbook pages for the Primer forms framework 🎉 

### Integration

No

### Merge checklist

~- [ ] Added/updated tests~


lib/primer/yard/lookbook_pages_backend.rb
```diff
@@ -0,0 +1,194 @@
...
+      def page_id
+        @page_id ||= docs.short_name.dasherize.underscore.tap do |page_id|
+          page_id << "_input" unless page_id.end_with?("_input")
+        end
+      end
+
+      def generate
+        path = File.expand_path(
+          File.join(
+            *%w[.. .. .. demo test components docs forms inputs],
```
jonrohan, line 40: Should we unify these with the `previews/` folder? Just thinking about dotcom's lookbook grabbing that folder for displaying previews. Combining might allow us to serve these docs in dotcom as well

camertron: Oh yeah that's a great idea! When you say "unify," do you mean writing them to another directory at the same level as previews/ or _inside_ previews/?

jonrohan: Inside I think but separate from the actual previews. Like previews/pages/ ?

PR #1818, opened 2023-02-08
Fix issue in TabPanels causing extra content to render twice

### LINKED ISSUE #1790: Duplicated "extra" (renders_one)  when alignment is on the right side.
Hi,

I'm seeing a very weird behavior with the usage of `Primer::Alpha::TabPanels`.

The `extra`  (renders_one) is being displayed both on the left and right side if configured to be on the right side.

This is happening on production environment only, which is even more weird to me. On Staging environment it works as expected. 
I was able to reproduce it locally after changing env vars to "production", but unfortunately not to the point I can create a PR to share.
 
Some env settings:
```
Ruby 3.1.2p20
rails 7.0.4.2
gem "primer_view_components", "~> 0.0.119"
gem "view_component", "~> 2.82.0"
gem "view_component-contrib", "~> 0.1.2"
```

Teh code have a custom component that is using `Primer::Alpha::TabPanels`. This custom component has a `renders_one :extra_content`.
This is the component code

```ruby
class Shared::TabPanel::Component < ApplicationViewComponent
  option :label
  renders_many :items # (code simplified)

  renders_one :extra_content
end

<%= render(Primer::Alpha::TabPanels.new(label: label, display: :flex, body_arguments: { flex: 1 })) do |component| %>
  <% items.each do |item| %>
    <% component.with_tab(id: item.id, selected: item.selected) do |t| %>
      <% t.with_text { item.title } %>
      <% t.with_panel(classes: class_for(:panel)) do %>
        <%= item %>
      <% end %>
    <% end %>
  <% end %>

  <% component.with_extra(align: :right) do %>
    <%= extra_content %>
  <% end %>
<% end %>
```

What I noticed while debugging (with simple `puts` and changing the gem code locally) is that `@align` always starts with `:left` 
and then is changed to `:right`
on the code from 
https://github.com/primer/view_components/blob/972dd7bd92e79b2e8b54555f91770c4c1b280709/app/components/primer/alpha/tab_panels.html.erb#L3-L9
so the line with `<%= extra if @align == :left %>` always evaluates to true and renders the `extra`. 

Any help is appreciated.
Thank you for the great work on primer.
Best,

camertron:
### Description

The `TabPanels` component appears to double-render the `extra` slot in certain scenarios. It happens only when:

1. The `extra` slot setter is called with `align: :right`, eg:

    ```ruby
    render(Primer::Alpha::TabPanels.new) do |component|
      component.with_extra(align: :right) { "Foo" }
    end
    ```
1. In production environments.

In development, the `validate_single_selected_tab` method  calls the `tabs` slot getter, which causes view_component to eagerly evaluate `TabPanel`'s content. Doing so calls the `extra` slot setter _before_ the component is rendered, which correctly sets the `@align` instance variable to `:right`.

In production, `TabPanel`'s content is not eagerly evaluated, and the `extra` slot setter is called _during_ render instead. Since `@align` is set to `:left` in the constructor, `<%= extra if @align == :left %>` emits the `extra` content. Eventually the `extra` slot setter is called, which sets `@align` to `:right` and causes `<%= extra if @align == :right %>` to emit the `extra` content a second time.

This PR eagerly evaluates slots by calling `content` in `before_render`.

Fixes #1790 

### Integration

### Merge checklist

- [x] Added/updated tests

### DIFF
--- a/.changeset/flat-cups-press.md
+++ b/.changeset/flat-cups-press.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Fix issue in TabPanels causing extra content to render twice
--- a/app/components/primer/alpha/tab_panels.rb
+++ b/app/components/primer/alpha/tab_panels.rb
@@ -77,6 +77,13 @@ module Primer
         @body_arguments[:role] = :tablist
         @body_arguments[:"aria-label"] = label
       end
+
+      def before_render
+        # Eagerly evaluate content to avoid https://github.com/primer/view_components/issues/1790
+        content
+
+        super
+      end
     end
   end
 end
--- a/test/components/alpha/tab_panels_test.rb
+++ b/test/components/alpha/tab_panels_test.rb
@@ -87,6 +87,7 @@ class PrimerAlphaTabPanelsTest < Minitest::Test
   end
 
   def test_does_not_double_render_extra_content_in_production
+    # rubocop:disable Rails/Inquiry
     Rails.stub(:env, "production".inquiry) do
       # Since we've forced ourselves into the prod environment and there's no secret key base
       # configured for prod, we have to fake it by setting the appropriate environment variable.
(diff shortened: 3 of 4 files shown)


jonrohan (approved):
Great catch

khiga8 (approved):
amazing find. thank you!

PR #1823, opened 2023-02-10
Lint for reserved HTML classes

neall:
### Description

This linter checks for HTML classes that we consider "internal" to Primer ViewComponents.

### Integration

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


lib/primer/view_components/linters/view_component_html_classes_counter.rb
```diff
@@ -0,0 +1,30 @@
...
+# CLASSES—they will be responsible for complaining about the use of
+# those HTML classes.
+Dir[File.join(__dir__, "*.rb")].sort.each do |file|
+  require file unless file == __FILE__
+end
+
+module ERBLint
+  module Linters
+    # Counts the number of times a class reserved for ViewComponents is used
+    class ViewComponentHtmlClassesCounter < BaseLinter
```
jonrohan, line 15: One suggestion here, I think we could have a better name like `DisallowComponentCssCounter`

camertron (approved):
🎉

lesliecdubs:
@neall is this one ready to go out? 🚀

lib/primer/view_components/linters/disallow_component_css_counter.rb
```diff
@@ -0,0 +1,30 @@
...
+          File.read(
+            File.join(__dir__, "..", "..", "..", "..", "static", "classes.json")
+          )
+        ) - BaseLinter.subclasses.reduce([]) do |html_classes, klass|
+          html_classes.concat(klass.const_get(:CLASSES))
+        end
+      ).freeze
+
+      TAGS = nil
+      MESSAGE = "Primer ViewComponents defines some HTML classes with associated styles that should not be used outside those components. (These classes might have their styles changed or even disappear in the future.) Instead of using this class directly, please use its component if appropriate or define the styles you need some other way."
```
jonrohan, line 27: Should we include what component we think they should be using?

neall: Maybe.

We can definitely split up classes based on which components' stylesheets they come from. However, some of our classes are referenced by multiple components. For instance, the `Button--Small` class is referenced by `Primer::Beta::Button`, but it's also used in a selector for `Primer::Alpha::SegmentedControl`. I don't know a good way to determine which component we should suggest. We could suggest all the components that reference a class if you think that would be helpful.

jonrohan: hmm, yeah I see the problem. I think this is good enough for now then. 👍🏻

jonrohan (approved):

PR #1826, opened 2023-02-13
Fix production Docker builds

camertron:
### Description

Lookbook has failed to deploy for our last few PVC releases because of an issue with Docker builds.

We run a purge job before every deploy that is meant to clean up old Docker tags. We then pull the `:latest` app and asset images so the build step can use them in the `--cache-from` Docker build argument. Unfortunately it appears our purge job is a bit overzealous and recently deleted the `:latest` tag for the app image. This means the pull fails and kills the whole deploy job.

This PR does the following:

1. Ensures we never delete the `:latest` or `:latest-assets` tags.
2. Ensures pulling a non-existent tag doesn't fail the entire deploy.
3. Keeps the latest 20 images instead of the latest 10.

Note that the problem only crops up in production because preview deploys each use their own repository (i.e. URL) in our Azure Docker registry.

### Integration

No.

### Merge checklist

~- [ ] Added/updated tests~


jonrohan (approved):
🚀

PR #1828, opened 2023-02-15
Action list a11y

camertron:
### Description

This PR addresses accessibility feedback for the `NavList` and `ActionList` components as identified in https://github.com/github/primer/issues/1120. Namely it:

1. renders much clearer markup. Several unnecessary `<ul>`s have been removed and separators are now `<div>`s instead of `<li>`s
1. adjusts a number of roles
1. allows customizing the heading level
1. adds `aria-labelledby` to sub-list `<ul>`s. If a heading _and_ an `aria-label` are both provided, an error is now raised.
1. renames Section -> Group to align better with the React implementation
1. ~removes the `show_on_hover` attribute from `ActionList` and `NavList` items, as only showing buttons on hover is not an accessible pattern~ The `show_on_hover` option has been left in. We will address it (if necessary) in a future PR.
1. ensures sub-lists collapse when the user presses escape
1. updates the `NavList`'s trailing action preview so that the buttons actually do something
1. ensures `NavList`s are wrapped with the `<nav>` tag

### Integration

Yes. Since the `sections` slot has been renamed to `groups`, any place in dotcom where we're calling `with_section` will need to change to `with_group`. In addition, we will need to remove `aria-label`s from any section/group that also contains a header.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews


app/components/primer/alpha/action_list/heading.rb
```diff
@@ -11,21 +11,27 @@ class Heading < Primer::Component
...
         # @param title [String] Sub list title.
+        # @param heading_level [Integer] Heading level. Level 2 results in an `<h2>` tag, level 3 an `<h3>` tag, etc.
         # @param subtitle [String] Optional sub list description.
         # @param scheme [Symbol] Display a background color if scheme is `filled`.
-        # @param tag [Symbol] Semantic tag for the heading.
+        # @param tag [Integer] Semantic tag for the heading.
         # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
-        def initialize(list_id:, title:, tag: :h3, scheme: DEFAULT_SCHEME, subtitle: nil, **system_arguments)
-          @tag = tag
+        def initialize(list_id:, title:, heading_level: 2, scheme: DEFAULT_SCHEME, subtitle: nil, **system_arguments)
```
langermank, line 25: There was conversation about h2 vs h3 _somewhere_ in dotcom that I would need to go digging for, but I think we landed on h3 being the default. I believe @kendallgassner was working on this..

kendallgassner: oof I don't remember if I was talking about a specific feature case 🤔 .  If I am looking at the [searchResult](https://github.com/search?q=search&type=repositories) navigation menu for example it makes sense for `filterBy` to be a `h2` and languages to be an `h3`. I imagine most cases would be formatted in similarly so defaulting to `h3` would make most sense to me but I do want to ensure we have the option to change the level if needed (i.e. `as='h2`)?

app/components/primer/alpha/action_list/heading.html.erb
```diff
@@ -1,5 +1,5 @@
-<%= render(Primer::BaseComponent.new(**@system_arguments)) do %>
-  <%= render(Primer::BaseComponent.new(tag: @tag, classes: "ActionList-sectionDivider-title", id: @list_id)) do %>
+<%= render(Primer::BaseComponent.new(tag: :div, **@system_arguments)) do %>
+  <%= render(Primer::BaseComponent.new(tag: @tag, classes: "ActionList-sectionDivider-title", id: @list_id, role: :heading, aria: { level: @heading_level })) do %>
```
langermank, line 2: Does this really need a `role="heading"` if its a heading tag?

app/components/primer/alpha/action_list.rb
```diff
@@ -73,6 +73,7 @@ def initialize(
           system_arguments[:classes],
           "ActionListWrap",
           "ActionListWrap--subGroup",
+          "ActionListItem--hasSubItem",
```
langermank, line 76: I'm confused about the addition of classes here. This case should only have `ActionListWrap` for example 
image

`subGroup` and `hasSubItem` are all used for collapsible sections

app/components/primer/alpha/action_list.rb
```diff
@@ -81,18 +82,14 @@ def initialize(
 
       # @private
       def before_render
+        aria_label = aria(:label, @system_arguments)
+
         if heading.present?
           @system_arguments[:"aria-labelledby"] = @id
-        elsif aria(:label, @system_arguments).blank?
+          raise ArgumentError, "An aria-label should not be provided if a heading is present" if aria_label.present?
```
kendallgassner, line 89: 💙

app/components/primer/alpha/action_list/heading.rb
```diff
@@ -11,21 +11,27 @@ class Heading < Primer::Component
...
+        # @param heading_level [Integer] Heading level. Level 2 results in an `<h2>` tag, level 3 an `<h3>` tag, etc.
         # @param subtitle [String] Optional sub list description.
         # @param scheme [Symbol] Display a background color if scheme is `filled`.
-        # @param tag [Symbol] Semantic tag for the heading.
+        # @param tag [Integer] Semantic tag for the heading.
         # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
-        def initialize(list_id:, title:, tag: :h3, scheme: DEFAULT_SCHEME, subtitle: nil, **system_arguments)
-          @tag = tag
+        def initialize(list_id:, title:, heading_level: 3, scheme: DEFAULT_SCHEME, subtitle: nil, **system_arguments)
+          raise "Heading level must be between #{HEADING_MIN} and #{HEADING_MAX}" unless HEADING_LEVELS.include?(heading_level)
```
kendallgassner, line 26: 💙

app/components/primer/alpha/action_list/heading.html.erb
kendallgassner: Agreed if its `h1-h6` we shouldn't need the role or aria-level

static/constants.json
```diff
@@ -27,6 +27,16 @@
   },
   "Primer::Alpha::ActionList::Heading": {
     "DEFAULT_SCHEME": "subtle",
+    "HEADING_LEVELS": [
```
kendallgassner, line 30:
```suggestion
    "HEADING_LEVELS": Array.from({length: HEADING_MAX}, (_, i) => i + 1)
```

no big deal just a thought

app/components/primer/alpha/action_list/heading.rb
```diff
@@ -11,21 +11,27 @@ class Heading < Primer::Component
           :filled => "ActionList-sectionDivider--filled"
         }.freeze
         SCHEME_OPTIONS = SCHEME_MAPPINGS.keys.freeze
+        HEADING_MIN = 1
+        HEADING_MAX = 6
```
kendallgassner, line 15: Can you use the defined constants in static/constants.json?

app/components/primer/alpha/nav_list/item.html.erb
```diff
@@ -1,7 +1,7 @@
 <% with_private_content do %>
   <% unless items.empty? %>
     <% capture do %>
-      <%= render(Primer::BaseComponent.new(tag: :ul, **@sub_list_arguments)) do %>
+      <%= render(Primer::BaseComponent.new(tag: :ul, role: :list, **@sub_list_arguments)) do %>
```
lindseywild, line 4: Curious why we added the `role="list"` since the tag is a `<ul>`?

previews/primer/alpha/nav_list_preview/trailing_action.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= render(Primer::Alpha::NavList.new) do |list| %>
+  <% list.with_group do |group| %>
+    <%= group.with_heading(title: "Shopping list") %>
+    <% group.with_item(label: "Bread", href: "/list/1") do |item| %>
+      <%= item.with_trailing_action(show_on_hover: true, icon: :plus, aria: { label: "Click for alert" }, name: "bread_button") %>
```
lindseywild, line 5: This is super nitpicky but can we change the phrasing of the `aria-label`? Just because not everyone will be clicking on the item, they may be pressing the enter key! I know this is just for an example but it might prevent someone from trying to do this in the Real World™️ 
```suggestion
      <%= item.with_trailing_action(show_on_hover: true, icon: :plus, aria: { label: "Activate alert" }, name: "bread_button") %>
```

previews/primer/alpha/nav_list_preview/trailing_action.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <% end %>
+    <% group.with_item(label: "Cheese", href: "/list/2") do |item| %>
+      <%= item.with_trailing_action(icon: :plus, aria: { label: "Click for alert" }, name: "cheese_button") %>
```
lindseywild, line 8:
```suggestion
      <%= item.with_trailing_action(icon: :plus, aria: { label: "Activate alert" }, name: "cheese_button") %>
```

previews/primer/alpha/nav_list_preview/trailing_action.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <% end %>
+  <% end %>
+<% end %>
+
+<script type="text/javascript" data-eval="true">
+  const breadButton = document.querySelector("[name=bread_button]")
+  breadButton.addEventListener("click", () => alert("You clicked on bread."))
```
lindseywild, line 15:
```suggestion
  breadButton.addEventListener("click", () => alert("You selected bread."))
```

previews/primer/alpha/nav_list_preview/trailing_action.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+  const cheeseButton = document.querySelector("[name=cheese_button]")
+  cheeseButton.addEventListener("click", () => alert("You clicked on cheese."))
```
lindseywild, line 18:
```suggestion
  cheeseButton.addEventListener("click", () => alert("You selected cheese."))
```

static/constants.json
lindseywild: Sorry to disagree but I feel like this makes it harder to read 😅

lindseywild (commented):
This is awesome 🤩  Just a few suggestions!

app/components/primer/alpha/action_list/heading.html.erb
camertron: I think this must have just been us misinterpreting the a11y feedback 🥴

app/components/primer/alpha/action_list.rb
camertron: I think we added these here because they didn't seem to adversely affect the rendered result and because they were difficult to add elsewhere. Let me take another stab at this.

static/constants.json
camertron: Hmm, so this constants.json file is auto-generated from constants found in the component source code. We'd have to make changes there in order for them to be reflected here.

However, I don't understand why we'd want to increment all the heading levels by 1. Wouldn't that give us `h2` - `h7`?

app/components/primer/alpha/action_list/heading.rb
camertron: Ah, constants.json is generated from constants in the source code, so this is the right place for them to be.

static/constants.json
lindseywild: `<h7>` is not a thing! 😅

app/components/primer/alpha/nav_list/item.html.erb
camertron: Good question! I also wondered about this. Here's the relevant part of @jscholes' [comment](https://github.com/github/primer/issues/1120#issuecomment-1279387613) explaining:

> The `<ul>` wrapping the items in the "Moderation options" submenu does not have an accessible name. It also has no role of `"list"`, meaning that WebKit-based browsers will strip the list semantics due to the lack of bullets.
>
>Recommendations: add `role="list"` to the `<ul>`, as well as an `aria-labelledby` attribute referencing the ID of the "Moderation options" button.

previews/primer/alpha/nav_list_preview/trailing_action.html.erb
camertron: Love the ™ lol

static/constants.json
kendallgassner: That code would output 1-6 it (its using index so it starts at 0). That being said if its autogenerated then its not needed.

app/components/primer/alpha/nav_list/item.html.erb
lindseywild: Love it, thank you!

lindseywild (approved):

langermank (approved):
Looks great @camertron !

PR #1829, opened 2023-02-15
Add responsive values for `border` and `border-radius` system argument

josepmartins:
### Description

It's currently not possible to use responsive values for `border` even though the corresponding Primer CSS classes do exist: https://primer.style/design/foundations/css-utilities/borders#responsive-borders

Slack thread: https://github.slack.com/archives/CSGAVNZ19/p1676065040884579

primer/design adds responsive border-radius docs: https://github.com/primer/design/pull/397
primer/css adds responsive border-radius stories: https://github.com/primer/css/pull/2386

Screenshot 2023-02-15 at 11 31 01

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated previews


josepmartins:
> Hey @josepmartins, thanks for this! As it happens though, utilities.yml is generated by a rake task (`bundle exec rake utilities:build`). You'll need to modify it to include all the border stuff, otherwise your changes will get overwritten by the next run. For utilities that aren't well-suited to programmatic generation we use custom_utilities.yml, which is merged into the final result at the very end of `utilities:build`. It looks like custom_utilities.yml contains hard-coded entries for border classes, so we'll need to remove them and add the appropriate regex to `SUPPORTED_KEYS`.

@camertron thanks for reviewing it! ✨  I've tried adding the regex to the `SUPPORTED_KEYS` in the utilities, the problem is that generates a weird tree structure for borders. So the expected output for each border side [top, right, bottom, left] should be something like this:

```yml
:border_right:
  0:
     - border-right-0
     - border-sm-right-0
     - border-md-right-0
     - ... 
```


But instead, I'm getting this + the extra `:sm`, `:md`, `:lg` and `:xl`.

Screenshot 2023-02-16 at 10 57 53


I've ended up adding them to the `custom_utilities.yml` along with the missing responsive `border-radius` classes too, and updated the responsive classify tests to verify passing the arguments.

camertron:
@josepmartins ahh ok thanks for explaining!

lib/tasks/custom_utilities.yml
```diff
@@ -1,310 +1,370 @@
 :font_size:
-  "00":
```
camertron, line 2: What's up with all these indentation changes?

camertron (changes requested):

camertron:
Hey @josepmartins is this ready for another review?

camertron (approved):

PR #1832, opened 2023-02-17
support ViewComponent v3.0.0.rc2

joelhawksley:
This PR makes the only changes necessary to support ViewComponent v3.0.0.rc2. I intentionally tread carefully here so as to allow PVC to work with older versions of ViewComponent.


camertron (approved):
Nice, thanks!

PR #1845, opened 2023-02-27
Use system arguments instead of utility classes for NavList's "show more" item

### LINKED ISSUE #1843: NavList::Section `show_more_item` slot uses Primer CSS class
👋 Hello! I am unable to use the `show_more_item` slot of the `NavList::Section` component because it is using a [Primer CSS class under the hood](https://github.com/primer/view_components/blob/462bfb63e83d6a3535cd68c3c9b2176bb5e27717/app/components/primer/alpha/nav_list/section.rb#L31-L34), which causes the System Arguments exception to throw:

Screenshot 2023-02-22 at 5 57 00 PM

I can try to cut a PR to fix this tomorrow but wanted to flag this!

camertron:
### Description

`NavList` features a special "show more" item for lazy loading additional list items. Unfortunately it uses a Primer utility class instead of the corresponding system argument, and is causing an error in dotcom's Lookbook.

This would have been caught by our test suite except that we have `raise_on_invalid_options` set to `false` in the test environment 🤦 

Closes #1843.

### Integration

No. Should be backwards-compatible.

### Merge checklist

- [x] Added/updated tests

### DIFF
--- a/.changeset/tasty-rules-end.md
+++ b/.changeset/tasty-rules-end.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Use system arguments instead of utility classes for NavList's "show more" item
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,6 +32,9 @@ Naming/MethodParameterName:
 Rails/RefuteMethods:
   Enabled: false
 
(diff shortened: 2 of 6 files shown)


jonrohan (approved):
Thanks for the fix!

PR #1850, opened 2023-02-28
Fix beta button tooltip compatibility with display :block

mayamessinger:
### Description

Addresses a bug in which a beta Button with a tooltip does not fully respect the block parameter. This was because a tooltipped button is wrapped in a div that inherits its display width, the fix was to also add display: :block to that wrapper when appropriate.

### Integration

Yes

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/.changeset/red-crews-worry.md
+++ b/.changeset/red-crews-worry.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+Fix beta button tooltip compatibility with display :block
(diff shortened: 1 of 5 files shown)


mayamessinger:
Did not add any tests, since this is an edge case and tests seem to be per-parameter only, but happy to if wanted

jonrohan (approved):
Fantastic, thanks! ❤️

PR #1875, opened 2023-03-09
Allow ActionList dividers to be added manually

camertron:
### Description

Dividers in an `ActionList` can currently only be added by passing `show_dividers: true`, which will add a divider between each list item:

image

I'm working on integrating `ActionList` inside `ActionMenu` (currently an experimental component in dotcom) that needs to be able to insert dividers manually at user-defined positions within the list. This PR introduces the `#with_divider` method which offers such functionality. Here's an example:

image

### Why not use polymorphic slots?

This is a use-case tailor-made for polymorphic slots, but using them would necessitate a painful API change. Fortunately custom polymorphic slot names are [on the horizon](https://github.com/ViewComponent/view_component/pull/1652), but until then I've added a bit of a hacky workaround in the form of a `#with_divider` method that appends to the existing `items` slot using ViewComponent's private `#set_slot` method.

### Integration

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated previews

### DIFF
--- a/app/components/primer/alpha/action_list.rb
+++ b/app/components/primer/alpha/action_list.rb
@@ -39,28 +39,23 @@ module Primer
         Heading.new(list_id: @id, **system_arguments)
       }
 
-      # Adds an item to the list.
+      # Items.
       #
       # @param system_arguments [Hash] The arguments accepted by <%= link_to_component(Primer::Alpha::ActionList::Item) %>.
-      def with_item(**system_arguments)
-        @items << build_item(**system_arguments).tap do |item|
-          yield item if block_given?
-
+      renders_many :items, lambda { |**system_arguments|
+        build_item(**system_arguments).tap do |item|
           will_add_item(item)
         end
-      end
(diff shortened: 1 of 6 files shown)


app/components/primer/alpha/action_list.rb
```diff
@@ -48,6 +48,15 @@ def custom_element_name
         end
       }
 
+      # Adds a divider to the list of items.
+      #
+      # @param system_arguments [Hash] The arguments accepted by <%= link_to_component(Primer::Alpha::ActionList::Divider) %>.
+      def with_divider(**system_arguments, &block)
+        # This is a giant hack that should be removed when :items can be converted into a polymorphic slot.
+        # This feature needs to land in view_component first: https://github.com/ViewComponent/view_component/pull/1652
```
jonrohan, line 56: 👍🏻 A pretty wild hack

jonrohan (approved):

PR #1876, opened 2023-03-10
Bump `octicons` to v18

antn:
### Description

This updates `octicons` to use v18+, so we can use the icons added in https://github.com/primer/octicons/pull/928. This is part of https://github.com/github/discussions/issues/3079.

### Integration

The only breaking change in v18 is changing the following icon names:

- `issue-tracked-by` → `issue-tracks`
- `issue-tracked-in` → `issue-tracked-by`

I don't see those being used anywhere in Dotcom.

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,7 +4,7 @@ PATH
     primer_view_components (0.1.1)
       actionview (>= 5.0.0)
       activesupport (>= 5.0.0)
-      octicons (>= 17.0.0)
+      octicons (>= 18.0.0)
       view_component (> 2.0, < 4.0)
 
 GEM
--- a/primer_view_components.gemspec
+++ b/primer_view_components.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency     "actionview", ">= 5.0.0"
   spec.add_runtime_dependency     "activesupport", ">= 5.0.0"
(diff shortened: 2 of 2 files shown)


antn:
Not sure how we handle bumping versions of dependencies with breaking changes for this project, since there isn't a 1.0 yet. Should this just be `0.2.0`?

camertron:
Hey @antn, we've been bumping the patch version only for this project, regardless of breaking changes. Unfortunately I let a minor version bump through by accident which is how we ended up on v0.1.x 🤦 Let's keep bumping the minor version.

antn:
@camertron yeah, I went to make a changeset but I wasn't sure how we wanted to handle the versioning. Thanks for clarifying! Added in ec4f99c.

.changeset/witty-coats-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/view-components': minor
```
keithamus, line 2: This needs to be `patch`

```suggestion
'@primer/view-components': patch
```

keithamus:
> the minor version

Minor or Patch, just to be sure @camertron?

camertron:
Woops sorry @keithamus and @antn, I meant to say let's bump the _patch_ version 🤦 

No problem, I can adjust the changeset in master.

camertron:
Ah never mind, looks like @jonrohan took care of it before merge 🚀

PR #1879, opened 2023-03-10
Use ViewComponent::InlineTemplate

joelhawksley:
### Description

I've updated two components to use ViewComponent::InlineTemplate as an example. We'd need to require ViewComponent v3.0.0.rc4 or later if we land this change.


jonrohan:
Closing until 3.0 is more widely adopted

PR #1895, opened 2023-03-21
Modify merge_aria to combine plural attributes; introduce merge_data

camertron:
### Description

We've had a `#merge_aria` method for a little while now that handles merging aria attributes from system arguments. The problem is a bit tricky because aria attributes can be passed as a hash with an `aria:` key (eg. `aria: { foo: "bar" }`) or as individual key/value pairs (eg. `"aria-foo": "bar"`).

I had a need today to handle so-called "plural" aria attributes like aria-describedby and aria-labelledby, which support multiple space-delimited values, eg. `aria-labelledby="foo bar"`. I also wanted to handle `data` attributes.

I will submit another PR soon that actually uses these methods.

### Integration

No

### Merge checklist

- [x] Added/updated tests

### DIFF
--- a/.changeset/eight-rockets-brush.md
+++ b/.changeset/eight-rockets-brush.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Modify merge_aria to combine plural attributes; introduce merge_data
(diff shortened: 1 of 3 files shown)


jonrohan (approved):
Nice, I didn't realize labelledby could have multiple arguments.

PR #1897, opened 2023-03-23
Fix width of text fields and multi inputs

camertron:
### Description

We have a few forms in dotcom right now that look kinda funky:

Screenshot of the billing upgrade form. Some fields look fine but the city, zip, state, and country fields are all unequal in width, which is jarring to the eye

### Text inputs

A recent PR wrapped text inputs in the `<primer-text-field>` custom element, which has no styling. The framework expects all inputs to be wrapped with the `FormControl` class, which sets some necessary `flex-*` rules. This PR ensures this class is attached to `<primer-text-field>` and not one if its children.

### Multi inputs

Multi inputs suffer from the same problem, although the `<primer-multi-input>` custom element has been around for a bit longer.

### Integration

No

### Merge checklist

- [x] Added/updated tests

### DIFF
--- a/.changeset/tough-bats-refuse.md
+++ b/.changeset/tough-bats-refuse.md
@@ -2,4 +2,4 @@
 '@primer/view-components': patch
 ---
 
-Fix width of text fields
+Fix widths of text fields and multi inputs.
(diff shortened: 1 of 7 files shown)


jonrohan (approved):
Looks great 👍🏻

PR #1900, opened 2023-03-27
add icon button to dialog

### LINKED ISSUE #1384: Request: Allow `with_show_button` slot of `Dialog` to be rendered with an `IconButton`
Hello! Currently the `Dialog` has an optional `with_show_button` slot that can be rendered as `Button` to open the dialog.
We ran into a case where we wanted the slot to be rendered as an `IconButton`.
We were able to render this button standalone, but there weren't clear instructions around this.
Given `IconButton` is a PVC component, I think it would be acceptable to allow consumers to specify that the `with_show_button` should be rendered as an `IconButton`.

Request:
- Add clear instructions for how a button can be associated with the dialog. (via `data-show-dialog-id` attribute)
- Allow `with_show_button` to be rendered as an `IconButton`.

[Staff-only PR context](https://github.com/github/github/pull/237479#discussion_r973089992)

keithamus:
### Description

This adds the ability to pass an `icon` to `Dialog` `with_show_button`, which will turn it from a `Beta::Button` to a `Beta::IconButton`.

A similar technique is used in `Alpha::Overlay` & consequently the upcoming `Alpha::ActionMenu`.

Closes #1384 

### Integration

### Merge checklist

- [ ] Added/updated tests
- [x] Added/updated previews


jonrohan (approved):
Looks good, I think the only failure is that it needs a test for the new option

PR #1907, opened 2023-03-30
Suggest ViewComponents on reserved class use

neall:
### Description

When developers use HTML classes that we reserve for Primer ViewCompoents, we want to suggest relevant components they might want to use instead.

Closes [Primer#1901](https://github.com/github/primer/issues/1901)

### Implementation

This changes the generated `static/classes.json` file from a pure array of HTML classes to restrict into an object where the keys are those HTML classes and the values are arrays of the Ruby class (or classes) associated with that HTML class.

### Merge checklist

- [x] Added/updated tests
- ~[ ] Added/updated documentation~
- ~[ ] Added/updated previews~

### DIFF
--- a/.changeset/tall-phones-battle.md
+++ b/.changeset/tall-phones-battle.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Fix up reserved CSS class linter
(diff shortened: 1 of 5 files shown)


camertron (approved):
Boom!! Awesome! 🎉

PR #1933, opened 2023-04-05
Display Buttons inline

simurai:
### Description

This changes 

- the default `Button` from `flex` to `inline-flex`
- the `Button--iconOnly` from `grid` to `inline-grid`

This came up in https://github.com/github/github/pull/268064#pullrequestreview-1370516944. Most often the parent element is a flexbox, which takes over the children's display property. But it could cause regressions when migrating cases without a flexbox container or floats since the old `btn` uses `inline-block`.

Before | After
--- | ---
Screen Shot 2023-04-05 at 14 19 35 | Screen Shot 2023-04-05 at 14 19 04

### Integration

I don't think so. Use cases [that already changed to something like `inline_block`](https://github.com/github/github/pull/268064#pullrequestreview-1370516944.) should not be affected.

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


keithamus (approved):
LGTM

jonrohan (approved):

PR #1938, opened 2023-04-06
Fix and expose ActionMenu docs

camertron:
### Description

I totally forgot to generate the docs for the newly upstreamed `ActionMenu`. I had to bump view_component to v3 in order to bring in the new capture compatibility module, without which I observed some double-rendering in the `ActionMenu` examples. The problem does not appear to happen in Lookbook or in dotcom, so I believe it has something to do with the non-standard way we render examples for the legacy Gatsby site.

### Integration

No

### DIFF
--- a/demo/Gemfile
+++ b/demo/Gemfile
@@ -35,7 +35,7 @@ gem "puma", "~> 6.2.1"
 gem "bootsnap", ">= 1.4.2", require: false
 
 gem "primer_view_components", path: "../"
-gem "view_component", '>= 2.81.0'
+gem "view_component", '= 3.0.0.rc5'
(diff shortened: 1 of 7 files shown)


demo/Gemfile
```diff
@@ -35,7 +35,7 @@ gem "puma", "~> 6.2.1"
 gem "bootsnap", ">= 1.4.2", require: false
 
 gem "primer_view_components", path: "../"
-gem "view_component", '>= 2.81.0'
+gem "view_component", '= 3.0.0.rc5'
```
jonrohan, line 38: Should this be a `>=`?

camertron: Ah yeah good catch.

jonrohan (approved):

PR #1960, opened 2023-04-18
don't use to_sentence in linter

joelhawksley:
### Description

We run our linters outside of Rails in some cases, which means we can't use `to_sentence`.


jonrohan (approved):

camertron (approved):
Woops, this one was my fault 😬

Issue #1937, opened 2023-04-06
Missing border for `:neutral` scheme row in BorderBox component


gnfisher:
👋 We are using the Box component in discussions for our categories index view. We are adding a new feature, sections, and sections appear in a row with a `:neutral` rather than `:default` scheme. When one of these rows is the first row, there are some visual mishaps where the black bottom border of the header becomes invisible:

<img width="700" alt="Screenshot 2023-04-06 at 11 44 29 AM" src="https://user-images.githubusercontent.com/4742306/230433661-71ce6622-0f3d-43f9-a00f-493e74371285.png">

It looks to be from this css:

<img width="228" alt="image" src="https://user-images.githubusercontent.com/4742306/230433962-d3aa0bd3-8c1e-4a41-9b8f-09b3a39421b4.png">

There are curved radiuses that are visible when the background is dark. I don't believe the curved radius here is desired? Or not if there is a header?

<img width="238" alt="Screenshot 2023-04-06 at 11 50 50 AM" src="https://user-images.githubusercontent.com/4742306/230433704-d02b15c1-d441-4e56-8d3b-61b57114e027.png">

And this somehow inches the element a pixel higher, so the grey hides the black border. I don't think this is necessary, either? But I may be missing something.

lesliecdubs:
Thanks for reporting, this looks like a bug to us.

@simurai PVC maintainers said you've looked into similar bugs in the past; do you have any thoughts on a quick fix? If it's not something you have a good sense about how to fix within 15minutes or less, let us know and maybe Primer Design will want to look into prioritizing it. cc @tallys

simurai:
The [Box styles](https://primer.style/css/components/box) that the `BorderBox` component uses has [lots of rounded corner bugs](https://github.com/github/primer/issues/499) and is quite brittle/unreliable by using the `:first-of-type` selector. And there are like [454 files](https://github.com/search?q=repo%3Agithub%2Fgithub+Box-row&type=code) that use `Box-row` and that is just custom markup without using ViewComponents. So any change to the `Box` styles has a high risk of introducing other regressions. 😩 

That said, we also have never really tried to fix them, but would require lots of testing, just because of the high usage count and maybe would warrant a "Batch" effort?

> We are using the Box component in discussions for our categories index view.

Would it be possible to use a utility class like `rounded-0` to remove the rounded corners? And maybe also `mt-0` to add the border back?

gnfisher:
[...]

PR #1961, opened 2023-04-19
Only smooth radius if row is first/last and first/last child of Box

gnfisher:
### Description

This changes the selector to make sure the parent `ul` element is either the first child or last child of the parent `.Box` element as well as if the row is the first or last child in the list. Previously we were only doing the latter, which lead to the rounded corners showing up when a row was below a header element, for example.

My CSS skills aren't the sharpest. If there is a better way to write this rule, please suggest a change. 🙏 

I've also not updated the changeset. Let me know if you would like me to. I've made this PR editable by maintainers as well. 👍 

Closes https://github.com/github/discussions/issues/3192

### Integration

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

## Before
row1-before

## After
row1-after

### DIFF
--- a/app/components/primer/beta/border_box.pcss
+++ b/app/components/primer/beta/border_box.pcss
@@ -105,16 +105,6 @@
   border-top-style: solid;
   border-top-width: var(--primer-borderWidth-thin, 1px);
 
-  &:first-of-type {
-    border-top-left-radius: var(--primer-borderRadius-medium, 6px);
-    border-top-right-radius: var(--primer-borderRadius-medium, 6px);
-  }
-
-  &:last-of-type {
-    border-bottom-right-radius: var(--primer-borderRadius-medium, 6px);
-    border-bottom-left-radius: var(--primer-borderRadius-medium, 6px);
-  }
-
   /* Adds a blue vertical line to the left of the row
   ** For indicating a row item is unread */
   &.Box-row--unread,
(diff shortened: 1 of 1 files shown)


app/components/primer/beta/border_box.pcss
```diff
@@ -153,6 +143,18 @@
   }
 }
 
+/* Properly round corners if the first element of the Box is a row element */
+.Box > ul:first-child > li.Box-row:first-of-type {
```
simurai, line 147: Unfortunately these styles are not being used for only Primer ViewComponents, but also for lots of custom markup on dotcom that use the `Box` classes directly. And sometimes it's a `<div>` or other HTML elements. E.g. [this example](https://github.com/github/github/blob/c42ea27d018b69e8704fa5e624c1dcc9671efb60/app/views/orgs/people/_sidebar.html.erb#L7-L8) would probably need:

```scss
.Box > .Box-row:first-of-type {
```

So I'm pretty sure this change would cause regressions where some `Box` use cases would miss the rounded corners due to custom markup use. 🤷 

Also see [this other comment](https://github.com/primer/view_components/issues/1937#issuecomment-1516124055).

simurai: > We have implement a hack with hx to style it properly for our use case

Ok, great.

> or I can close this issue and leave it for your team to handle if thats more helpful at the moment.

Yeah, I think I'm gonna close this again until we can prioritize it. It just feels like it's something that needs extensive testing, just because there are hundreds of use cases. All with custom implementations from the time before we had ViewComponents and more control over the markup.

Thanks for the PR though.

PR #1987, opened 2023-05-03
Bug: Autocomplete results padding

langermank:
### Description

While updating Autocomplete to use the new ActionList CSS, the inset padding was left off (which was previously the default setting). This adds the proper class.

Closes https://github.com/github/issues-graph/issues/2003

### Integration

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/.changeset/strange-suns-study.md
+++ b/.changeset/strange-suns-study.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+add padding back to autocomplete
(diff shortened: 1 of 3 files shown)


langermank:
@jonrohan would it be possible to add snapshot tests of the Autocomplete component _open_?

jonrohan:
> would it be possible to add snapshot tests of the Autocomplete component _open_?

possible, we would need to write custom actions for that preview. Unfortunately don't have any precedence for that yet, so will be hardest initially getting it done.

jonrohan (approved):

PR #1998, opened 2023-05-10
use correct psuedo selector for popover

keithamus:
### Description

`popover` has changed to use `:popover-open`

### Integration

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews

### DIFF
--- a/app/components/primer/alpha/overlay.pcss
+++ b/app/components/primer/alpha/overlay.pcss
@@ -6,7 +6,7 @@
 }
 
 /* stylelint-disable-next-line selector-pseudo-class-no-unknown */
-.Overlay[popover]:not(:open) {
+.Overlay[popover]:not(:popover-open) {
   display: none;
 }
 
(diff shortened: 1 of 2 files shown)


langermank (approved):

app/components/primer/alpha/overlay.pcss
```diff
@@ -6,7 +6,7 @@
 }
 
 /* stylelint-disable-next-line selector-pseudo-class-no-unknown */
-.Overlay[popover]:not(:open) {
+.Overlay[popover]:not(:popover-open) {
```
jonrohan, line 9: Should we have both selectors for a bit? So we cover browsers that don't use the new selector yet
```suggestion
.Overlay[popover]:not(:open),
.Overlay[popover]:not(:popover-open) {
```

keithamus: We should avoid this, as `:open` is no longer a valid selector. Current browsers will throw away the whole declaration and future ones might have `:open` which will cause unintended consequences.

Issue #1986, opened 2023-05-02
[Bug] Can't set `aria-label` on `IconButton` if tooltips are disabled


strackoverflow:
👋 Hello! I noticed an issue with the `IconButton` component in that you can't set an `aria-label` attribute on the button if you also pass in `show_tooltip: false`. 

**Steps to reproduce**

In the following example, one would expect to be able to set an `aria-label` on the button without enabling tooltips:

```
<%= render(Primer::Beta::IconButton.new(icon: :search, aria: { label: "Search" }, id: "search-button", show_tooltip: false)) %>
```

However, the `aria-label` attribute is not present in the rendered markup.

**Workaround**

This issue can be worked around by providing `aria-description` in addition to the `aria-label`. The value in `aria-label` will be preserved and rendered into the button's markup, while `aria-description` will be used for tooltip content (even if tooltips have been disabled).

It appears the issue is in [this block of code](https://github.com/primer/view_components/blob/23b6b4d76a3980358d628f91f778c8d363ad328d/app/components/primer/icon_button.rb#L89-L99), which removes the `aria-label` from the system arguments hash so that it can be used in the tooltip, but it does this regardless of if tooltips are enabled or not.

lesliecdubs:
👋 @strackoverflow would you be down to pair with @joelhawksley to upstream a fix for this?

strackoverflow:
Sure thing! I can coordinate with Joel on this.

PR #2025, opened 2023-05-22
Preserve aria-label when icon button show_tooltip is false

strackoverflow:
Closes https://github.com/primer/view_components/issues/1986


camertron (approved):
Thanks! 🎉

PR #2034, opened 2023-05-31
Partially address a11y issues in Link component.

radglob:
### Description

This PR partially addresses some of the concerns in https://github.com/github/primer/issues/1880.

It removes the option to render Links as `<span>`s and removes the negative `outline-offset` causing focus outlines to overlap with the first and last letters in the link text.

It does not fix issues with the Tooltip component (which is somewhat out of scope of the Link component), address color contrast issues in the dark-dimmed theme (this is something we're just not addressing at this time), or force links to show an underline at all times if surrounded by other text.

### Integration

This will, as written, break usages of Link where the tag is set to `:span`. I've found three places in dotcom which use this argument.
* https://github.com/github/github/blob/1c1cfd98dfbf432ac0ce8255a2bb9f35cc6535d6/app/views/code_navigation/_timed_out.html.erb#L9
* https://github.com/github/github/blob/1c1cfd98dfbf432ac0ce8255a2bb9f35cc6535d6/app/views/code_navigation/_no_results.html.erb#L9
* https://github.com/github/github/blob/1c1cfd98dfbf432ac0ce8255a2bb9f35cc6535d6/app/views/code_navigation/_results_backend.html.erb#L5.

These will need to be changed, and perhaps styled to maintain their appearance. It seems like the most appropriate change would be for these Links to use the Button component with the `link` variant.

### Merge checklist

- [X] Added/updated tests
- [X] Added/updated documentation
- [X] Added/updated previews


camertron (commented):
Hey @radglob thanks for this 👍 Any idea how many places in dotcom are passing `tag: :span`?

radglob:
> Hey @radglob thanks for this 👍 Any idea how many places in dotcom are passing `tag: :span`?

@camertron Not sure how to check that. I've seen [this](https://musical-adventure-wlr3n3k.pages.github.io/?=s) for checking JS code. Is there something similar for checking Ruby stuff?

app/components/primer/beta/link.rb
```diff
@@ -55,30 +52,29 @@ class Link < Primer::Component
...
       # @param underline [Boolean] Whether or not to underline the link.
       # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
-      def initialize(href: nil, tag: DEFAULT_TAG, scheme: DEFAULT_SCHEME, muted: false, underline: true, **system_arguments)
+      def initialize(href: nil, scheme: DEFAULT_SCHEME, muted: false, underline: true, **system_arguments)
         @system_arguments = system_arguments
 
         @id = @system_arguments[:id]
 
-        @system_arguments[:tag] = fetch_or_fallback(TAG_OPTIONS, tag, DEFAULT_TAG)
+        @system_arguments[:tag] = :a
```
camertron, line 65: Can we use the [`deny_tag_argument`](https://github.com/primer/view_components/blob/9829b632e1f6c16262377fa2c7e5dcb7bc4d929b/app/components/primer/component.rb#L133) helper here to raise an error if a tag is passed? That should prevent us having to write a linter, which I think is a bit overkill.

camertron (changes requested):

radglob:
> @radglob I usually just use VSCode's search feature in a Codespace. I just spun one up because I was curious and found only 3 usages, which isn't that many. Here's what I found:
> 
> 1. https://github.com/github/github/blob/1c1cfd98dfbf432ac0ce8255a2bb9f35cc6535d6/app/components/pages/protected_domains/domain_component.html.erb#L4
> 2. https://github.com/github/github/blob/1c1cfd98dfbf432ac0ce8255a2bb9f35cc6535d6/app/views/orgs/invitations/teams/_list.html.erb#L7
> 3. https://github.com/github/github/blob/1c1cfd98dfbf432ac0ce8255a2bb9f35cc6535d6/app/views/verifiable_domains/index.html.erb#L32
> 
> It would be great to have a plan for migrating these that you could include in the PR description under the "Integration" heading. We would need to potentially reach out to the teams who own these files and understand why they chose to use a `<span>` instead of an `<a>`.

I found three places in dotcom (the examples you have are for the Label component, not Link): https://github.com/github/github/blob/1c1cfd98dfbf432ac0ce8255a2bb9f35cc6535d6/app/views/code_navigation/_timed_out.html.erb#L9, https://github.com/github/github/blob/1c1cfd98dfbf432ac0ce8255a2bb9f35cc6535d6/app/views/code_navigation/_no_results.html.erb#L9, https://github.com/github/github/blob/1c1cfd98dfbf432ac0ce8255a2bb9f35cc6535d6/app/views/code_navigation/_results_backend.html.erb#L5.

I don't see anything obvious about these examples that would make the change a slow process. I imagine it was probably for stylistic reasons or something.

camertron:
Ah woops, sorry about that 😅 Ok cool! It would be great to verify that tho, just in case the team chose `<span>` for some specific reason. We could always make the change and then ask them to raise any objections during code review.

jonrohan (approved):

camertron:
Hey @radglob apologies for the slow response. Were you able to track down if changing any of the `<a>`s to `<span>`s is going to be problematic?

PR #2038, opened 2023-05-31
replace `@changesets/cli` with `primer-changesets-cli`

gr2m:
### Description

This is part of an effort to provide a notification system to users about Primer releases.

### Screenshots

https://github.com/primer/react/assets/39992/bf795016-570d-4462-9cde-e8e423e3d964

### Integration

No

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated previews


camertron (commented):
Hey @gr2m this looks great! I'm curious where the list of components comes from though. For example, `BranchName` and `FilteredSearch` aren't available in PVC. I would like to make sure PVC components that aren't available in PRC are represented in the list as well.

gr2m:
@siddharthkp gave me these tips:

1. `@primer/react`: use the `generated/components.json` file
2. `@primer/view_components`: retrieve components from https://primer.github.io/view_components/components.json

When you run the CLI it will log out where it retrieves the components.

Ideally I'd like it to retrieve the componenents in both cases from local files.

camertron:
@gr2m interesting, thanks for that explanation. I looked through our codebase and can't figure out what generates components.json, do you know where that happens?

> When you run the CLI it will log out where it retrieves the components.

Ah ok I was a bit thrown off by the screen recording, but upon closer inspection it was recorded inside the primer/react repo as opposed to PVC.

> Ideally I'd like it to retrieve the components in both cases from local files.

There are a number of other files in the static/ directory that contain all the information you're looking for. Check out static/info_arch.json.

gr2m:
> I looked through our codebase and can't figure out what generates components.json, do you know where that happens?

No 🤣 I had the same question. It's easy to change the source of the components though

> There are a number of other files in the static/ directory that contain all the information you're looking for. Check out static/info_arch.json.

Okay I created an issue here: https://github.com/gr2m/primer-changesets-cli/issues/3

Looks like `static/info_arch.json` has 7 more components. Also it includes the capitalized module names, which I guess is preferable?

camertron:
> No 🤣 I had the same question. It's easy to change the source of the components though

Ok cool 😎 If we can figure out what generates components.json I'd be happy to update it to include the missing components and/or include the fully-qualified constant names, but barring that info_arch.json should work.

> Looks like `static/info_arch.json` has 7 more components. Also it includes the capitalized module names, which I guess is preferable?

Yes, definitely. Unlike React, our components go through a series of stages, i.e. experimental -> alpha -> beta -> stable, which is reflected in the component's namespace. In other words, `Primer::Alpha::Label` and `Primer::Beta::Label` are actually different components.

camertron:
Hey @gr2m this is looking good 👍 As soon as these merge conflicts are resolved we should be good to go.

gr2m:
I added a CI check that not only checks for the presence of `.changesets/*.md` files, but also validates the `<!-- Changed components: ... -->` footer.

I realized that `primer/react` use [skip changeset](https://github.com/primer/react/issues?q=is%3Apr+is%3Aopen+label%3A%22skip+changeset%22) as label, while `primer/view_components` is using [skip changelog](https://github.com/primer/view_components/labels/skip%20changelog). No problem, I'll make the CI work with both, but maybe it would make sense to normalize that across the repositories?

gr2m:
This should be ready to be merged

camertron:
Yeah, we should use `skip changeset` everywhere.

camertron:
See: https://github.com/primer/view_components/pull/2089

camertron (approved):

gr2m:
The failing CI is not required, but the error seems related to changes in this pull request:

> invalid reference: primer-changesets-cli

I'll have another look

gr2m:
> The failing CI is not required, but the error seems related to changes in this pull request:
> 
> > invalid reference: primer-changesets-cli
> 
> I'll have another look

Nevermind, I'm pretty sure the `primer-changesets-cli` is the branch name, not related to the new npm package.

gr2m:
@camertron I'm not authorized to merge, but this is good to go

PR #2070, opened 2023-06-09
Fix disabled button behavior

camertron:

This PR updates all of our buttons (via the `BaseButton` component) to support being disabled. This is done by forcing the tag to `<button>` if `disabled: true` is specified.

### Screenshots

Screenshots are not necessary in this case because all buttons still appear visually disabled with this change.

### Integration

I wrote up a [fake ERB linter and Ruby code scanner](https://gist.github.com/camertron/e2cfea771cc7479acbc150dfdc0c5665) to identify cases where the `disabled:` option is passed to `Beta::Button`, `Alpha::ButtonMarketing`, `Alpha::HellipButton`, `Beta::IconButton`, `Beta::CloseButton`, and `Beta::BaseButton` that also specify a tag other than `<button>`.

The scripts linked above only identified a single usage that fits the criteria:

1. [app/components/repos/advisories/report_vulnerability_button_component.html.erb:L1](https://github.com/github/github/blob/e45446afa961a170b27390185da25aecfc353e19/app/components/repos/advisories/report_vulnerability_button_component.html.erb#L1)

Strictly speaking, we don't actually have to fix the usage above - the code will continue to work as written. No doubt the author discovered what this PR is trying to fix - namely that `<a>` tags cannot be disabled. That said, we should take the opportunity to clean this up by setting `tag: :a`.

Fixes https://github.com/github/primer/issues/2222

#### Risk Assessment
  

Currently a number of our buttons that allow tag customization can't be disabled if a tag other than `<button>` is used. This is because the other two types of allowed tag, `<a>` and `<summary>`, cannot be disabled by setting the `disabled` HTML attribute as can be done for `<button>`s. In the `ActionMenu` component, we solved this by forcing the tag to `<button>` if `disabled: true` is specified. This PR does the same for all users of `BaseButton`.

This change could cause problems with code that assumes a particular tag will be used for any given button. I feel as though I did my due diligence tracking down usages though, so we should be good to go.

### Accessibility

This change doesn't fix an Axe violation (that I know of), but it does make our buttons more accessible overall.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/tall-ducks-brake.md
+++ b/.changeset/tall-ducks-brake.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': minor
+---
+
+Fix disabled button behavior
--- a/app/components/primer/alpha/button_marketing.rb
+++ b/app/components/primer/alpha/button_marketing.rb
@@ -42,12 +42,14 @@ module Primer
       # @param variant [Symbol] <%= one_of(Primer::Alpha::ButtonMarketing::VARIANT_OPTIONS) %>
       # @param tag [Symbol] <%= one_of(Primer::Alpha::ButtonMarketing::TAG_OPTIONS) %>
       # @param type [Symbol] <%= one_of(Primer::Alpha::ButtonMarketing::TYPE_OPTIONS) %>
+      # @param disabled [Boolean] Whether or not the button is disabled. If true, this option forces `tag:` to `:button`.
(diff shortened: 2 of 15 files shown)


jonrohan (approved):
🤘🏻

PR #2143, opened 2023-07-13
Update index.md documentation with correct install path

writercoder:
Use the full import path in documentation as recommended in the PR here https://github.com/primer/view_components/pull/1731#issuecomment-1367848262 to avoid `Uncaught DOMException` errors around double registration of web components.


jonrohan (approved):
Thanks for the contribution!

PR #2190, opened 2023-08-03
Button tooltips should not have the label type.

radglob:
_Authors: @radglob_

`Primer::Beta::Button.with_tooltip` should no longer use the `label` type. This causes an issue documented in https://github.com/github/primer/issues/2194 where the generated `aria-labelledby` attribute would override the inner text of a button to users using screen readers. Tooltips on buttons with visible text should only use the `description` type, which generates an `aria-describedby` attribute.

Updated the previews for the Primer::Alpha::Tooltip (and removed one) to only show examples with the `:label` type and added a deprecation warning for cases where the `:label` type is used.

### Integration
Cases where `Primer::Beta::Button.with_tooltip` is called with the `:label` type will continue to render successfully, defaulting to the `:description` type instead. They should probably be updated to avoid the deprecation warning.

Closes https://github.com/github/primer/issues/2194

#### Risk Assessment
  

- [X] **Low risk** the change is small, highly observable, and easily rolled back.

Considering that it seemed somewhat difficult to detect all cases where a button has visible text, imploring users to use the `IconButton` component in cases where they want to use the tooltip as a label makes more sense.

### Accessibility

### Merge checklist

- [ ] Added/updated tests
- [X] Added/updated previews (Lookbook)
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge

### DIFF
--- a/.changeset/tasty-waves-rhyme.md
+++ b/.changeset/tasty-waves-rhyme.md
@@ -1,0 +1,7 @@
+---
+'@primer/view-components': patch
+---
+
+Updates Primer::Beta::Button.with_tooltip to not accept `:label` type.
+
+<!-- Changed components: Primer::Beta::Button -->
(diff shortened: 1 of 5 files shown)


camertron:
Hey @radglob, thanks for this 😄 It looks like a few tests are failing because there are some template-based previews that still need to be updated.

camertron (approved):

PR #2182, opened 2023-08-01
add button migration linter

keithamus:
_Authors: Please fill out this form carefully and completely._

_Reviewers: By approving this Pull Request you are approving the code change, as well as its deployment and mitigation plans._
_Please read this description carefully. If you feel there is anything unclear or missing, please ask for updates._

#### List the issues that this change affects.

Closes # (type the GitHub issue number after #)

#### Risk Assessment
  

- [ ] **Low risk** the change is small, highly observable, and easily rolled back.
- [ ] **Medium risk** changes that are isolated, reduced in scope or could impact few users. The change will not impact library availability.
- [ ] **High risk** changes are those that could impact customers and SLOs, low or no test coverage, low observability, or slow to rollback.

### Accessibility

- **Fixes axe scan violation** - This change fixes an existing [axe scan](https://thehub.github.com/epd/engineering/dev-practicals/frontend/accessibility/readiness-routine/development/#axe-scans) violation.
- **No new axe scan violation** - This change does not introduce any new [axe scan](https://thehub.github.com/epd/engineering/dev-practicals/frontend/accessibility/readiness-routine/development/#axe-scans) violations.
- **New axe violation** - This change introduces a new [axe scan](https://thehub.github.com/epd/engineering/dev-practicals/frontend/accessibility/readiness-routine/development/#axe-scans) violation. Please describe why the violation cannot be resolved below.

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Lookbook)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2205, opened 2023-08-14
Add `IconButton` linter

TylerJDev:

Builds off the work done in https://github.com/primer/view_components/pull/2182.

Introduces a linter for IconButton, while also autocorrecting some instances.

### Integration

N/A

https://github.com/github/primer/issues/2342

#### Risk Assessment
  

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


test/lib/rubocop/iconbutton_component_test.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def test_warn_on_icon_button_simple_arg
+    investigate(cop, <<~RUBY)
+      Primer::IconButton.new(0)
+    RUBY
+
+    assert_equal 1, cop.offenses.count
+    assert cop.offenses.first.corrector.present?
+
+    assert_equal "Primer::Beta::IconButton.new(0)", cop.offenses.first.corrector.rewrite.strip
```
camertron, line 34: I believe `Primer::Beta::IconButton.new(0)` will raise an error when executed because the constructor does not accept any positional arguments. The name of this test suggests the cop should warn in this case, not autocorrect - is that right?

camertron (changes requested):

test/lib/rubocop/iconbutton_component_test.rb
keithamus: This invocation is more meant to just test that arguments are correctly transferred when correcting. The arguments are not altered, and this isn't a linter to check for correctness in the arguments, so I think this case is okay.

camertron: Ah ok I see. I think I would feel better if the example used kwargs since that's much closer to how these instances will look in dotcom.

test/lib/rubocop/iconbutton_component_test.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def test_warn_on_icon_button_no_args
+    investigate(cop, <<~RUBY)
+      Primer::IconButton.new()
+    RUBY
+
+    assert_equal 1, cop.offenses.count
+  end
+
+  def test_warn_on_icon_button_with_simple_arg
```
camertron, line 26:
```suggestion
  def test_warn_on_icon_button_with_args
```

camertron (approved):

test/lib/rubocop/iconbutton_component_test.rb
camertron: Yes, that looks good to me!

PR #2210, opened 2023-08-15
Fix accessibility bug of missing accessible name on Dialog

khiga8:

When I was doing unrelated work in dotcom, I noticed that `Primer::Alpha::Dialog` is not rendering with an accessible name resulting in an Axe failure of [aria-dialog-name](https://dequeuniversity.com/rules/axe/4.1/aria-dialog-name):

> Ensures every ARIA dialog and alertdialog node has an accessible name"

This is unexpected since having an accessible name is also [clearly documented in the docs](https://primer.style/design/components/dialog/rails/alpha#accessibility).

There's currently no corresponding behavioral test written for this component. It's likely we're not running an axe scan when the Dialog is an open state, causing us to miss this error.

This PR fixes the bug and adds a behavioral system test.

Closes https://github.com/github/primer/issues/2556

#### Risk Assessment
  

Fix bug and add missing behavioral test.

### Accessibility

### Merge checklist

- [x] Added/updated tests

### Questions to reviewers

📣 Are behavioral components (or components that render with JS) identified in any way? I couldn't find anything that would indicate a component has JS associated with it. I was thinking it might be worth having a check that ensures all behavioral components have a corresponding `system_test` written. 

This is beneficial because it would ensure that we have checks in place for how a component should behave, and would also make sure that axe is ran against various states of the component (e.g. when dialog is open). This is also extra important because we don't seem to have custom element tests. This would help surface more accessibility or behavioral issues at the testing stage.

### DIFF
--- a/.changeset/rich-donkeys-beam.md
+++ b/.changeset/rich-donkeys-beam.md
@@ -3,3 +3,5 @@
 ---
 
 Fix accessibility bug of missing accessible name on `Primer::Alpha::Dialog`
+
+<!-- Changed components: Primer::Alpha::Dialog -->
--- a/.changeset/rich-donkeys-beam.md
+++ b/.changeset/rich-donkeys-beam.md
@@ -2,4 +2,4 @@
 "@primer/view-components": minor
 ---
 
-Fix accessibility bug of missing accessible name on Dialog
+Fix accessibility bug of missing accessible name on `Primer::Alpha::Dialog`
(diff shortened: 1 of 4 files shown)


TylerJDev (approved):
Nice! Thanks for this fix! ✨

khiga8:
I am not authorized so please merge in, thank you!

PR #2245, opened 2023-09-13
Exclude hidden menu items from focus group

TylerJDev:

Currently, if `ActionMenu` contains items that are hidden via `hidden: true`, these items will still be present in the list of focusable items. This causes the focus trap to be blocked, as it tries to focus onto a item that is currently hidden from view.

This PR adds onto the selector to ensure that the parent is not `[hidden]`.

### Screenshots

<details>
<summary>Before PR; video shows focus trapped being blocked in overflow menu within ActionBar</summary>

https://github.com/primer/view_components/assets/26746305/952afe2d-14ab-4738-a208-586af181cb88

</details>

<details>
<summary>Current PR; video shows focus trap looping properly in overflow menu</summary>

https://github.com/primer/view_components/assets/26746305/680eb269-7e2e-4ddc-9451-91ef55c7400e

</details>

Relates to https://github.com/github/primer/issues/1131

#### Risk Assessment
  

Adds onto the menu item selector to exclude items that are hidden via `[hidden]`.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/silent-ravens-hide.md
+++ b/.changeset/silent-ravens-hide.md
@@ -1,0 +1,7 @@
+---
+"@primer/view-components": patch
+---
+
+Fix bug in ActionMenu keyboard navigation, where items aren't skipped when they're functionally `[hidden]`.
+
+<!-- Changed components: Primer::Alpha::ActionMenu -->
--- a/app/components/primer/alpha/action_menu/action_menu_element.ts
+++ b/app/components/primer/alpha/action_menu/action_menu_element.ts
@@ -9,7 +9,8 @@ type SelectedItem = {
   element: Element
 }
 
-const menuItemSelectors = ['[role="menuitem"]', '[role="menuitemcheckbox"]', '[role="menuitemradio"]']
+const validSelectors = ['[role="menuitem"]', '[role="menuitemcheckbox"]', '[role="menuitemradio"]']
+const menuItemSelectors = validSelectors.map(selector => `:not([hidden]) > ${selector}`)
 
 @controller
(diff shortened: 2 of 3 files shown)


TylerJDev:
Would love your opinion here @camertron. 

Focus trap [currently breaks if there's an item in the group that's hidden](https://primer.style/view-components/lookbook/inspect/primer/alpha/action_bar/playground). Since we use `[hidden]` to explicitly hide elements, I added it as something that isn't included within the list of valid focusable elements. This only effects controls that have their direct parent hidden (e.g. 

```html
<!-- The `menuitem` is not included in the focus group due to its direct parent being hidden via `[hidden]` -->
<div hidden>
  <button role="menuitem">Menu 1</button>
</div>
```

If this is a valid solution, then we might want to refine the selectors further to ensure no hidden/non-focusable element is included in the group.

camertron (approved):
Interesting, nice catch! I think this is definitely a valid solution, since the hidden items can't be interacted with in any way. What additional refinements were you considering?

jonrohan (approved):
Fix works great! Ship it 🚀

TylerJDev:
> What additional refinements were you considering?

I think there may be cases where we could also include the selector on the individual menu items themselves, but I'm not entirely sure if you can directly apply `[hidden]` to the menu item, rather than just the parent? We also might want to include `aria-hidden`, but I haven't seen a use case of this yet, so I think this is good for now!

Issue #2318, opened 2023-10-19
primer/form_helper (LoadError) with the latest version of Rails. (7.1.1)


safeforge:
I attempted to utilize the gem with the most recent version of Rails (7.1.1), but it appears there is an issue with files loading. Below, I outline the steps to replicate the issue:

```
rails new buggy --css postcss --js esbuild -d sqlite3;
bundle add primer_view_components;
echo 'require "view_component"\nrequire "primer/view_components/engine"' >> config/application.rb;
```

Running the application with the command  `rails s` raises an error in the browser

```
Puma caught this error: cannot load such file -- primer/form_helper (LoadError)
```

camertron:
Hey @safeforge, thanks for bringing this to our attention. I followed the steps you posted to create a new Rails app and was able to reproduce the behavior you're seeing. Curiously, the same behavior does _not_ appear in a new Rails 7.0 app. I've played around a bit trying to figure it out, and it looks like something changed in the way 7.1 adds to the Ruby load path. Even more curiously, the issue doesn't appear to happen in the github.com repo, which has been running 7.1 for months now. I'll try to get to the bottom of the issue today.

safeforge:
hey @camertron thank you for the prompt response. I have the same feeling, it's related to the load path upgrade introduced in Rails 7.1 and the new defaults. In fact, if you revert the defaults in the application from 7.1 to 7, it works correctly.

```ruby
  class Application < Rails::Application
    # Initialize configuration defaults for originally generated Rails version.
    config.load_defaults 7.0
  end
```
I attempted to activate each feature individually listed in `new_framework_defaults_7_1.rb`, and the application still works properly. The issue only arises when the application starts with the 7.1 defaults.

camertron:
Ok, I think I've figured this out. Rails 7.1 no longer adds eager load paths to the Ruby load path, meaning you can no longer `require` files that have been added to Rails' eager load paths. Instead Zeitwerk autoloads the correct file whenever the corresponding constant is referenced. This has been possible for a long time in Rails, but 7.1 removed the ability to `require` at all. I have therefore removed the offending `require`. Here's the PR: https://github.com/primer/view_components/pull/2336

PR #2336, opened 2023-10-26
Add Rails 7.1 support

camertron:

This PR makes the changes necessary to support Rails 7.1.

### Integration

Fixes https://github.com/primer/view_components/issues/2318

#### Risk Assessment
  

The only thing necessary to support Rails 7.1 is to remove an unnecessary `require` statement in the Rails engine. Rails 7.1 no longer adds eager load paths to Ruby's load path, preferring instead to let Zeitwerk autoload missing constants.

### Accessibility


safeforge:
@camertron Removing the `require "primer/form_helper"` does not appear to be sufficient, but you have correctly identified the issue.

I'm experiencing issues with `Form` usage, and the error appears to be similar to the previous one: "cannot load such file -- primer/class_name_helper."  

Indeed, by removing `require "primer/class_name_helper"` from the file `lib/primer/forms/base_component.rb`, the issue with my application appears to be resolved. I wonder if there are any other `require` causing errors.

camertron:
@safeforge interesting, thanks for the heads-up. Let me see if I can get our test suite to fail, that would really help catch these issues.

camertron:
Ok, I totally forgot to bump the framework defaults in the demo app's application.rb. Doing so uncovered several additional bad `require` statements that I have since removed.

@safeforge can you verify this addresses the issues you're seeing?

.github/workflows/test.yml
```diff
@@ -51,6 +51,8 @@ jobs:
             ruby_version: "2.7"
```
jonrohan, line 51: How do you feel about also dropping this matrix?

camertron: I'm ok with that 👍

jonrohan: I like the simplicity of having 3 groups, and adding a new one makes me want to drop the old one. I don’t know what sorta social contract we have to the community to always test against 6.1. I say let's go for it cause it lets us move our dependencies forward

camertron: Decision: leave 6.1 support in as long as it's easy. Remove it as soon as it's inconvenient.

jonrohan (approved):
👍🏻

safeforge:
> @safeforge can you verify this addresses the issues you're seeing?

@camertron, I can confirm that your fix has resolved the issue in my use case. Thank you!

PR #2411, opened 2023-11-30
ensure scroll position does not change when opening actionmenu

keithamus:

Sometimes opening the actionmenu causes the page to jump to the top, due to subtle timing issues with the anchored element positioning itself; it starts at the top of the page and only positions itself when visible. The actionmenu focusses the first element and if the menu hasn't had time to position itself, the page will jump. This fixes that.

### Integration

No

Closes https://github.com/github/primer/issues/2897, https://github.com/github/primer/issues/2900

#### Risk Assessment
  

I moved the focus to the `toggle` event, which is fired _after_ the popover is opened. `anchored-position` positions the element on a `beforetoggle` event which is always fired before the element becomes visible, so the position should be correct and the page won't jump.

No

### Accessibility

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/app/components/primer/alpha/action_menu/action_menu_element.ts
+++ b/app/components/primer/alpha/action_menu/action_menu_element.ts
@@ -107,6 +107,7 @@ export class ActionMenuElement extends HTMLElement {
     this.addEventListener('mouseover', this, {signal})
     this.addEventListener('focusout', this, {signal})
     this.addEventListener('mousedown', this, {signal})
+    this.popoverElement?.addEventListener('toggle', this, {signal})
     this.#setDynamicLabel()
     this.#updateInput()
     this.#softDisableItems()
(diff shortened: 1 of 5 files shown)


app/components/primer/alpha/action_menu/action_menu_element.ts
```diff
@@ -261,7 +261,7 @@ export class ActionMenuElement extends HTMLElement {
       this.#hide()
     } else {
       this.#show()
-      this.#firstItem?.focus()
+      setTimeout(() => this.#firstItem?.focus(), 10)
```
orhantoy, line 264: `this.#show()` just calls `showPopover` and in the related [docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/showPopover) it says

> When showPopover() is called on an element with the [popover](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover) attribute that is currently hidden, a [beforetoggle](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforetoggle_event) event will be fired, followed by the popover showing, and then the [toggle](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/toggle_event) event firing. If the element is already showing, an error will be thrown.

That sounds like we could wait for the `toggle` event, no? But I'm not sure as that would depend on not just if the popover is shown but also that the position has been set.

jonrohan (approved):

PR #2477, opened 2024-01-02
Propose removing the validate classname check that forces system arguments instead of utility classes

jonrohan:
_Authors: Please fill out this form carefully and completely._

_Reviewers: By approving this Pull Request you are approving the code change, as well as its deployment and mitigation plans._
_Please read this description carefully. If you feel there is anything unclear or missing, please ask for updates._

### What are you trying to accomplish?

This removes the dev/test environment warning that we put in place to force systems arguments instead of using utility classes. I think at the time we had a vision that we would control these more tightly, but we ended up making more friction for our selves. [example](https://github.com/github/github/pull/305716/commits/e92015e3cea902c4494e6ed148bd4b491e0fbba9)

### Integration

No

#### Risk Assessment

- [x] **Low risk** the change is small, highly observable, and easily rolled back.
- [ ] **Medium risk** changes that are isolated, reduced in scope or could impact few users. The change will not impact library availability.
- [ ] **High risk** changes are those that could impact customers and SLOs, low or no test coverage, low observability, or slow to rollback.

### Anything you want to highlight for special attention from reviewers?

This pr doesn't change support for system arguments only removes the barrier of using classnames instead.

### Accessibility

- **Fixes axe scan violation** - This change fixes an existing [axe scan](https://thehub.github.com/epd/engineering/dev-practicals/frontend/accessibility/readiness-routine/development/#axe-scans) violation.
- **No new axe scan violation** - This change does not introduce any new [axe scan](https://thehub.github.com/epd/engineering/dev-practicals/frontend/accessibility/readiness-routine/development/#axe-scans) violations.
- **New axe violation** - This change introduces a new [axe scan](https://thehub.github.com/epd/engineering/dev-practicals/frontend/accessibility/readiness-routine/development/#axe-scans) violation. Please describe why the violation cannot be resolved below.

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Lookbook)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2510, opened 2024-01-16
Revert "Propose removing the validate classname check that forces system arguments instead of utility classes"

jonrohan:
Reverts primer/view_components#2477


camertron:
Uh oh, what happened?

jonrohan:
We never had the chance to discuss it in ui club. So wanted to revert until we can

PR #2575, opened 2024-02-05
Lint for `stopPropagation` calls in JavaScript

camertron:

There was [an incident](https://github.com/github/availability/issues/2935) last week that resulted in dotcom's global nav list appearing to load forever. The issue was caused by adding a `stopPropagation()` call to the dialog invoker's `click` event handler, which prevented dotcom's own `click` handler from running.

Except in very specific circumstances, we should not be swallowing events using `stopPropagation`. See [this article](https://css-tricks.com/dangers-stopping-event-propagation/) for an explanation. This PR adds eslint rules for both `stopPropagation` and `stopImmediatePropagation`. The idea here isn't to stop us from ever using these functions, but to make sure we consider the ramifications of doing so before disabling the lint rule.

This PR also disables the lint rule for all the places we currently call `stopPropagation` and `stopImmediatePropagation`. I will file a separate issue for auditing these existing usages.

### Integration

No changes necessary in production; this change is dev-only.

Fixes https://github.com/github/primer/issues/3034

#### Risk Assessment
  

### Accessibility

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,7 @@
   ],
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
-    "ecmaVersion": "2022",
+    "ecmaVersion": 2022,
     "sourceType": "module"
   },
   "plugins": [
@@ -38,7 +38,7 @@
       ],
       "parser": "@typescript-eslint/parser",
       "parserOptions": {
-        "ecmaVersion": "2022",
+        "ecmaVersion": 2022,
         "sourceType": "module"
       },
       "plugins": [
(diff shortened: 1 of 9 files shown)


keithamus (commented):
`stopPropagation` is a very useful tool. I'm not sure we should be adding this much friction to usage of it, but I'd be happy to hear from others.

camertron:
@keithamus yeah, I understand where you're coming from, it is absolutely the right tool for the job in certain scenarios. Personally tho I don't feel like adding a disable is very much friction. It's designed to make us pause momentarily and maybe ask about the need for `stopPropagation()` when reviewing code. Plus, if we feel like it's too onerous, we can always remove the linter; it's only a few lines of config.

PR #2576, opened 2024-02-07
Allow Banner component to receive focus, adds optional `id` parameter

lindseywild:

**Note: I am mostly looking for feedback on this approach before I test it fully 🙏🏻** 

On the Accessibility team, we've discovered a few audit issues are marked as blocked by Primer. After more investigation, we realized this is not the case, and that in some scenarios, we want to move focus to the Banner component when a user takes an action. We can do this in dotcom with JS, as most consumers have logic to show/hide the banners based on certain conditions.

This PR adds a `focusBanner()` method to allow consumers to hook into to focus the banner in certain scenarios. The element we want to focus on in a Banner is subject to change, so this approach allows that element to be managed at the Primer level. It also allows an optional `id` parameter to be passed in which is applied to the `<x-banner>` element (currently passing in `id` as a system arg applies it to the nested `<div>` under `x-banner`, so calling a method on the ID won't work).

### Integration

No

* https://github.com/github/accessibility-audits/issues/5108
* https://github.com/github/accessibility-audits/issues/5090
* https://github.com/github/accessibility-audits/issues/5492

#### Risk Assessment
  

I chose to add a method to add the `tabindex` and `.focus()` functionality to focus the banner. @khiga8 and I considered adding a parameter to the ruby component, such as `focus_on_show`, that would hide the banner by default, and when shown, it would focus the banner.

The reasons we didn't go with this approach right now (it may be a good thing to discuss in the future!):

* Consumers currently handle the logic for hiding/showing the banner. We'd need to build this in at the component-level and update instances where a consumer is showing/hiding their banner
* It'd be more work to implement on the dotcom consumer end
* It'd be a larger time commitment on our end and Primer's (for testing, reviewing, validating, etc)
* It'd be a more risky change

A consumer could use this new functionality by calling the method this way:

```js
document.querySelector("#banner_id").focusBanner();
```

Is this an anti-pattern? Should we invest more time in allowing a consumer to do something like:

```rb
<%= render(Primer::Alpha::Banner.new(hidden: :true, focus_on_show: :true)) do %>
  Please update your billing information in order to add a payment method.
<% end %>
```

and then the component would be responsible for showing/hiding itself? I think we'd still need to have an event listener present to know when to show/hide the component... Open to other ideas here!

### Accessibility

Allows the banner to be focusable, which will ensure screen reader announcements are consistent in certain scenarios when relevant

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/app/components/primer/alpha/banner.rb
+++ b/app/components/primer/alpha/banner.rb
@@ -68,13 +68,15 @@ module Primer
       # @param description [String] Description text rendered underneath the message.
       # @param icon [Symbol] The name of an <%= link_to_octicons %> icon to use. If no icon is provided, a default one will be chosen based on the scheme.
       # @param scheme [Symbol] <%= one_of(Primer::Alpha::Banner::SCHEME_MAPPINGS.keys) %>
+      # @param id [String] an optional ID a user can pass in to target the `x-banner` component.
       # @param system_arguments [Hash] <%= link_to_system_arguments_docs %>
-      def initialize(tag: DEFAULT_TAG, full: false, full_when_narrow: false, dismiss_scheme: DEFAULT_DISMISS_SCHEME, dismiss_label: DEFAULT_DISMISS_LABEL, description: nil, icon: nil, scheme: DEFAULT_SCHEME, **system_arguments)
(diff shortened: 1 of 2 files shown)


keithamus:
```ruby
<%= render(Primer::Alpha::Banner.new(hidden: :true, focus_on_show: :true)) do %>
  Please update your billing information in order to add a payment method.
<% end %>
```

This seems like a better idea to me. With an `IntersectionObserver` it would be quite straightforward to implement, and would avoid adding new public API to the web components, which I'd like to avoid unless there was a _very compelling_ reason to do so.

jonrohan (approved):

PR #2583, opened 2024-02-09
[WIP] Support prop to announce banner when it is shown

khiga8:
This is a **proof of concep**t, and has a pre-req to ship:

* https://github.com/primer/view_components/pull/2625
  * `<live-region>` should be part of the page layout of the consuming application. We probably want a PVC wrapper for this like`Primer::Alpha::LiveRegion`, and have guidance that nudges consumers to add it to their page layout.

When banners are shown async to communicate feedback, we should either be focusing or announcing the banner content to draw attention to it. Read more at [Banner docs](https://primer-0fe4ab137c-26441320.drafts.github.io/components/banner).

This is a proof of concept to support a `announce_on_show` prop that, when set, **announces** the banner content upon it's visibility being updated by the consumer.

This should support a live region announcement being made across the following scenarios:

* The banner is rendered as hidden. Then the consumer uses JS to unhide the banner. At this point, we want the banner to be announced when `announce_on_show: true`. (Test written ✅ )
* A parent of the banner is set to hidden (e.g. an entire container). Then the consumer uses JS to unhide the container. At this point, we want the banner to be announced when `announce_on_show: true` is set. (Test written ✅ )
* The banner gets dynamically injected into the DOM via JS. At this point, we want the banner to be announced when `announce_on_show: true`. (Needs test ❌  Any tips for how to go about dynamically injecting a view component in our system tests? )

#### IMPORTANT 🔈 

**Dynamically injected live region attributes don't work reliably** so we cannot just set `aria-live` on the Banner and expect it to announce when the Banner appears. See [Staff only: Challenges with dynamically injected live region](https://github.com/github/accessibility/blob/main/docs/coaching-recommendations/toast-flash-banner/accessible-banner-prototype.md#challenges-with-dynamically-inserted-live-region).

This approach depends on hooking into [live-region-element](https://github.com/primer/live-region-element/) which is ideally already part of the page layout. As called out in the beginning, there is pre-req work necessary to support live-region. Consumer needs to add this to their page layout.

### Integration

This depends on `<live-region>` to be part of the page layout of the repos it is called in.

 Relates to: https://github.com/github/accessibility/issues/5649

#### Risk Assessment
  

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


demo/app/views/layouts/component_preview.html.erb
```diff
@@ -33,5 +33,7 @@
         </div>
       <% end %>
     </div>
+    <%# TODO: Use live-region-element %>
+    <div aria-live="polite" aria-atomic="true" class="sr-only">
```
khiga8, line 37: I'm thinking we should embed [live-region-element](https://github.com/primer/live-region-element/) within the lookbook preview once it's available!

app/components/primer/alpha/x_banner.ts
```diff
@@ -1,8 +1,35 @@
...
+      const callback = (entries: IntersectionObserverEntry[]) => {
+        for (const entry of entries) {
+          if (entry.isIntersecting || entry.intersectionRatio > 0) {
+            // Uncomment when available: https://github.com/primer/live-region-element/
+            // announceFromElement(this.banner)
+            this.observer?.disconnect()
+          }
+        }
+      }
+      this.observer = new IntersectionObserver(callback, options)
```
khiga8, line 26: Looked into this because it was mentioned in https://github.com/primer/view_components/pull/2576#issuecomment-1933983766, but I dont know if its handling all the scenarios and if it'll work the way we expect yet.

Needs further testing!

app/components/primer/alpha/x_banner.ts
```diff
@@ -1,8 +1,39 @@
 import {controller, target} from '@github/catalyst'
+// eslint-disable-next-line import/no-unresolved
+import '@primer/live-region-element/define'
```
khiga8, line 3: How to fix this? Also, I noticed we don't need to make this call in other custom elements since it's exported as part of the index. Wonder if we should do the same for `live-region`?

previews/primer/alpha/banner_preview.rb
```diff
@@ -24,6 +23,16 @@ def default
...
+      # @label Announce
+      def announce_on_show
+        render_with_template
+      end
+
+      # @label Announce when updating wrapper visibility
+      def announce_on_show_with_hidden_wrapper
+        render_with_template
+      end
+
```
khiga8, line 35: I want to add a test for when `Primer::Alpha::Banner` is dynamically injected. How should I set this up?

test/system/alpha/banner_test.rb
```diff
@@ -21,5 +21,27 @@ def test_dismiss_button_hide
...
+      visit_preview(:announce_on_show_with_hidden_wrapper)
+
+      assert_selector("#wrapper", visible: :hidden)
+      assert_equal page.evaluate_script("document.querySelector('live-region').shadowRoot.querySelector('[aria-live=\"polite\"]').innerText"), ""
+
+      find("#toggle-wrapper").click
+
+      assert_selector("#wrapper", visible: :visible)
+      assert_equal page.evaluate_script("document.querySelector('live-region').shadowRoot.querySelector('[aria-live=\"polite\"]').innerText.trim()"), "Hello."
+    end
```
khiga8, line 45: I want to add a test for when `Primer::Alpha::Banner` is dynamically injected. How should I set this up?

PR #2625, opened 2024-02-21
Add live-region-element package with a PVC wrapper

khiga8:

Prereq to https://github.com/primer/view_components/pull/2583

This PR adds:

* The `@primer/live-region-element` package to PVC.
* A wrapper component for this element, `Primer::Alpha::LiveRegionAnnouncement` which should act as a global live region. This component should be added to the application layout of the consuming application.

### Why

Live regions can be finicky. They do not work well when they are dynamically injected. So for example, if a user does something resulting in a success banner to be shown on a page async, having `aria-live="polite"` on the banner will not result in an announcement. See [Staff only: Challenges with dynamically inserted live region](https://github.com/github/accessibility/blob/main/docs/coaching-recommendations/toast-flash-banner/accessible-banner-prototype.md#challenges-with-dynamically-inserted-live-region) and https://github.com/github/primer/discussions/2942 for more information.

There are ways to get around this – such as by having a global live region that is reliably part of the page layout and populating that to make an announcement. We rely on this solution in dotcom as part of the [application layout](https://github.com/github/github/blob/master/app/views/layouts/_document.html.erb#L40) with [aria-live.ts](https://github.com/github/github/blob/master/ui/packages/aria-live/aria-live.ts). We want Primer components to be able to hook into this global live region. In addition, other applications may find this global live region utility useful.

@joshblack proposed and published a [live-region-element](https://github.com/primer/live-region-element) as a solution that encapsulates these methods in `aria-live.ts` and would allow Primer components across PVC and Primer React to hook into it. 

### Integration

No

#### Risk Assessment
  

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


app/components/primer/alpha/live_region_announcement.html.erb
```diff
@@ -0,0 +1,16 @@
...
+        clip-path: inset(50%);
+        height: 1px;
+        overflow: hidden;
+        position: absolute;
+        white-space: nowrap;
+        width: 1px;
+      }
+    </style>
+    <div id="polite" aria-live="polite" aria-atomic="true"></div>
+    <div id="assertive" aria-live="assertive" aria-atomic="true"></div>
```
khiga8, line 14: I noticed the `announce` utility depends on these `id="polite"` and `id="assertive"` being set as it calls `document.querySelector`. If there are multiple `<live-region>` element on a page (e.g. because there is a modal), this results in duplicate IDs. Is this problematic? @joshblack 🤔

joshblack: Hey @khiga8! Are these for the `id="polite"` and `id="assertive"` values specifically? If so, these are in the Shadow DOM and won't conflict with other live regions on the page as a result.

app/components/primer/alpha/live_region_element.ts
```diff
@@ -0,0 +1,3 @@
+import '@primer/live-region-element'
+// eslint-disable-next-line import/no-unresolved
+import '@primer/live-region-element/define'
```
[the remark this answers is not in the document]
khiga8: We chatted about this. It's for SSR things, but we might not need to do this?

khiga8:
I am putting this on pause until we've had a chance to validate `live-region` in dotcom per my chat with @joshblack!

PR #2643, opened 2024-02-27
Swap bg-gradient with bg-color for Banner/Flash

langermank:
Replace `background-image` gradient with `background-color` as the gradient used a single color value.


jonrohan (approved):
Was there some reason we were doing this? weird cause linear-gradient was going from the same color to the same color without really doing a gradient

PR #2666, opened 2024-03-06
Make sure labels and form inputs are properly associated when a custom ID is provided

camertron:

This PR fixes the association between `<label>`s and their form inputs when a custom ID is provided. Normally, Rails will generate an ID for a given input, and use the same ID for the label. However, creating the input with a custom ID that is not also passed to the label results in the label's `for` attribute pointing to a non-existent element. For example,

```ruby
<%= render(Primer::Alpha::TextField(name: :foo, id: "bar", label: "Foobar")) %>
```

results in the following HTML:

```html
<label for="foo">Foobar</label>
<input type="text" name="foo" id="bar" />
```

The forms framework (via Rails) appears to link the label to the input's name instead of its ID, which is incorrect.

This PR ensures the two elements are always properly associated.

### Integration

#### Risk Assessment
  

@kendallgassner and I discovered we need to pass the given ID to the label's `for:` argument. This effectively bypasses Rails' default ID generation code.

### Accessibility


kendallgassner (approved):
💖 thanks for adding a test too!

jonrohan (approved):

keithamus (approved):

PR #2670, opened 2024-03-06
Give ownership of screenshots to design-reviewers

jonrohan:
Follow up to https://github.com/primer/view_components/pull/2664

This gives the playwright screenshots ownership to @primer/design-reviewers My thoughts are that if the style changes then design should 👓


langermank (approved):
I love this! Can we do this in Primer React as well?

PR #2750, opened 2024-04-05
Allow `:link` Button scheme text label to wrap

langermank:
As a button styled like a link, the label text should be allowed to wrap.


joshblack (approved):

previews/primer/beta/button_preview/link_scheme_label_wrap.html.erb
```diff
@@ -0,0 +1,8 @@
+<div style="width: 200px; background: pink;">
```
langermank, line 1:
```suggestion
<div style="width: 200px;">
```

jonrohan (approved):
🎉

PR #2751, opened 2024-04-05
Add `label_wrap` param to `Beta::Button`

langermank:
- Adds a new param that allows button label text to wrap
- Fixes overflow for `SegmentedControl`

After testing this change behind in feature flag in dotcom, we've come to the conclusion that the change is too risky. Instead, we're introducing it with a param so we can target specific areas with overflow issues.

Tested in dotcom:

change username button with text wrapping


previews/primer/beta/button_preview.rb
```diff
@@ -39,9 +41,10 @@ def playground(
                  align_content: align_content,
                  tag: tag,
                  disabled: disabled,
-                 inactive: inactive
+                 inactive: inactive,
+                 label_wrap: label_wrap
                )) do |_c|
-          "Button"
+          "Button heheheh whoaaaa wowowow wowowo omg"
```
langermank, line 47:
```suggestion
          "Button"
```

😆

joshblack (approved):

camertron (approved):

PR #2843, opened 2024-05-15
Add live region to announce "Copied!" on ClipboardCopy component

lindseywild:

Adds a live region to the ClipboardCopy component so that by default, "Copied!" is announced when a user successfully copies something.

#### Before
Screen recording of the ClipboardCopy button on Lookbook, when pressing "Enter" or "Control + Option + Space", no feedback is announced but there is a visual icon change from 2 pages stacked on top of each other to a green check mark

https://github.com/primer/view_components/assets/35239154/5221b6e6-f399-461c-92f2-9f1cf512e2b9

#### After
Screen recording of the ClipboardCopy button on Lookbook, when pressing "Enter" or "Control + Option + Space", "Copied!" is announced

https://github.com/primer/view_components/assets/35239154/3f6db4af-c313-4aa9-bd5e-1a33b630059e

### Integration

No

Related to https://github.com/github/accessibility-audits/issues/6469 & https://github.com/github/accessibility/issues/6096.

#### Risk Assessment
  

I am adding a live region to the ClipboardCopy button so that, by default, screen reader feedback will be announced.

I considered adding a live region to the [`clipboard-copy-element` web component](https://github.com/github/clipboard-copy-element) which is used under the hood, but changing that component to include a live region would be a breaking change since we'd need to wrap the current inner text with another element so that we can have a live region as a sibling. [Nesting a live region inside of a `<button>` element has poor support](https://github.com/github/accessibility/blob/main/docs/wiki/screen-reader-testing/dynamically-updated-buttons-support-april-2024.md#6-bonus-setting-a-nested-polite-span-live-region-inside-the-button) and is not recommended.

* Is there a better way to wrap the component?
* Is there a better way than looking for the `parentNode` of the `target` element to target the specific aria-live region for one component (thinking if there are multiple ClipboardCopy elements on the page and not wanting them to conflict)? I didn't want to change the API too much by making the wrapper I added the new `<clipboard-copy>` component but maybe it's ok?

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/app/components/primer/beta/clipboard_copy.ts
+++ b/app/components/primer/beta/clipboard_copy.ts
@@ -37,7 +37,7 @@ document.addEventListener('clipboard-copy', ({target}) => {
   if (!target.hasAttribute('data-view-component')) return
 
   const currentTimeout = clipboardCopyElementTimers.get(target)
-  const clipboardCopyLiveRegion = document.querySelector<HTMLDivElement>('[data-clipboard-copy-feedback]')
+  const clipboardCopyLiveRegion = target.parentNode?.querySelector<HTMLDivElement>('[data-clipboard-copy-feedback]')
   const copiedAnnouncement = 'Copied!'
 
   if (currentTimeout) {
--- a/app/components/primer/beta/clipboard_copy.ts
+++ b/app/components/primer/beta/clipboard_copy.ts
@@ -37,7 +37,7 @@ document.addEventListener('clipboard-copy', ({target}) => {
   if (!target.hasAttribute('data-view-component')) return
 
   const currentTimeout = clipboardCopyElementTimers.get(target)
-  const clipboardCopyLiveRegion = target.parentNode?.querySelector<HTMLDivElement>('[data-clipboard-copy-feedback]')
+  const clipboardCopyLiveRegion = target.parentNode?.querySelector<HTMLElement>('[data-clipboard-copy-feedback]')
   const copiedAnnouncement = 'Copied!'
 
   if (currentTimeout) {
--- a/app/components/primer/beta/clipboard_copy.ts
+++ b/app/components/primer/beta/clipboard_copy.ts
@@ -37,18 +37,35 @@ document.addEventListener('clipboard-copy', ({target}) => {
   if (!target.hasAttribute('data-view-component')) return
 
   const currentTimeout = clipboardCopyElementTimers.get(target)
+  const clipboardCopyLiveRegion = document.querySelector<HTMLDivElement>('[data-clipboard-copy-live-region]')
+  const copiedAnnouncement = 'Copied!'
 
   if (currentTimeout) {
     clearTimeout(currentTimeout)
     clipboardCopyElementTimers.delete(target)
   } else {
     showCheck(target)
+    if (clipboardCopyLiveRegion) {
+      if (clipboardCopyLiveRegion.textContent === copiedAnnouncement) {
+        /* This is a hack due to the way the aria live API works.
+    A screen reader will not read a live region again
+    if the text is the same. Adding a space character tells
+    the browser that the live region has updated,
+    which will cause it to read again, but with no audible difference. */
+        clipboardCopyLiveRegion.textContent = `${copiedAnnouncement}\u00A0`
+      } else {
+        clipboardCopyLiveRegion.textContent = copiedAnnouncement
+      }
+    }
   }
 
   clipboardCopyElementTimers.set(
     target,
     setTimeout(() => {
       showCopy(target)
+      // if (clipboardCopyLiveRegion) {
+      //   clipboardCopyLiveRegion.textContent = ''
+      // }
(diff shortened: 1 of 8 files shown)


app/components/primer/beta/clipboard_copy.ts
```diff
@@ -37,12 +37,26 @@ document.addEventListener('clipboard-copy', ({target}) => {
   if (!target.hasAttribute('data-view-component')) return
 
   const currentTimeout = clipboardCopyElementTimers.get(target)
+  const clipboardCopyLiveRegion = target.parentNode?.querySelector<HTMLElement>('[data-clipboard-copy-feedback]')
```
camertron, line 40: Can we use a Catalyst target for this?

app/components/primer/beta/clipboard_copy.ts
```diff
@@ -37,12 +37,26 @@ document.addEventListener('clipboard-copy', ({target}) => {
...
     clipboardCopyElementTimers.delete(target)
   } else {
     showCheck(target)
+    if (clipboardCopyLiveRegion) {
+      if (clipboardCopyLiveRegion.textContent === copiedAnnouncement) {
+        /* This is a hack due to the way the aria live API works.
+    A screen reader will not read a live region again
+    if the text is the same. Adding a space character tells
+    the browser that the live region has updated,
+    which will cause it to read again, but with no audible difference. */
```
camertron, line 54:
```suggestion
       A screen reader will not read a live region again
       if the text is the same. Adding a space character tells
       the browser that the live region has updated,
       which will cause it to read again, but with no audible difference. */
```

test/system/beta/clipboard_copy_button_test.rb
```diff
@@ -19,7 +19,7 @@ def test_copies_text
     def test_includes_tooltip
       visit_preview(:with_tooltip)
 
-      assert_selector ".Button + tool-tip", visible: :all
+      assert_selector "tool-tip[for='clipboard-button']", visible: :all
```
camertron, line 22: Can we add a system test that ensures the aria-live region gets populated with the "Copied!" text when the button is clicked?

camertron (changes requested):
Thanks! 🎉

app/components/primer/beta/clipboard_copy.ts
lindseywild: Sure, good call! Do I have to convert the rest of this file to a Catalyst component too? Or is there a way to just use the target without the rest of the component being a Catalyst component?

camertron: Oh I think the whole thing would need to be a Catalyst component. If that's too much work, then feel free to disregard 😄

app/components/primer/beta/clipboard_copy.ts
```diff
@@ -37,12 +37,26 @@ document.addEventListener('clipboard-copy', ({target}) => {
...
     clipboardCopyElementTimers.delete(target)
   } else {
     showCheck(target)
+    if (clipboardCopyLiveRegion) {
+      if (clipboardCopyLiveRegion.textContent === copiedAnnouncement) {
+        /* This is a hack due to the way the aria live API works.
+       A screen reader will not read a live region again
+       if the text is the same. Adding a space character tells
+       the browser that the live region has updated,
+       which will cause it to read again, but with no audible difference. */
```
camertron, line 54: Ughh sorry this is my bad 😞 

```suggestion
        /* This is a hack due to the way the aria live API works.
        A screen reader will not read a live region again
        if the text is the same. Adding a space character tells
        the browser that the live region has updated,
        which will cause it to read again, but with no audible difference. */
```

camertron (changes requested):

test/system/beta/clipboard_copy_button_test.rb
```diff
@@ -16,10 +16,22 @@ def test_copies_text
       assert_equal "Text to copy", clipboard_text
     end
 
+    def test_populates_live_region
+      visit_preview(:playground)
+
+      clipboard_text = capture_clipboard do
+        find("#clipboard-button").click
+      end
```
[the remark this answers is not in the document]
lindseywild: Thanks for this! I am not too great with Ruby so all explanations help 😄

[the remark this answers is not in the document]
lindseywild: This was odd, when I removed the variable assignment the test did not pass 🤔 I tried a few different things but it seemed like the click event wasn't being picked up... Not sure what was going on, but I added the assertion to the existing test so hopefully that is ok!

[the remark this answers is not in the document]
camertron: Oh weird! I wonder if `capture_clipboard` waits for the text to be copied or something?

camertron (approved):

PR #2884, opened 2024-06-04
Add sr-only option to spinner

owenniblock:

Add a default, screen reader accessible text to the Spinner component. This is set when no `aria-label` or `aria-labelledby` is present.

Also when an `aria-label` is not set on the Spinner component, the `aria-hidden` attribute is set to `true`.

### How did you accomplish it?

I added a new `sr_text` argument to the `Spinner` component. This argument is set to a default value of "Loading" if no `aria-label`. The `aria-hidden` attribute is set to `true` when an `aria-label` is not present.

### Integration
No

Relates to https://github.com/github/primer/issues/2202

#### Risk Assessment
  

I opted for a change which would allow existing aria-labels to still work.

I would love to know if we feel like it's necessary to be able to customise the sr_text. Do we just have the default if no aria-label is defined?

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [x] Tested in Edge


app/components/primer/beta/spinner.html.erb
```diff
@@ -2,3 +2,6 @@
   <circle cx="8" cy="8" r="7" stroke="currentColor" stroke-opacity="0.25" stroke-width="2" vector-effect="non-scaling-stroke" fill="none" />
   <path d="M15 8a7.002 7.002 0 00-7-7" stroke="currentColor" stroke-width="2" stroke-linecap="round" vector-effect="non-scaling-stroke" />
 <% end %>
+<% if no_aria_label? %>
+  <span class="sr-only"><%= @sr_text %></span>
```
mperrotti, line 6: I think this needs to be announced in a live region. You can use https://github.com/primer/live-region-element

This was discussed a while ago in this discussion: https://github.com/github/primer/discussions/2976

Let me know if I misunderstood and I'll revise https://github.com/primer/react/pull/4140 to align with this PR.

owenniblock: @mperrotti the idea is to eventually move to an announced version, however this first step will resolve [the audit issue](https://github.com/github/primer/issues/2202).

I felt like the announcement API needs some consideration (for example we don't want many Spinners on one page to all announce individually) and since the audit issues them. We could certainly add announcements (on this PR if we need to do so to get it released) however I would not feel comfortable if the Spinner announced by default, I think that's too risky given the number of Spinners we have in the wild already.

TylerJDev (approved):
LGTM! 

> I would love to know if we feel like it's necessary to be able to customise the sr_text. Do we just have the default if no aria-label is defined?

I feel like in cases where the spinner is not unique and/or what is being loaded should be defined is a good use case for sr_text customization. Outside of that I think the default "Loading" is fair more most cases.

PR #2964, opened 2024-07-23
Add `turbo: true` as an option for ToggleSwitch

### LINKED ISSUE #2940: `ToggleSwitchForm` does not work with Turbo Responses
Hi there,

in our application we are currently introducing a `ToggleSwitchForm` and we want the response to change a bit of state in the page. We are using turbo streams. The response we get is a proper turbo-stream response but since the request does not originate from a proper form request but instead a fetch request, turbo does not trigger.

See 

https://github.com/primer/view_components/blob/ae8dbe23040ded74fab853ac3dc2068711c625f1/app/components/primer/alpha/toggle_switch.ts#L161-L172

In my opinion we have 2 options to tackle this:

1. We allow to pass a parameter into the ToggleSwitchForm that adds a data attribute to tell the component that it is rendered in a turbo context and then change the `fetch` request to include `headers: { Accept: "text/vnd.turbo-stream.html" }` and when we have the response we trigger `Turbo.renderStreamMessage(response.text())`. But I am not sure if you want to tie your implementation to Turbo this closely.
2. Another option would be that instead of building a fetch request, we add hidden form inputs for the CSRF token and the `value`, fill those and then trigger a `form.submit()` to have the request be executed as a normal form submission.

I'll be happy to submit a PR for either solution if you let me know which one is the preferred way.

klaustopher:

As described in #2940, we are using a `ToggleSwitchForm` and in our case the response is a turbo stream that should replace some other elements in the page. The current implementation does not allow for that as the `fetch` request does not trigger any behavior associated with Turbo. To do that we need to explicitly set the `Accept` header to allow `text/vnd.turbo-stream.html` and when the response comes back, we need to pass it to `Turbo.renderStreamMessage` to be rendered. 

In #2940 we also discussed making the `ToggleSwitch` have a proper form, but while experimenting with this I found that this completely changes the existing behavior and I wanted this change to be as unobstrusive as it could be.

### Integration

The default is `turbo: false`, so nothing changes in existing code.

Closes #2940 

#### Risk Assessment

The code does not change any default behavior. One has to intentionally activate turbo with `turbo: true` as a param to the `ToggleSwitch` to get the new behavior. Also the response needs to provide the correct content type to be forwarded into the turbo renderer.

See above

There are currently no tests for the call to `Turbo.renderStreamMessage`, would be interested in any pointers how to add proper tests.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/hip-glasses-wave.md
+++ b/.changeset/hip-glasses-wave.md
@@ -1,0 +1,6 @@
+---
+"@primer/view-components": minor
+---
+
+Add `turbo: true` as an parameter for the `ToggleSwitch` component and treat the respoonse (when it has the correct MIME type) as a [Turbo Stream](https://turbo.hotwired.dev/handbook/streams). The new `turbo` paramater defaults to false
+and with that the HTTP response is simply ignored as is the current behavior.
--- a/app/components/primer/alpha/toggle_switch.rb
+++ b/app/components/primer/alpha/toggle_switch.rb
@@ -26,12 +26,14 @@ module Primer
       # @param enabled [Boolean] Whether or not the toggle switch responds to user input.
       # @param size [Symbol] What size toggle switch to render. <%= one_of(Primer::Alpha::ToggleSwitch::SIZE_OPTIONS) %>
       # @param status_label_position [Symbol] Which side of the toggle switch to render the status label. <%= one_of(Primer::Alpha::ToggleSwitch::STATUS_LABEL_POSITION_OPTIONS) %>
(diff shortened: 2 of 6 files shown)


jonrohan (approved):
This looks great! Thanks

camertron (approved):
Awesome, thank you!

klaustopher:
Will do tomorrow morning 👍

klaustopher:
Changeset is added

PR #2986, opened 2024-08-01
[Alpha][SelectPanel] Fix single select remote loading bug

kendallgassner:
### What

Fixes a remote single-select filtering has a bug where ...

When a user clicks an item `handleItemActivated `is called. In single select mode this function de-selects all items stored in `this.items`.

However, `this.items` contains only SelectPanel items that the server has returned to the SelectPanel. Meaning, if a user filters items using the input field the server may have returned a subset of items and therefore, we cannot deselect all items.

A similar problem will occur if a server has a limit and only returns a subset of the results to begin with.

Additionally, I fixed `removeSelecteditems` as it was checking for the `data-value` on the li as opposed to the button which has the actual value.

- https://github.com/github/primer/issues/3694
- 
- https://github.com/github/primer/issues/3694

#### Risk Assessment
  

- [X] **Low risk** the change is small, highly observable, and easily rolled back.

Added test cases to test functionality and mitgate regression 

### Merge checklist

- [X] Added/updated tests
- [X] Added/updated previews (Lookbook)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/dull-knives-leave.md
+++ b/.changeset/dull-knives-leave.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Fixes Alpha::SelectPanel remote loading bug where items from the server were overriding the users selections. Additionally, update the #removeSelectedItem function to grab the data-value from the correct element.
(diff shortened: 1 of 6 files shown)


camertron (approved):
🎉

PR #3011, opened 2024-08-16
Hide "loading" screen reader-only text after visual spinner leaves

JoyceZhu:
Fixes 2 bugs which are related, in closing the issue referenced below:
- `eventually_local` `SelectPanel`s don't have a `src` attribute, which they should according to the `SelectPanel` documentation.
- The `loading` screen reader text for spinners should be removed after the content has loaded/the visual spinner has been hidden so users don't discover it and get confused exploring post-load.

Closes https://github.com/github/primer/issues/3690

#### Risk Assessment
  

I manually tested this against the [Lookbook examples](https://primer-lookbook.github.com/view-components/lookbook/inspect/primer/alpha/select_panel/fetch_strategies) for different `fetch_strategies` to ensure there aren't surprise visual regressions.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


app/components/primer/alpha/select_panel.html.erb
```diff
@@ -13,7 +13,7 @@
             <%= render(Primer::BaseComponent.new(
               tag: :"remote-input",
               aria: { owns: @body_id },
-              **(@fetch_strategy == :remote ? { src: @src } : {}),
+              **(@fetch_strategy == :local ? {} : { src: @src }),
```
camertron, line 16: For `eventually_local` panels, loading is done via an `<include-fragment>`, which is why `src` isn't attached to the `<remote-input>` element in that case.

app/components/primer/beta/spinner.html.erb
```diff
@@ -3,5 +3,5 @@
   <path d="M15 8a7.002 7.002 0 00-7-7" stroke="currentColor" stroke-width="2" stroke-linecap="round" vector-effect="non-scaling-stroke" />
 <% end %>
 <% if no_aria_label? %>
-  <span class="sr-only"><%= @sr_text %></span>
+  <span class="spinner-screenreader-text sr-only"><%= @sr_text %></span>
```
camertron, line 6: Can we use a Catalyst `data-target` here instead of a class?

lib/primer/forms/primer_text_field.ts
```diff
@@ -97,11 +97,19 @@ export class PrimerTextFieldElement extends HTMLElement {
 
   showLeadingSpinner(): void {
     this.leadingSpinner?.removeAttribute('hidden')
+    const spinnerScreenreaderText = this.leadingSpinner?.nextElementSibling
```
camertron, line 100: Same question re: using a Catalyst target for this. Then we could do `this.spinnerScreenreaderText` or similar.

camertron (changes requested):
Nice, thanks! Just a few minor questions.

lib/primer/forms/primer_text_field.ts
JoyceZhu: Can that be referenced from the text field TS file, or will I need to set up a new corresponding TS file for the spinner component? The reasons I reverted to this class-based solution (besides being new to Catalyst) were 1) I wasn't sure if it'd be overkill to set up an entirely new file just for this screen reader only text, and 2) logging the `leadingSpinner`object only seemed to report the content inside the `svg` tag (the offending text is a sibling instead of a child), so I was unsure if setting Catalyst up for the spinner would be less convoluted than this.

camertron: Ahh ok I totally missed these changes were in the `Spinner` and `PrimerTextField` code. I agree it feels like overkill to define a custom element for `Spinner` just to expose a single DOM element.

What you've got here seems ok then. Although `nextElementSibling` makes me a little nervous 😅 Maybe we can do `this.leadingSpinner?.querySelector('.spinner-screenreader-text)?`.

app/components/primer/alpha/select_panel.html.erb
JoyceZhu: Ah, helpful -- I think the problem in that case is the spinner hiding function call only fires when the input itself gets the `loadend` event. Ended up just explicitly saying "hide the spinner" after the `include-fragment` is done, which I think is easier to follow than making `handleEvent` even bigger.

camertron: Ah ok that makes sense 👍

camertron (approved):
🎉

PR #3021, opened 2024-08-20
When hiding a `Spinner`, also hide its screen reader text

camertron:

Unfortunately I merged https://github.com/primer/view_components/pull/3011 a little too fast. This is a follow-up PR that adjusts the approach taken in the original PR by hiding/showing the entire `Spinner`, including the screen reader text.

### Integration

#### Risk Assessment
  

### Accessibility

This doesn't fix an axe scan violation, but it does address the following accessibility issue: https://github.com/github/primer/issues/3690

### Merge checklist

- [x] Added/updated tests
~- [ ] Added/updated previews (Lookbook)~
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/app/components/primer/beta/spinner.html.erb
+++ b/app/components/primer/beta/spinner.html.erb
@@ -1,7 +1,9 @@
-<%= render Primer::BaseComponent.new(**@system_arguments) do %>
-  <circle cx="8" cy="8" r="7" stroke="currentColor" stroke-opacity="0.25" stroke-width="2" vector-effect="non-scaling-stroke" fill="none" />
-  <path d="M15 8a7.002 7.002 0 00-7-7" stroke="currentColor" stroke-width="2" stroke-linecap="round" vector-effect="non-scaling-stroke" />
-<% end %>
-<% if no_aria_label? %>
-  <span class="spinner-screenreader-text sr-only"><%= @sr_text %></span>
+<%= render(Primer::BaseComponent.new(tag: :span, hidden: @hidden, data: { target: @target })) do %>
+  <%= render Primer::BaseComponent.new(**@system_arguments) do %>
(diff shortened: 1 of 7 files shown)


app/components/primer/beta/spinner.html.erb
```diff
@@ -1,7 +1,9 @@
-<%= render Primer::BaseComponent.new(**@system_arguments) do %>
-  <circle cx="8" cy="8" r="7" stroke="currentColor" stroke-opacity="0.25" stroke-width="2" vector-effect="non-scaling-stroke" fill="none" />
-  <path d="M15 8a7.002 7.002 0 00-7-7" stroke="currentColor" stroke-width="2" stroke-linecap="round" vector-effect="non-scaling-stroke" />
-<% end %>
-<% if no_aria_label? %>
-  <span class="spinner-screenreader-text sr-only"><%= @sr_text %></span>
+<%= render(Primer::BaseComponent.new(tag: :span, hidden: @hidden, data: { target: @target })) do %>
```
JoyceZhu, line 1: aaah, I'm guessing the `Primer::BaseComponent.new()` is the magic bit I needed to be able to access the screenreader-only element from other spinner files?

JoyceZhu (approved):

app/components/primer/beta/spinner.html.erb
camertron: Sort of, we just needed to wrap both the `<svg>` and the `<span class="sr-only">` in another `<span>` so we could hide them both together. It was a little tricky because usually we pass `**system_arguments` to the top-level element (eg. the new `<span>`), but that would break the existing API which puts `**system_arguments` on the `<svg>` element. I had to extract only certain attributes, eg. `data-target` and `hidden`, to be able to put them on the new `<span>` and pass everything else down to the `<svg>`.

PR #3044, opened 2024-08-30
[SelectPanel] Filter input border and loading/error panel min-height 

jamieshark:

This PR addresses a few design items towards the SelectPanel component (https://github.com/github/primer/issues/3875). Namely, it adds a divider below the filter input (if present), it moves the filter banner error to be above the input, and adds a min-height that corresponds to the overlay size for the loading / error states.

### Screenshots

Screenshot of loading state with min height
Screenshot of error state with min height
Screenshot of error banner
Screenshot of no results

- Closes https://github.com/github/primer/issues/3879
- Closes https://github.com/github/primer/issues/3881
- Closes https://github.com/github/primer/issues/3880
- Closes https://github.com/github/primer/issues/3934#issue-2514579172

#### Risk Assessment
  

### Accessibility

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


app/components/primer/alpha/select_panel.rb
```diff
@@ -388,7 +388,10 @@ def initialize(
         @title = title
         @system_arguments[:tag] = :"select-panel"
         @system_arguments[:preload] = true if @src.present? && preload?
-
+        @system_arguments[:classes] = class_names(
```
jamieshark, line 391: not sure if this is necessary since we're not actually using this class for styles, but thought it felt like good best practice, but happy to take it out if it is more misleading.

app/components/primer/alpha/select_panel.html.erb
```diff
@@ -32,20 +41,12 @@
                   position: :absolute
                 }
               )) %>
+              <%= render(Primer::Alpha::ActionList::Divider.new(mb: 0)) %>
```
camertron, line 44: Hmm the `Divider` component isn't really designed to be used like this. Can we write a bit of CSS instead?

app/components/primer/alpha/select_panel.pcss
```diff
@@ -0,0 +1,3 @@
+.SelectPanel-loadingPanel {
```
camertron, line 1:
```suggestion
.SelectPanel-loadingSpinner {
```

camertron (changes requested):
Nice! I'll let @langermank give the final go-ahead tho 😄

app/components/primer/alpha/select_panel.pcss
langermank: I think this will still look odd if there aren't many results, but I guess it's still an improvement? Maybe we could add a preview to show how that would look (so setting the size to `medium_portrait` for example, and then only return one or two results

app/components/primer/alpha/select_panel.html.erb
langermank: Doesn't the header have a divider param? The `Dialog` has `show_divider`

camertron: Oh dang it does! https://github.com/primer/view_components/blob/6348ede37a08d9b325aa24bd30156bd9a8377a08/app/components/primer/alpha/dialog/header.rb#L56

app/components/primer/alpha/select_panel.pcss
jamieshark: @camertron - I don't know if `loadingSpinner` is totally accurate because this style also applies to the error message when the spinner is hidden. Do you still want the class name change?

camertron: Hmm... alright let's keep `loadingPanel` then 👍

jamieshark:
@langermank - I've added a `limit_items` param for the playground where you can preview the various sizes with a specified number of items. Here's an example of `:medium_portrait` size with `2` items showing after load.

http://primer-view-components-preview-3044.eastus.azurecontainer.io/view-components/lookbook/inspect/primer/alpha/select_panel/playground?size=medium_portrait&limit_items=2

jamieshark:
@maximedegreve - the error message shown in the playground is only the default shown by the element. The consumer can provide custom `error_content` if they like, so I don't think a separate slot is needed, since the consumer can just provide that directly as the `error_content` slot. I would recommend creating another issue for the loading-state slot, so we are not preventing the other 3 issues from being closed as part of this work.

With regards to the other changes, I've added a little padding between the icon and text (the spec specifies `14px` but in the interest of keeping with the system I've given `mt-3` which compiles to `16px) and updated the font size of the filter banner.

maximedegreve:
Noticed a few small things. Some might be able to be addressed in this issue others might need opening of new issues.

1. The margin above the title should only be 8px and not 16px.



2. The example below closes the panel when I initiate a filter query and press enter. Since the focus is the input and not on an item this is unexpected and incorrect. The cursor is also in the wrong spot after opening it again. I can't remember 100% sure anymore but I think it should wipe the filter value when closing the panel.

https://github.com/user-attachments/assets/35aa8dd1-4ff3-4b7f-9027-ab4c5599da3a

3. Multi select should have visual checkboxes (not semantically) to visually indicate to users it won't close the panel when making a multiple selection. Single select is correct to use the checkmark but multi select isn't. See the [screenshots in our docs](https://primer.style/components/selectpanel).

4. The empty state is also missing a min-height similar to the error/loading states

Screenshot 2024-09-09 at 10 29 25

5. The playground example changes in width after results are loaded in. This causes a layout swift we want to avoid.

https://github.com/user-attachments/assets/64c0c83a-0f77-4481-9bb9-b225e13f651a

jamieshark:
@maximedegreve 

1. I have addressed this with my recent changes, though I would like to note that the spec specified `14px` and I'm wondering what can be done to mitigate this confusion in the future.
2. I have created another issue for this: https://github.com/github/primer/issues/3932
3. I have created another issue for this: https://github.com/github/primer/issues/3933
4. I have addressed this in my changes.
5. I believe this is happening because the overlay size is set to `:auto`. If you would like to avoid width changes, it might help to select a different size. Otherwise, there's not really any way for us to know the exact width of the content that's coming in.

maximedegreve (approved):
Approving since the other reported issues will be saved as separate issues.

demo/app/controllers/select_panel_items_controller.rb
```diff
@@ -35,9 +35,10 @@ def index
     query = (params[:q] || "").downcase
 
     results = if show_results
-                SELECT_PANEL_ITEMS.select do |item|
+                res = SELECT_PANEL_ITEMS.select do |item|
                   [item[:title], item[:description]].join(" ").downcase.include?(query)
                 end
+                params.fetch(:limit_items, -1).to_i >= 0 ? res.take(params[:limit_items].to_i) : res
```
kendallgassner, line 41: Can we throw this in a reivew labs -- I want to make sure there are no issues with single select, limits, and searching

kendallgassner:
✅ For single-select, I tested that if a page limit is added the selected state is updated correctly as Items are loaded/unloaded. 


A video showing that the selected item state is retained as I filter using the SelectPanel's input.

kendallgassner:
🔴 I did find one niche bug. If a selectedItem does not initial show due to a page limit... It is never selected. See video:



 
A video where I change the page limit to 1. The selected item is not loaded. I then filter by the selectItems name showing it is not selected.

kendallgassner (approved):
💟 ✅

camertron (approved):
Nice!

PR #3052, opened 2024-09-04
Remove OpenStruct from primer_octicon cop

HParker:
OpenStruct will not be included in modern Ruby versions, so we should not rely on it being there. This removes usage and instead builds the same class instance that is returned above.

```
irb(main):003> h = RuboCop::AST::HashNode.new(hash) => s(:hash)
irb(main):004> h.keys
=> []
irb(main):005> h.pairs
=> []
irb(main):006> h.type
=> :hash
```

### DIFF
--- a/test/lib/rubocop/primer_octicon_test.rb
+++ b/test/lib/rubocop/primer_octicon_test.rb
@@ -273,4 +273,12 @@ class RubocopPrimerOcticonTest < CopTestCase
 
     assert_correction "primer_octicon(:icon, size: :medium)"
   end
+
+  def test_no_args
+    investigate(cop, <<-RUBY)
+      octicon
+    RUBY
+
+    assert_correction "primer_octicon"
+  end
 end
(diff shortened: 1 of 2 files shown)


test/lib/rubocop/primer_octicon_test.rb
```diff
@@ -273,4 +273,12 @@ def test_medium_size
...
     assert_correction "primer_octicon(:icon, size: :medium)"
   end
+
+  def test_no_args
+    investigate(cop, <<-RUBY)
+      octicon
+    RUBY
+
+    assert_correction "primer_octicon"
+  end
```
camertron, line 283: It looks like the cop explicitly does nothing if no arguments are passed to the `octicon` method, so that's why this test is failing 😄

camertron (changes requested):

camertron:
Hey @HParker, several of my branches are failing CI with error messages related to openstruct, so I went ahead and merged your fix in https://github.com/primer/view_components/pull/3055 👍

PR #3057, opened 2024-09-05
Don't select the item if it's an href

owenniblock:

Don't display a checkmark when anchor items are activated

We could alternatively use the `tag` and check for `:a` but I figured this approach would work and was less prone to causing problems if folks are using tags in a weird way but still want stuff to be selectable.

### Integration

No

Closes https://github.com/github/primer/issues/3886

#### Risk Assessment
  

Because this only affects SelectPanel directly, we should know if we have any SelectPanels with links for items which don't actually redirect to a new page.

### Accessibility

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


owenniblock:
Questions:

* Do we want to do the same for multi-select?
* Should we add a selected item to the example so it's obvious that this is supported?

camertron (approved):
Awesome, thanks!

camertron:
> Do we want to do the same for multi-select?

Hmm, I don't think so? I can see a scenario where you might want to show two selections, but nobody's asking for that right now, so let's not worry about it.

> Should we add a selected item to the example so it's obvious that this is supported?

Like to the "list of links" example? Yes, that would be nice to have 👍

PR #3095, opened 2024-09-17
Clear input when SelectPanel closed

owenniblock:

Clearing the input when the SelectPanel dialog closes.

### Integration
No

Relates to https://github.com/github/primer/issues/3932

#### Risk Assessment
  

Easiest way to fix this was to clear the input in TypeScript when the `hide` method is called.

### Accessibility

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [x] Tested in Edge


camertron (changes requested):
Hey @owenniblock, looking good!

I'm concerned that clearing out the filter input by doing `this.filterInputTextField.value = ''` won't trigger a remote request in remote fetch mode. Could you add a test for that case and ensure it works as expected?

Expected behavior:

1. On close, filter input text is cleared.
2. Panel closes and is hidden from view.
3. Since the filter text is now empty, the panel kicks off a remote fetch from the server and updates the list of items.
4. User re-opens the panel and sees the original list of items.

owenniblock:
@camertron quick question on the exact functioning you're after here.

I can see that currently it doesn't do the fetch until the dialog is re-opened (it does do the fetch so it could be worse).

Do we want to trigger the fetch in the background when the dialog is hidden? I figure the positive to doing it this way is that the data is loaded ready when the user re-opens the dialog. Alternatively, if we leave it how it is, the positive is that if the user doesn't re-open the dialog (perhaps they opened it in error) we don't do an unnecessary call to re-populate the data.

Happy either way, just wanted to confirm before making a change from what's already there 😅

camertron:
@owenniblock yeah, that's a great question. I personally think it's a bit jarring to see a list of items when you open the panel only to have those items more or less immediately replaced with another set of items. You'd be a second or so into visually scanning the items and then whoops! All replaced 🥴 

Thoughts?

owenniblock:
@camertron yeah I think that makes sense. It's also a bit annoying that the loading is gone (there's a small loading spinner in the input itself). I'll see what I can do 😄

camertron (approved):
Cool 😎

PR #3144, opened 2024-10-09
reduce cost of style calc+layout in --dialog-scrollgutter

keithamus:

In our mob pairing session called "Making Dotcom Fast" we ran into an issue where the Primer Dialog component is causing an expensive recalc operation on every page load; this is due to it setting the `--dialog-scrollgutter` css property, which is a property derived from the `innerWidth` & `body.clientWidth` properties.

In order to prevent this cost on page load, we decided to move the operation to the first `click` event that can be handled by the `dialogInvokerButtonHandler`. This moves the cost of the recalc away from document load to document click:

image

image

However this change alone still leaves the recalc cost during the first click event which is still costly and could impact INP. 

image

So we added some CSS - inspired by [this article on the `@property` rule and how it can lower recalc costs by disabling inheritance](https://web.dev/blog/at-property-performance) - which is perfect for this use case. This reduced recalc cost from ~70-90ms on my machine to under 2ms:

image

We've _not_ moved the code away from the hot-path, and into the click event listener, due to compat risk with the Dialog element and its use in dotcom. This can come at a later date, perhaps.

### Screenshots

(See above perf traces)

#### Risk Assessment
  

### Accessibility

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

/cc the commit co-authors:
@ansballard, @AdamShwert, @arelia, @arielvalentin, @chrisgavin, @Cbodfield, @composerinteralia, @garman, @dgreif, @erinnachen, @georgebrock, @jibrang, @jfuchs, @khiga8, @manuelpuyol, @mattcosta7, @another-mattr, @MelissaPastore, @theinterned, @phillmv, @Raffo, @cheshire137, @srt32

### DIFF
--- a/app/components/primer/alpha/dialog.pcss
+++ b/app/components/primer/alpha/dialog.pcss
@@ -8,8 +8,22 @@
 
 /* Overlay */
 
+/* The --dialog-scrollgutter property is used only on the body element to
+ * simulate scrollbar-gutter:stable. This property is not and should not
+ * be used elsewhere in the DOM. There is a performance penalty to
+ * setting inherited properties which can cause a large style recalc to
+ * occur, so it benefits us to prevent inheritance for this property.
+ * See https://web.dev/blog/at-property-performance
+ */
+@property --dialog-scrollgutter {
+  initial-value: 0;
+  inherits: false;
+  syntax: "<length>";
+}
+
 /* TODO: One day this can be :has(:modal), when it is better supported */
 body.has-modal {
+  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
   padding-right: var(--dialog-scrollgutter) !important;
   overflow: hidden !important;
 }
@@ -350,6 +364,7 @@ dialog.Overlay:not([open]) {
   &:hover,
   &:focus {
     background-color: var(--button-default-bgColor-hover);
+    /* stylelint-disable-next-line primer/colors */
     border: var(--borderWidth-thin) solid var(--control-bgColor-hover);
   }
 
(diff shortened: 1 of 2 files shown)


app/components/primer/dialog_helper.ts
dgreif, line 1: Are there ever cases where a dialog can be open on page load? Or is a click _absolutely required_ to get a dialog open?

langermank (approved):
This is cool! 😄

app/components/primer/dialog_helper.ts
cheshire137: > Are there ever cases where a dialog can be open on page load?

I know we do some of these internally, but the approach there is to use a [Catalyst component](https://github.com/github/catalyst) that, when connected, either calls `showModal` on an `HTMLDialogElement` or sets the `open` property for other elements. So I think that would still work with this change.

app/components/primer/alpha/dialog.pcss
```diff
@@ -8,8 +8,15 @@
 
 /* Overlay */
 
+@property --dialog-scrollgutter {
```
mattcosta7, line 11: nit: can we comment here, maybe with a link to this issue?  future us might wonder why we're doing it

app/components/primer/dialog_helper.ts
keithamus: This does make it more challenging. Perhaps we can just ship the CSS fix for now. Or we can dispatch our own toggle events which that Catalyst component can emulate.

keithamus: Given the compat risk involved here with this catalyst component, let's drop the JS for now and think about how to resolve the JS issue at a later time (maybe when events are added).

PR #3145, opened 2024-10-10
Add option to Dialog to permit/disable page scrolling

camertron:

The `SelectPanel` component uses parts of the `Dialog` component under the hood, which comes with custom CSS that prevents the page from scrolling when the dialog is opened as a modal. This means that `SelectPanel`s, despite behaving more like popovers, prevent page scrolling too. We'd like to enable scrolling for `SelectPanel`s but keep the existing behavior for other dialogs.

### Integration

Closes https://github.com/github/primer/issues/4126

#### Risk Assessment
  

I refactored our CSS rules and added a parameter to `Dialog` that conditionally adds the `.Overlay--disableScroll` CSS class to the `<dialog>` element. The parameter defaults to `true` to preserve existing behavior.

Since we can use it now, I also switched to using `body:has(...)` instead of the hackish `body.has-modal` selector for applying the scroll gutter and scroll disabling rules.

I really tried to write a test for this, but wasn't able to figure out how to simulate scrolling appropriately. If scrolling is disabled in CSS via `overflow: hidden`, JavaScript's `scroll`, `scrollTo`, and `scrollBy` functions scroll the page anyway. Firing manually-created `wheel` and `touch` events also didn't work - the page never scrolls, even when `overflow` is not set to `hidden`. If y'all have any ideas here, I'm all ears.

### Accessibility

### Merge checklist

~- [ ] Added/updated tests~
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


jonrohan (approved):
`:has` is technically supported now, think you should try and test this on a browser that doesn't support :has just to see what a fail case might look like

camertron:
@jonrohan just verified that the failure case simply allows the page to scroll, which I don't think is a blocker. There's also a scroll gutter property that won't apply, but that also doesn't seem to affect functionality.

PR #3154, opened 2024-10-15
Upgrade stylelint to latest version

jonrohan:

Upgrades to the latest version of the stylelint-config https://github.com/primer/stylelint-config/releases/tag/v13.1.1

There is a bug in the browser-compat plugin that fails https://github.com/3846masa/stylelint-browser-compat/pull/807

To get around the error I turned off checking for the `scrollbar-width` property


langermank (approved):
Supa fast

PR #3159, opened 2024-10-18
Increase the default space between form elements

camertron:
This is a copy of https://github.com/primer/view_components/pull/3087 that I'm submitting under the Primer org so all our CI checks will run correctly.


camertron:
CI is passing and the original PR is approved. Merging using admin privileges.

PR #3204, opened 2024-11-21
 fix(popover): update caret styles to offset ::after element

joshblack:
Context: https://github.com/github/primer/issues/4414

Update the Popover styles so that the caret is styled correctly

#### Risk Assessment
  


francinelucca (approved):
Looks great 🎉

langermank (approved):

Issue #3259, opened 2025-01-10
Memory leak and slowdowns in `Primer::Forms` components


cbliard:
In `Primer::Forms::BaseComponent`, each time `#compile!` is called, it calls [this line](https://github.com/primer/view_components/blob/9f87ab122fbe7a85552b3124c4efbcec060e9af4/app/lib/primer/forms/base_component.rb#L19):

```
renders_template File.join(dir, "#{self.name.demodulize.underscore}.html.erb"), :render_template
```

This adds the template file to the `template_globs` array of the `Primer::Forms::ActAsComponent` module ([source](https://github.com/primer/view_components/blob/9f87ab122fbe7a85552b3124c4efbcec060e9af4/app/lib/primer/forms/acts_as_component.rb#L53-L55)), which is then used in `compile!` in the included `ActAsComponent` module to compile the templates before rendering ([source](https://github.com/primer/view_components/blob/9f87ab122fbe7a85552b3124c4efbcec060e9af4/app/lib/primer/forms/acts_as_component.rb#L62-L64)).

The problem is that `compile!` is called each time `Primer::Forms::BaseComponent#compile_and_render_template` is called ([source](https://github.com/primer/view_components/blob/9f87ab122fbe7a85552b3124c4efbcec060e9af4/app/lib/primer/forms/base_component.rb#L63)) which happens every time a component is rendered ([source](https://github.com/primer/view_components/blob/9f87ab122fbe7a85552b3124c4efbcec060e9af4/app/lib/primer/forms/base_component.rb#L30)).

I confirmed it by adding a `p template_globs` in `#renders_templates`. The printed list keeps growing on each rendering.

It's not a big memory leak, but the problem becomes noticeable in development: because templates are always recompiled in dev, and the template list is ever growing, the same template gets compiled multiple times which eventually slows down the application.

I noticed it as I was waiting for more than 1 second for a simple page update. This was confirmed in the application logs:

```
D, [2025-01-10T14:48:30.157985 #37009] DEBUG -- :   ↳ app/forms/work_packages/dialogs/create_form.rb:63:in `block (2 levels) in <class:CreateForm>'
I, [2025-01-10T14:48:31.238213 #37009]  INFO -- : Completed 200 OK in 1131ms (Views: 0.0ms | ActiveRecord: 10.7ms | Allocations: 3158630)
```

After a restart, on first request:

```
[...]

PR #3266, opened 2025-01-13
Fix memory leak in forms framework

camertron:

This PR fixes an issue introduced in https://github.com/primer/view_components/pull/3236 that causes templates to be compiled more than once, i.e. every time a component is rendered. While definitely a problem, there should be no negative effects aside from extra computation time and memory usage. The problem should only happen in development, i.e. not production.

### Integration

Fixes https://github.com/primer/view_components/issues/3259

#### Risk Assessment
  

Based on the excellent description in https://github.com/primer/view_components/issues/3259, I moved the code that registers the component's template from the `compile!` method to to a `self.inherited` callback. This should result in only compiling the template a single time when the component class is first loaded.

### Accessibility

### DIFF
--- a/app/lib/primer/forms/acts_as_component.rb
+++ b/app/lib/primer/forms/acts_as_component.rb
@@ -66,30 +66,45 @@ module Primer
         @compiled = true
       end
 
-      private
-
       def template_root_path
         return @template_root_path if defined?(@template_root_path)
 
         form_path = Utils.const_source_location(self.name)
-        @template_root_path = if form_path
-          File.join(File.dirname(form_path), self.name.demodulize.underscore)
+        @template_root_path =
+          if form_path
+            File.join(File.dirname(form_path), self.name.demodulize.underscore)
+          end
+      end
+
+      def base_template_path
+        return @base_template_path if defined?(@base_path)
+        base_path = Utils.const_source_location(self.name)
+
+        if base_path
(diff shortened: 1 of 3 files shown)


jonrohan (approved):

app/lib/primer/forms/acts_as_component.rb
```diff
@@ -66,30 +66,45 @@ def compile!
...
-        @template_root_path = if form_path
-          File.join(File.dirname(form_path), self.name.demodulize.underscore)
+        @template_root_path =
+          if form_path
+            File.join(File.dirname(form_path), self.name.demodulize.underscore)
+          end
+      end
+
+      def base_template_path
+        return @base_template_path if defined?(@base_path)
```
cbliard, line 80: Shouldn't it be `if defined?(@base_template_path)` instead of `if defined?(@base_path)`?
`@base_path` is never defined.

cbliard (commented):
Thanks for the quick fix and the explanations @camertron.

It works well now, and I did not observe any slowdowns or regressions so you fixed the issue 👍 

Looking at the code, I just noticed one minor issue with a instance variable caching not working.

app/lib/primer/forms/acts_as_component.rb
camertron: Woops! Good catch, thanks 👍

PR #3271, opened 2025-01-16
Introduce ARIA snapshot tests for all components

khiga8:
### What are you trying to accomplish?

This PR:
* upgrades to latest Playwright version.
* introduces [ARIA snapshot tests](https://playwright.dev/docs/aria-snapshots) for all components, alongside existing visual regression tests. These ARIA snapshot tests store the role and accessible name exposed to assistive technologies.

**Why is this helpful?** 
- Right now, there's no great way to test accessible markup in PVC. 
- Capybara, the main testing library used in PVC, doesn't natively support any accessible name/ARIA assertions making it difficult to test the semantics of the markup. 
- Consequently, it's very easy to push a change that unintentionally ends up changing the underlying accessible markup, and breaks the experience for assistive technology users.
  - Here's [an example bug](https://github.com/primer/view_components/pull/3270) (which.. I introduced 🤦‍♀️ ) where there's a regression in the accessible name. 
- By having ARIA snapshot tests in place, it should be much easier to detect changes in the accessibility markup, and become easier to catch regressions.
  - Here's the [corresponding snapshot](https://github.com/primer/view_components/blob/kh-test-aria-snapshots/.playwright/screenshots/snapshots.test.ts-snapshots/primer/beta/button/trailing_counter/default.yml#L1) of the accessible name which CLEARLY shows the missing counter text.

**Consequences**
- Additional build time

#### List the issues that this change affects.

Closes # (type the GitHub issue number after #)

#### Risk Assessment
  

- [x] **Low risk** the change is small, highly observable, and easily rolled back.
- [ ] **Medium risk** changes that are isolated, reduced in scope or could impact few users. The change will not impact library availability.
- [ ] **High risk** changes are those that could impact customers and SLOs, low or no test coverage, low observability, or slow to rollback.

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Lookbook)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3275, opened 2025-01-16
Experiment: Bump down playwright

khiga8:
In newly open PRs, 900+ snapshots get re-generated (example: https://github.com/primer/view_components/pull/3273). I don't know why this didn't happen before.

I want to see if the snapshots get regenerated in the previous playwright version. If it doesn't, we can attribute this to something in the new playwright version. If that's the case, what's a mystery is why this didn't happen in https://github.com/primer/view_components/pull/3271/...


khiga8:
These snapshots are getting re-generated even when bumping back down, so looks like it might not be playwright related?

PR #3314, opened 2025-02-03
[Primer::Alpha::FormControl] Remove required example for SegmentedControl

khiga8:
Relates to: https://github.com/github/accessibility-audits/issues/10021

A SegmentedControl cannot have a required state so this example doesn't make sense:
* The `<ul>` role, `list`, doesn't support a required state. We could try to set `aria-required="true"` but that's considered invalid HTML.
* SegmentedControl will never have an empty state so marking as required doesnt make sense.

⚠️ **I'm confused on how this component is meant to be used and wondering if it should be deprecated or have clearer guidance? Advice needed!**  

Currently:
* This component is not shown on the Docs page for [FormControl](https://primer.style/components/form-control). The overview for FormControl claims that `required` will be forwarded to the input, but that's not what happens here. In fact, there's an [explicit note in the code](https://github.com/primer/view_components/blob/main/app/components/primer/alpha/form_control.rb#L16) saying this component only visually renders an asterisk.
* Consumers are able to pass in any arbitrary component as a child to give this child a label, caption, required asterisk.
   * BUT even though labels and required asterisks are visually rendered, there's actually no programmatic association which is problematic accessibility wise since consumers might set `label:` expecting the component to handle things appropriately (when it actually doesn't).
  * We could try to have the component handle setting `id`/`for` or `aria-labelledby`, but it depends on what's being passed in, and there doesn't seem to be any restrictions.

#### Risk Assessment
  

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/previews/primer/alpha/form_control_preview.rb
+++ b/previews/primer/alpha/form_control_preview.rb
@@ -74,19 +74,6 @@ module Primer
         )
       end
 
-      # @label Required
-      def required
-        render_with_template(
-          template: "primer/alpha/form_control_preview/playground",
-          locals: {
-            system_arguments: {
-              label: "Best character",
-              required: true
-            }
-          }
-        )
-      end
-
       # @label With visually hidden label
       def with_visually_hidden_label
         render_with_template(
--- a/test/components/alpha/form_control_test.rb
+++ b/test/components/alpha/form_control_test.rb
@@ -29,14 +29,6 @@ module Primer
         assert_includes(described_by_ids, validation_id)
       end
 
-      def test_required
-        render_preview(:required)
-
-        assert_selector(".FormControl-label", text: "Best character") do
-          assert_selector("[aria-hidden=true]", text: "*")
-        end
-      end
-
       def test_visually_hidden_label
         render_preview(:with_visually_hidden_label)
 


camertron (changes requested):
Hey @khiga8, thank you for this! The purpose of these previews is to showcase what `FormControl` can do, so I think we still want the "required" example. What would you think of swapping out the `SegmentedControl` for a `Select`, or another component that can be required?

khiga8:
> Hey @khiga8, thank you for this! The purpose of these previews is to showcase what `FormControl` can do, so I think we still want the "required" example. What would you think of swapping out the `SegmentedControl` for a `Select`, or another component that can have an empty state?

Hey @camertron! I'd love clarification on the intended usecase for `FormControl`. 

Initially, I did explore swapping this segmented control with [Primer::Alpha::TextInput](https://primer.style/components/text-input/rails/alpha). However, that component already supports a label/caption, so using it in conjunction with this component will result in a double label situation. 

Same deal for [Primer::Alpha::Select](https://primer.style/components/select/rails/alpha) which renders a label by default.

So it appears that the existing Primer form control components are incompatible and should never be used with `FormControl` .

Next, I looked into just wrapping a plain old HTML element, which I had to render like the following in order to be able to pass in `input_arguments` generated by this component.

```rb
<%= render(Primer::Alpha::FormControl.new(label: "Write something", required: true)) do |component| %>
  <% component.with_input do |input_arguments| %>
    <%= render(Primer::BaseComponent.new(tag: :textarea, **input_arguments)) %>
  <% end %>
<% end %>
```

However, this resulted in a label/asterisk being _visually_ rendered but without any proper programmatic associations, which is an accessibility issue:

```html
<div data-view-component="true" class="FormControl">
  <label data-view-component="true" class="FormControl-label">
    Write something
    <span aria-hidden="true">*</span>
  </label>        
  <textarea data-view-component="true">
</textarea>
</div>
```

But if we don't expect consumers to use this form control component with form elements anyways (since they should be using the relevant form component), I wasn't sure that it made sense to invest in having this component render these attributes. 

I think it'd help to understand what the component _should_ and _shouldn't_ provide, so we can document it!

apologies for this block of text... I hope my question and points of confusion are clear! Let me know if you'd like to hop on a call :)  🙇

TylerJDev:
Thanks so much for creating this @khiga8! 

I agree, I don't believe `SegmentedControl` should act as a form control, and we should probably remove the example that we currently have in Lookbook.  

Looking at Primer query, it seems like [the usage of it is limited](https://primer-query.githubapp.com/?Constant=%22Primer%3A%3AAlpha%3A%3AFormControl%22&onstant=%22Primer%3A%3AAlpha%3A%3AFormControl%22&constant=%22Primer%3A%3AAlpha%3A%3AFormControl%22). I see it's mainly being used in conjunction with menus, or SelectPanel. I'm guessing these might be problematic implementations as the labels won't be programmatically connected with the component. A very similar issue came up in [another audit issue for `ActionMenu`](https://github.com/github/accessibility-audits/issues/10133).

I'm wondering if it's worth trying to associate the labels with the component's via `aria-labelledby` or not? As for the `aria-required`, I think I'd lean more on the side of deprecation, as there doesn't seem to be an actual usage of it in Dotcom according to Primer query.

camertron:
@khiga8 and I met on Zoom to chat about this yesterday and we were able to figure out a path forward. Some notes from that call:

1. `FormControl` is really designed to be used internally by the Primer forms framework. It's exposed to external consumers only to enable advanced use-cases, or cases the framework doesn't currently support.
2. As @khiga8 mentioned, the `TextField` component already features a label, caption, and validation message, so can't be used in the `FormControl` previews. I needed something to wrap that didn't have these labels, so I chose `SegmentedControl`.
3. It's clear we're missing accessibility documentation for `FormControl`, specifically around how to connect the label and input together.
4. There's a bug causing `aria-required` to not get attached to the input.
5. There's a bug preventing users from attaching a `for` attribute to the label.

On the call, @khiga8 and I worked up a PR that I polished up and submitted this morning: https://github.com/primer/view_components/pull/3316 It should address these issues, including the previews, which now use a regular 'ol `<input type="text">` rather than a `SegmentedControl`.

camertron:
Closing in favor of https://github.com/primer/view_components/pull/3316

PR #3319, opened 2025-02-07
update ToggleSwitch to support autofocus

ktravers:

_Reviewers: By approving this Pull Request you are approving the code change, as well as its deployment and mitigation plans. Please read this description carefully. If you feel there is anything unclear or missing, please ask for updates._

Related to https://github.com/github/accessibility-audits/issues/10006
Required by https://github.com/github/github/pull/361376

This PR updates the ToggleSwitch component to allow the `autofocus` property to be applied to its button element. Doing so will allow this component to be fully accessible when used in conjunction with the [Rails turbo framework](https://github.com/hotwired/turbo-rails). See https://github.com/github/accessibility-audits/issues/10006 for more context.

### Screenshots

https://github.com/user-attachments/assets/bc33f286-983f-4180-b086-3fc6537fb099

### Integration

No prod changes required.

* Helps with https://github.com/github/accessibility-audits/issues/10006

#### Risk Assessment
  

I explored two approaches for allowing users to optionally apply the `autofocus` attribute to the ToggleSwitch button element:
1. [Add generic `button_arguments` argument](https://github.com/primer/view_components/pull/3319/commits/bb0c6cdd245561575e37a59aa9d96fefdb1bebfc), allowing users to pass any supported attribute to the button
2. [Add specific `autofocus` argument](https://github.com/primer/view_components/pull/3319/commits/bb05db7199a348fa6397e8ad755ae55ec254cc8c), so users can apply this single attribute directly

I opted to use the second approach, as it is the minimum viable change required to make this component accessible when used with [turbo-rails](https://github.com/hotwired/turbo-rails). Using a more surgical, targeted approach seemed safer and easier to maintain/test than the more permissive, open-ended first approach. This also seemed more in line with Primer view component design principles.

* What do reviewers think of the two approaches described above? Would folks prefer to ship the more permissive `button_arguments` approach? 🤔  I didn't see this kind of argument in many existing components; however, it does resemble generic system arguments, which are allowed for all components (if not the child elements they render, necessarily).
* Are there any concerns around supporting `autofocus` on a toggle switch button? It is an allowed attribute for buttons in general, but I wasn't sure if there was some reason we didn't already support it here.
* Should this be a minor release instead of a patch?

### Accessibility

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/previews/primer/alpha/toggle_switch_preview.rb
+++ b/previews/primer/alpha/toggle_switch_preview.rb
@@ -62,6 +62,10 @@ module Primer
       def with_turbo
         render(Primer::Alpha::ToggleSwitch.new(src: UrlHelpers.toggle_switch_index_path, turbo: true))
       end
+
+      def with_autofocus
+        render(Primer::Alpha::ToggleSwitch.new(src: UrlHelpers.toggle_switch_index_path, autofocus: true))
+      end
     end
   end
 end
--- a/previews/primer/alpha/toggle_switch_preview.rb
+++ b/previews/primer/alpha/toggle_switch_preview.rb
@@ -62,6 +62,11 @@ module Primer
       def with_turbo
         render(Primer::Alpha::ToggleSwitch.new(src: UrlHelpers.toggle_switch_index_path, turbo: true))
       end
+
+      def with_button_arguments
+        button_arguments = { id: 'foo', data: { turbo_permanent: true } }
+        render(Primer::Alpha::ToggleSwitch.new(src: UrlHelpers.toggle_switch_index_path, button_arguments: button_arguments))
+      end
     end
   end
 end
--- a/previews/primer/alpha/toggle_switch_preview.rb
+++ b/previews/primer/alpha/toggle_switch_preview.rb
@@ -62,11 +62,6 @@ module Primer
       def with_turbo
         render(Primer::Alpha::ToggleSwitch.new(src: UrlHelpers.toggle_switch_index_path, turbo: true))
       end
-
-      def with_button_arguments
-        button_arguments = { id: 'foo', data: { turbo_permanent: true } }
-        render(Primer::Alpha::ToggleSwitch.new(src: UrlHelpers.toggle_switch_index_path, button_arguments: button_arguments))
-      end
     end
   end
 end
--- a/.changeset/late-jobs-think.md
+++ b/.changeset/late-jobs-think.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Adds optional autofocus to ToggleSwitch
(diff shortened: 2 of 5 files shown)


previews/primer/alpha/toggle_switch_preview.rb
```diff
@@ -62,6 +62,10 @@ def with_bad_csrf_token
       def with_turbo
         render(Primer::Alpha::ToggleSwitch.new(src: UrlHelpers.toggle_switch_index_path, turbo: true))
       end
+
```
ktravers, line 65: Should this be snapshotted? 🤔

camertron: I would say yes, but only if it's visually different from other previews.

camertron (approved):
Nice, thanks! I actually think we should implement _both_ of the approaches you outlined 😄 IMO, `autofocus:` should be an argument to `ToggleSwitch.new`, but so should `button_arguments:`. We actually do this in a number of places when we want to accept arguments for child elements.

Inside the constructor, we can do something like:

```ruby
@button_arguments = button_arguments
@button_arguments[:autofocus] = true if autofocus
@button_arguments[:aria] = merge_aria(
  @button_arguments,
  { aria: { pressed: on? } }
)
```

previews/primer/alpha/toggle_switch_preview.rb
ktravers: The only visual difference would be whatever focus styles are in place. I've got aggressive settings on my local machine when it comes to focus, so it's always visually distinct, so I'm not sure if that's _actually_ true for this component as well or just me 😅  Lemme take a quick look at the CSS and/or disable my focus stuff to confirm.

ktravers:
> I actually think we should implement _both_ of the approaches you outlined 😄 IMO, `autofocus:` should be an argument to `ToggleSwitch.new`, but so should `button_arguments`

@camertron Cool, that sounds good to me! Would it be ok to ship this PR as is to unblock https://github.com/github/github/pull/361376, then implement `button_arguments` in a separate fast follow PR?

previews/primer/alpha/toggle_switch_preview.rb
ktravers: Ok, AFAICT ToggleSwitch has [focus styling](https://github.com/primer/view_components/blob/59933d4eb741d307657d87c042f2ebc4671ffc7e/app/components/primer/alpha/toggle_switch.pcss#L45-L48), but it's only visible when `:focus-visible` is applied via the browser. So I don't think we need a separate preview 👍

PR #3320, opened 2025-02-07
Formatting tweak in NavList methods docs

mperrotti:

I was getting an error because the primer-docs Markdown parser was interpreting `'<ul>'` as HTML.


camertron (approved):
Thanks!

PR #3340, opened 2025-02-20
[SelectPanel]: Ensure formbuilder input/defaultvalue is available without opening the selectpanel

kendallgassner:

#### Risk Assessment
  

### Accessibility

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


app/components/primer/alpha/select_panel.html.erb
```diff
@@ -1,4 +1,9 @@
 <%= render Primer::BaseComponent.new(**@system_arguments) do %>
+  <% if acts_as_form_input? %>
+    <span data-select-panel-inputs="true"> 
+      <%= @form_builder.hidden_field(@input_name, multiple: multi_select?, skip_default_ids: true, value: @default_value) %>
+    </span> 
+  <% end %>
```
kendallgassner, line 6: Stop using form_builder input in ActionList. This is important since the ActionList input is only available once items are loaded.

Previously, we could only use form_builder with `fetch_strategy: :local` since the ActionList is loaded on page load using this strategy. This is problematic since most of our usecases are using `remote` or `eventually_local` fetch strategies.

app/components/primer/alpha/select_panel.rb
```diff
@@ -153,26 +153,6 @@ module Alpha
     # end
     # ```
     #
-    # If items are provided dynamically, things become a bit more complicated. The `form_for` or `form_with` method call
-    # happens in the view that renders the `SelectPanel`, which means the form builder object but isn't available in the
-    # view that renders the list items. In such a case, it can be useful to create an instance of the form builder maually:
-    #
-    # ```erb
```
kendallgassner, line 160: Remove inaccurate logic.

app/components/primer/alpha/select_panel.rb
```diff
@@ -375,7 +355,7 @@ def with_avatar_item(**system_arguments)
       # @param dynamic_aria_label_prefix [String] If provided, the prefix is prepended to the dynamic label and set as the value of the `aria-label` attribute on the show button.
       # @param body_id [String] The unique ID of the panel body. If not provided, the body ID will be set to the panel ID with a "-body" suffix.
       # @param list_arguments [Hash] Arguments to pass to the underlying <%= link_to_component(Primer::Alpha::ActionList) %> component. Only has an effect for the local fetch strategy.
-      # @param form_arguments [Hash] Form arguments to pass to the underlying <%= link_to_component(Primer::Alpha::ActionList) %> component. Only has an effect for the local fetch strategy.
```
kendallgassner, line 378: Not anymore ✨ !

app/components/primer/alpha/select_panel.rb
```diff
@@ -429,6 +409,11 @@ def initialize(
         @dynamic_aria_label_prefix = dynamic_aria_label_prefix
         @loading_label = loading_label
         @loading_description_id = nil
+
+        @form_builder = form_arguments[:builder]
+        @default_value = form_arguments[:default_value]  
+        @input_name = form_arguments[:name]
```
kendallgassner, line 415: Extract all form_arguments and add an optional argument `default_value` so that remote and eventually_local strategies can pass in a pre-set value.

app/components/primer/alpha/select_panel_element.ts
```diff
@@ -984,7 +985,7 @@ export class SelectPanelElement extends HTMLElement {
 
       for (const selectedItem of this.selectedItems) {
         const newInput = document.createElement('input')
-        newInput.setAttribute('data-list-input', 'true')
+        newInput.setAttribute('data-select-panel-input', 'true')
```
kendallgassner, line 988: Change the search term so we don't accidentally look for ActionList's form_builder input.

previews/primer/alpha/select_panel_preview.rb
```diff
@@ -259,6 +259,15 @@ def single_select_form(open_on_load: false, route_format: :html)
         render_with_template(locals: { open_on_load: open_on_load, route_format: route_format })
       end
 
+      # @label Remote fetch form
+      #
+      # @snapshot interactive
+      # @param open_on_load toggle
+      # @param selected_items text
+      def remote_fetch_form(open_on_load: false, selected_items: "Phaser")
```
kendallgassner, line 267: TODO: 
add an exentually_local fetch strategy

app/components/primer/alpha/select_panel.html.erb
khiga8: > Stop using form_builder input in ActionList.

[ActionMenu](https://github.com/primer/view_components/blob/2245a00013e3bf45bc2374a146a67a0a427ccc23/app/components/primer/alpha/action_menu.rb#L246) also accepts `form_arguments` and depends on `ActionList` rendering the hidden form inputs. Would we need to make similar changes in ActionMenu? 

There are [~8 instances](https://primer-query.githubapp.com/?constant=%22Primer%3A%3AAlpha%3A%3AActionMenu%22&attribute=form_arguments) of ActionMenu with `form_arguments`  and [~3 instances](https://primer-query.githubapp.com/?constant=%22Primer%3A%3AAlpha%3A%3ASelectPanel%22&attribute=form_arguments) of SelectPanel with `form_arguments`, so I think we'll need to tread carefully and make sure those existing instances don't break.

khiga8: 1. If we're expanding support for `form_arguments` to non-local strategies, would we also want to account for multi-select and support multiple initial hidden input name/values? ([multi-select form example](https://primer.style/view-components/lookbook/inspect/primer/alpha/select_panel/multiselect_form))

2. Just an observation.. in the [EMU select panel instance](https://github.com/github/github/blob/96d46cd9545d44bace3b6be1a72e86032920a85f/app/views/orgs/teams/_external_group_team.html.erb#L32), the `input_name` that's passed into the component is overridden with a different name when the action list opens. I was confused by that. The name, `input_name` is kind of ambiguous and it's not clear to me if it's supposed to be used as the initial hidden input name similar to the `default_value` you're suggesting here.

app/components/primer/alpha/select_panel.html.erb
```diff
@@ -1,4 +1,16 @@
 <%= render Primer::BaseComponent.new(**@system_arguments) do %>
+  <% if acts_as_form_input? %>
+    <span data-select-panel-inputs="true"> 
+      <% if @default_values.present? && @default_values.length > 0 ? true : false  %>
+        <% for value in @default_values %>
+
+          <%= @form_builder.hidden_field(@input_name, multiple: multi_select?, skip_default_ids: true, value: value) %>
```
kendallgassner, line 7: Based on the current [multi-select form example](https://primer-lookbook.github.com/view-components/lookbook/inspect/primer/alpha/select_panel/multiselect_form) I think it's okay for all of these inputs to have the same name.

app/components/primer/alpha/select_panel_element.ts
```diff
@@ -533,6 +533,7 @@ export class SelectPanelElement extends HTMLElement {
     switch (event.type) {
       case 'include-fragment-replaced': {
         this.#updateItemVisibility()
+        // this.#updateInput()
```
kendallgassner, line 536:
```suggestion
```

app/components/primer/alpha/select_panel.rb
```diff
@@ -429,6 +409,12 @@ def initialize(
         @dynamic_aria_label_prefix = dynamic_aria_label_prefix
         @loading_label = loading_label
         @loading_description_id = nil
+
+        @form_builder = form_arguments[:builder]
+        @default_value = form_arguments[:default_value]  
+        @default_values = form_arguments[:default_values]
```
kendallgassner, line 415: @khiga8 I was using `default_value` for single-select and `default_values` for multi-select but maybe we can just always use `default_values` instead of two props (i.e. pass an array no matter what).... I think it depends on if we keep the same name for all multi-select inputs [(see this comment for more information)](https://github.com/primer/view_components/pull/3340/files#r1968206977)

khiga8: I think that for most scenarios, we can expect the name to remain the same.
 
But in the one EMU example we saw,  the name AND value changes depending on selection, so I fear there are similar edge cases for multi-select that we need to support 🤔 . 

Is there a code search we can do in dotcom to verify?

app/components/primer/alpha/select_panel.html.erb
```diff
@@ -1,4 +1,15 @@
 <%= render Primer::BaseComponent.new(**@system_arguments) do %>
+  <% if acts_as_form_input? %>
+    <span data-select-panel-inputs="true">
```
khiga8, line 3: I wonder if slots would be useful here, where the consumer can render something like:

```html.erb
    <% panel.with_initial_form_inputs(builder: builder) do |c| %>
      <% c.hidden_field(name: "item", value: "item")%>
      <% c.hidden_field(name: "item", value: "item1")%>
    <% end %>
    <% panel.with_show_button { "Sci-fi equipment" } %>
    <% panel.with_footer(show_divider: true) do %>
        I'm a footer!
    <% end %>
```

Pros:
* it'd support both single field or multiple field
* there's flexibility in case we want to allows consumers to set additional attributes on the hidden field
* a bit easier to understand what's getting rendered.

Cons:
* potentially overkill?
* a bit more code to write.
* component is already set up with `form_arguments`

Just a thought!

kendallgassner: The slot would be really cool! 🤔 My only thought is then it differs from how ActionMenu and ActionList operate, and I feel like we should remain consistent. Also, I wonder the likelihood of folks forgetting `skip_default_ids` and `multiple`.

```
<%= @form_builder.hidden_field(@input_name, multiple: multi_select?, skip_default_ids: true, value: value) %>
```

app/components/primer/alpha/select_panel.html.erb
```diff
@@ -1,4 +1,15 @@
...
+      <% if @default_values.present? && @default_values.length > 0 ? true : false %>
```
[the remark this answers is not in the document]
kendallgassner: I would be fine with switching the language to `initial` once we agree on the slot or not.

app/components/primer/alpha/select_panel.html.erb
khiga8: Maybe there's a way to configure a default for them? 
So if a consumer renders:

```.html.erb
   <% panel.with_initial_form_inputs(builder: builder) do |c| %>
      <% c.hidden_field(name: "item", value: "item")%>
      <% c.hidden_field(name: "item", value: "item1")%>
    <% end %>
```

behind the scenes, we'd render:

```.html.erb
<%= @form_builder.hidden_field(name: name, value: value, multiple: multi_select?, skip_default_ids: true, value: value) %>
<%= @form_builder.hidden_field(name: name, value: value, multiple: multi_select?, skip_default_ids: true, value: value) %>
```

khiga8: I agree with you consistency concern. im fine sticking with the args approach too if you're leaning towards it!

khiga8 (commented):
Looking good! Just a few comments :)

khiga8:
Should we open an integration PR in dotcom?

[Releasing PVC docs](https://github.com/github/primer-engineering/blob/fe2a8a192186b66fea570740240d6bd4712a4a72/how-we-work/releasing-primer-view-components.md)

khiga8:
Do we expect these changes to _break_ any existing behavior? I don't think so, but just double checking!
Based on the docs, it doesn't look like there's feature flags in PVC so maybe spot checking is the best we can do?

test/system/alpha/select_panel_test.rb
```diff
@@ -1253,6 +1253,59 @@ def test_multi_select_form
       assert_equal ["item1", "item2"], response.dig(*%w(form_params item))
     end
 
+    def test_remote_single_select_form_pre_selection
```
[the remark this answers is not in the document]
kendallgassner: I don't think any new logic would be tested since we test remote with single_select.

kendallgassner:
> Do we expect these changes to break any existing behavior? I don't think so, but just double checking!
Based on the docs, it doesn't look like there's feature flags in PVC so maybe spot checking is the best we can do?


The only thing I think that could possibly break is if someone was using `data-list-input` to access information based on the [search results from github/github](https://github.com/search?q=repo%3Agithub%2Fgithub%20data-list-input&type=code). This is the only [file](https://github.com/github/github/blob/8fdac11e11763cb41389c188f4841a533a41a64c/app/views/repositories/_repository_results.html.erb#L6) we would need to update

kendallgassner:
@jonrohan since I am coordinating the release with you I am adding you as a reviewer.

jonrohan (approved):

langermank (approved):

emilybrick (approved):
Snapshot updates lgtm 📷

PR #3500, opened 2025-05-22
Address Zeitwerk reloading issue

neilang:
I observed when calling `reload!` in a Rails console (after the gem is eager loaded) results in an error:

```
Uncaught exception: TypeError - superclass mismatch for class CustomCitiesForm
```

I suspect this occurs because Zeitwerk expects a [single constant per file](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#one-file-one-constant-at-the-same-top-level). Calling `reload!` will remove the constant `ApplicationForm` but not `CustomCitiesForm` - which makes it a subclass of a [stale object](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#reloading-and-stale-objects). The next time the file is required, ruby assumes `CustomCitiesForm` is being redefined with a new superclass.

Steps to reproduce:
```
Zeitwerk::Loader.eager_load_all
reload!
Zeitwerk::Loader.eager_load_all
```


hectahertz (approved):
Good catch!

PR #3562, opened 2025-06-19
use ViewComponent 4.0.0.rc1

joelhawksley:

This PR integrates PVC with ViewComponent 4.0.0.rc1. This should be released in GitHub.com in concert with the ViewComponent release.

### DIFF
--- a/app/lib/primer/experimental_slot_helpers.rb
+++ b/app/lib/primer/experimental_slot_helpers.rb
@@ -22,7 +22,7 @@ module Primer
         registered_slots[slot_name][:renderable_hash][type] = poly_def
 
         define_method(:"with_#{type}") do |**system_arguments, &block|
-          set_slot(slot_name, poly_def, **system_arguments, &block)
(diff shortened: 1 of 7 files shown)


app/lib/primer/experimental_slot_helpers.rb
```diff
@@ -13,7 +13,7 @@ def add_polymorphic_slot_type(slot_name:, type:, callable:)
         slot_def = registered_slots[slot_name]
         raise "Unknown slot '#{slot_name}'" unless slot_def
 
-        poly_def = define_slot(
+        poly_def = __vc_define_slot(
```
marcelolx, line 16: @joelhawksley is there any plan on eventually turning `__vc_define_slot` and `__vc_set_slot` into public APIs so consumers of ViewComponents can do things like this? Just curious

joelhawksley: @marcelolx I'm pretty hesitant to do so, but if there were other use cases I'd consider them. Feel free to file something on the ViewComponent repo!

PR #3661, opened 2025-08-28
Render tooltip after the control

khiga8:

In the `NavList` component, @lindseywild and I noticed that when a tooltip is present, it gets rendered before the control its tied to.

This is unintuitive. For example, screen reader users who may want to navigate to the tooltip (e.g. to copy the text, or read it again). This PR moves the tooltip to render after the trigger.

### Screenshots

After:
Tooltips renders after button in the DOM

#### Risk Assessment
  

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/ninety-games-bake.md
+++ b/.changeset/ninety-games-bake.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+Render tooltip after the control


app/components/primer/beta/nav_list.ts
```diff
@@ -60,14 +60,20 @@ export class NavListElement extends HTMLElement {
 
   // expand collapsible item onClick
   expandItem(item: HTMLElement) {
-    item.nextElementSibling?.removeAttribute('data-hidden')
```
khiga8, line 63: This assumes the `ul` always appears after the `button`, which isn't the case when we're dealing with buttons.

.changeset/ninety-games-bake.md
```diff
@@ -0,0 +1,5 @@
+---
```
khiga8, line 1: Is there a way I can open a integration test PR in dotcom, similar to primer/react? Or, if you have any suggestions to safely rolling this out, let me know!

khiga8:
I think this update is valuable for consistent user experience. In our other components, we intentionally place tooltips after their triggers, so the content is available in a consistent, easily accessible location.

I expected this to be a straightforward ERB change, but I had to make CSS/JS adjustments to ensure the collapsing behavior works correctly...

Given the additional complexity, I'm not so sure... I'd be comfortable closing this/putting this on the backburner, if this seems challenging to roll out or is a risky change given widespread use in dotcom.  🤔 

What do you think @primer/engineer-reviewers?

lindseywild (approved):
Deferring to primer engineering for this, but interested to check it out on RL!

.changeset/ninety-games-bake.md
TylerJDev: cc: @jonrohan, is there an easy way to test this out before merging?

TylerJDev (approved):
Looks good to me! If we can test this behavior out beforehand that will be great, but if not then it seems low risk.

khiga8:
Got it @jonrohan!

Per [my last comment](https://github.com/primer/view_components/pull/3661#issuecomment-3237486289), since this one seems higher-risk, let's revisit this at a future time!

PR #3663, opened 2025-09-02
Remove early return from error announcement logic

jonrohan:
Part of https://github.com/github/primer/issues/5409

I'm trying to surface any exceptions that might be thrown lower down in the stack. In this case an event was triggered with type "error" so I'm re-throwing the event as an exception.

I'm unsure if this opens up to way more exceptions from more than my intention. Let me know if you think I should focus this to only the error that I'm trying to target. `"Failed to load resource: expected ..."`

#### Risk Assessment
  
There's risk this suddenly spikes errors being reported to Sentry

- [x] **Medium risk** changes that are isolated, reduced in scope or could impact few users. The change will not impact library availability.

### DIFF
--- a/.changeset/seven-peaches-serve.md
+++ b/.changeset/seven-peaches-serve.md
@@ -1,0 +1,5 @@
+---
+"@primer/view-components": patch
+---
+
+When an error occurs in the SelectPanel, continue throwing the error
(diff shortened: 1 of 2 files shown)


siddharthkp (approved):
LGTM!

PR #3686, opened 2025-09-19
Implement forced colors for progress bar

langermank:
Closes https://github.com/github/primer/issues/5820

### DIFF
--- a/app/components/primer/beta/progress_bar.pcss
+++ b/app/components/primer/beta/progress_bar.pcss
@@ -20,6 +20,7 @@
 
 .Progress-item {
   outline: 2px solid transparent; /* Support Firefox custom colors */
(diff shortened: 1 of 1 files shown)


app/components/primer/beta/progress_bar.pcss
```diff
@@ -25,3 +25,14 @@
 .Progress-item + .Progress-item {
   margin-left: var(--base-size-2);
 }
+
+@media (forced-colors: active) {
+  :root {
+    --progress-bg: LinkText;
```
jonrohan: Where does this variable connect?

langermank: Oops, just fixed it

PR #3696, opened 2025-09-24
Fix label `for` attribute when not scoped to model

### LINKED ISSUE #3695: Label `for` attribute incorrect when `scope_id_false: false`
## Description

The generated label `for` attribute value does not respect the `scope_id_false:` option on `TextField`.

## Reproduction

```ruby
  include Primer::ComponentTestHelpers

  class NoModelScopeForm < ApplicationForm
    form do |no_model_scope_form|
      no_model_scope_form.text_field(
        name: :ultimate_answer,
        label: "Ultimate answer",
        scope_name_to_model: false,
        scope_id_to_model: false,
        **(@id ? { id: @id } : {})
      )
    end

    def initialize(id: nil)
      @id = id
    end
  end

    model = DeepThought.new(42)

    render_in_view_context do
      primer_form_with(model: model, url: "/foo") do |f|
        render(NoModelScopeForm.new(f))
      end
    end

    puts rendered_content


```

## Expected value

```html
<primer-text-field class="FormControl width-full FormControl--fullWidth">
  <label for="ultimate_answer" class="FormControl-label">
    Ultimate answer
  </label>
  <div class="FormControl-input-wrap">

    <input name="ultimate_answer" id="ultimate_answer" aria-describedby="validation-8aef820b-8020-427a-a0e4-7fdefa498ac3"
      data-target="primer-text-field.inputElement " class="FormControl-input FormControl-medium" type="text" value="42" />
  </div>
  <!-- ... -->
</primer-text-field>
```

## Actual value

```html
<primer-text-field class="FormControl width-full FormControl--fullWidth">
  <label class="FormControl-label" for="deep_thought_ultimate_answer">
    Ultimate answer
  </label>
  <div class="FormControl-input-wrap">
  
    <input name="ultimate_answer" id="ultimate_answer" aria-describedby="validation-e5d33a6f-6e21-4ab2-a6fb-3f09b61f940c"
      data-target="primer-text-field.inputElement " class="FormControl-input FormControl-medium" type="text" value="42" />
  </div>
  <!-- ... -->
</primer-text-field>
```

myabc:

Defers assigning label `for` value to later in `initialize` method, ensuring the `scope_id_false: false` option is respected.

### Integration

It may require updating calling code that relies on the incorrect behavior.

Closes primer/view_components#3695

#### Risk Assessment

- [X] **Medium risk** changes that are isolated, reduced in scope or could impact few users. The change will not impact library availability.

### Merge checklist

- [X] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/app/lib/primer/forms/dsl/input.rb
+++ b/app/lib/primer/forms/dsl/input.rb
@@ -61,7 +61,6 @@ module Primer
           @input_arguments = system_arguments
           @input_arguments.delete(:id) unless @input_arguments[:id].present?
           @label_arguments = @input_arguments.delete(:label_arguments) || {}
-          @label_arguments[:for] = id if id.present?
 
           @label_arguments[:class] = class_names(
             @label_arguments[:class],
@@ -104,6 +103,8 @@ module Primer
             @input_arguments[:id] = @input_arguments.delete(:id) { name }
           end
           # rubocop:enable Style/IfUnlessModifier
+
+          @label_arguments[:for] = @input_arguments[:id]
 
           # Whether or not to wrap the component in a FormControl, which renders a
           # label above and validation message beneath the input.
--- a/app/lib/primer/forms/dsl/input.rb
+++ b/app/lib/primer/forms/dsl/input.rb
@@ -104,7 +104,10 @@ module Primer
           end
           # rubocop:enable Style/IfUnlessModifier
 
-          @label_arguments[:for] = @input_arguments[:id]
+          # Only set `for` when the id is known. Otherwise leave it unset so
+          # Rails generates the scoped id for both label and input; passing
+          # `for: nil` would suppress that and drop the association.
+          @label_arguments[:for] = @input_arguments[:id] if @input_arguments[:id].present?
 
           # Whether or not to wrap the component in a FormControl, which renders a
           # label above and validation message beneath the input.
--- a/.changeset/chatty-pants-hang.md
+++ b/.changeset/chatty-pants-hang.md
@@ -1,0 +1,5 @@
+---
+'@primer/view-components': patch
+---
+
+Fix incorrect label `for` attribute value when `scope_id_to_model: false`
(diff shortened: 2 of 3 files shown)


siddharthkp:
Passing over to @jonrohan who would know more about this than me

myabc:
Just a quick note: I've kept this PR narrow in scope (forgive the pun) and tried to solve this in the "least invasive" way possible. However, there are other issues (such as namespacing - see #3698) for which I'm trying to find a more robust solution.

app/lib/primer/forms/dsl/input.rb
```diff
@@ -58,7 +58,6 @@ def initialize(builder:, form:, **system_arguments)
           @input_arguments = system_arguments
           @input_arguments.delete(:id) unless @input_arguments[:id].present?
           @label_arguments = @input_arguments.delete(:label_arguments) || {}
-          @label_arguments[:for] = id if id.present?
```
jonrohan, line 64: This was calling `def id` below which is the same as `@input_arguments[:id]` curious how this is different than what was already in place

myabc: @jonrohan sorry for only just getting back to you. The order here is important. We should defer setting `@label_arguments[:for]` until after these lines, which can mutate `@input_arguments[:id]`:

```ruby
          unless @input_arguments.delete(:scope_id_to_model) { true }
            @input_arguments[:id] = @input_arguments.delete(:id) { name }
          end
```

jonrohan (approved):

myabc:
@jonrohan took a moment to revisit this. Unfortunately my solution was missing a `if @input_arguments[:id].present?` guard (which could have caused another regression). Fixed in [4a06e61](https://github.com/primer/view_components/pull/3696/commits/4a06e618e19213856e421bc3ffbd73c337656624)

jonrohan (approved):

PR #3938, opened 2026-02-13
Change `ToggleSwitch` ViewComponent `type` to `button`

acbreton:
Closes: https://github.com/primer/view_components/issues/3939

Ensure `Primer::Alpha::ToggleSwitch` does not accidentally submit its parent `<form>` when clicked. The ToggleSwitch track is rendered as a `<button>` and, without an explicit `type`, defaults to `type="submit"` inside forms. This triggers Chrome’s native required-field validation tooltips on unrelated empty required fields. This change sets `type="button"` on the ToggleSwitch track button.

### Screenshots
Before: Clicking ToggleSwitch inside a form shows native validation tooltips on other empty required fields (Chrome).

https://github.com/user-attachments/assets/3e29c41c-8bd6-42ac-afeb-82a2021927a0

After: Clicking ToggleSwitch no longer triggers native validation tooltips; only the toggle state changes.

https://github.com/user-attachments/assets/5d97ea29-0074-4255-a977-bb738e006d42

### Integration
No production code updates required beyond consuming the updated component. This is a component-level change.

Closes https://github.com/github/authz-exp/issues/1323

#### Risk Assessment
- [x] **Low risk** — One-line HTML attribute fix (`type="button"`) on a single internal button. Easy to observe and roll back. Adds test coverage.

Set `type="button"` on the internal `.ToggleSwitch-track` button unconditionally. ToggleSwitch is an interactive control (and optionally performs its own POST via `src`), so it should never behave like a form submit button by default. Making this explicit prevents accidental form submission and browser validation UI across all usages inside forms.

- Confirm there are no supported/intentional use cases where ToggleSwitch should submit a parent form via native submit semantics. (If needed, we can discuss an explicit API, but the current default behavior is a footgun.)
- Added a test asserting the `type="button"` attribute is present on the track button.

### Accessibility
- **No new axe scan violation** — Prevents unexpected browser validation UI; no changes to roles/aria beyond button type.

### Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/app/components/primer/alpha/toggle_switch.html.erb
+++ b/app/components/primer/alpha/toggle_switch.html.erb
@@ -8,7 +8,7 @@
     <%= render(Primer::Box.new(classes: "ToggleSwitch-statusOff").with_content(@off_label)) %>
   <% end %>
 
-  <%= render(Primer::BaseComponent.new(tag: :button, type: :button, classes: "ToggleSwitch-track", disabled: disabled?, data: { target: "toggle-switch.switch", action: "click:toggle-switch#toggle" }, **@button_arguments)) do %>
+  <%= render(Primer::BaseComponent.new(tag: :button, classes: "ToggleSwitch-track", disabled: disabled?, data: { target: "toggle-switch.switch", action: "click:toggle-switch#toggle" }, **@button_arguments)) do %>
     <%= render(Primer::Box.new(classes: "ToggleSwitch-icons", aria: { hidden: true })) do %>
       <%= render(Primer::Box.new(classes: "ToggleSwitch-lineIcon")) do %>
         <%= render(Primer::BaseComponent.new(


app/components/primer/alpha/toggle_switch.html.erb
```diff
@@ -8,7 +8,7 @@
     <%= render(Primer::Box.new(classes: "ToggleSwitch-statusOff").with_content(@off_label)) %>
   <% end %>
 
-  <%= render(Primer::BaseComponent.new(tag: :button, classes: "ToggleSwitch-track", disabled: disabled?, data: { target: "toggle-switch.switch", action: "click:toggle-switch#toggle" }, **@button_arguments)) do %>
+  <%= render(Primer::BaseComponent.new(tag: :button, type: :button, classes: "ToggleSwitch-track", disabled: disabled?, data: { target: "toggle-switch.switch", action: "click:toggle-switch#toggle" }, **@button_arguments)) do %>
```
jonrohan: In this case, if someone does want `type: :submit` they would pass it in with @button_arguments and I think might not resolve correctly, what do you think about making this a default in the `@button_arguments` that can be changed?

acbreton: Sure thing, I'll add that.

acbreton: Added in https://github.com/primer/view_components/pull/3938/commits/fb7110fdac109d8cacaabe07bc2b31ee7ede22cd + https://github.com/primer/view_components/pull/3938/commits/dad40e83ca518e223db79d9593b986ea0a917a5d

lesliecdubs:
Jon's out of office for the next week, tagging the review team to see about getting a faster ✅

